Issue #20, opened 2017-01-24
Build pipeline improvements


binarymash:
There's a number of improvements we can make to the build process...

- migrate build scripts to cake to encourage a more formal build pipeline that can also be easily used in AppVeyor
- SemVer versioning
- Cleaning up the nuget feed by publishing unstable builds to a separate feed.
- Hook into the github release process for stable releases, including documentation of what has changed

I have this kind of stuff working on another project, based on how the folks at BDDfy do it as documented at http://www.michael-whelan.net/continuous-delivery-github-cake-gittools-appveyor/ - shouldn't be too much effort to migrate the current bat scripts.

TomPallister:
@binarymash looks good. I'll re read the documentation as a few things I don't quite get yet. Do I need to give you access to AppVeyor?

binarymash:
There will be some changes to the current appveyor config, but lets worry about that once we've got a pull request that you're happy with. For now I'm testing it with my own appveyor project at https://ci.appveyor.com/project/binarymash/ocelot/

PR #21, opened 2017-01-29
#20 - refactored build pipeline

binarymash:
Ok, you definitely won't want to merge this yet - suggest you have a look and decide if this is what you want, branching-wise. 

This code assumes you'll want to git flow, which will give you control over when you publish releases - ie, a release might contain several stories. However, there's a few gotchas I discovered with the gitflow process, which seem to break the versioning, and the integration with github releases adds a bit to the release process which you might not think necessary. So, you might prefer moving to githubflow, which would force you into a new SemVer version for each commit. Shouldn't be difficult to modify this for githubflow, though I haven't tried that.

Anyway, have a look, and then I'll go through it with you in more detail.


binarymash:
Note that there's stuff in the cake file config that is specific to my fork - feed URLs etc - so these will need changing before any merge. And there'll be stuff to set up in AppVeyor too.

TomPallister:
Thanks @binarymash I'll look into this. Being fair any branching and versioning strategy is going to be better than what I'm doing now!

binarymash:
Hmm, maybe I should have left the existing .bat scripts in there so the current build process doesn't fail :)

I guess the best way to test this all is for you to fork my code and play around with creating releases, configuring AppVeyor etc.

TomPallister:
@binarymash setting aside time tonight to do this review!

TomPallister:
This all looks good @binarymash , I'm happy with the gitflow strategy so going to merge. Only thing is I don't think I can change is the nuget feeds in your pull request so will just merge and fix after. Not like anyone important is using this project yet. The tradeoff is having a broken build for a bit with a good build pipeline going forward I guess.

TomPallister:
@binarymash might move all the build scripts into a /build folder one day as root is getting a little packed!

binarymash:
ok, I can give you a hand getting AppVeyor reconfigured so everything works automagically again. Catch up tomorrow?

TomPallister:
@binarymash thanks sounds good, ill have a poke around see if i can get it building in AppVeyor :)

PR #22, opened 2017-02-01
Refactor load balancers and implementing Consul Service discovery

geffzhang:
Refactor load balancers  ILoadBalancer and implementing Consul Service
discovery , RegistryClient Responsible for the lease HostAndPort

### DIFF
--- a/test/Ocelot.UnitTests/ConsulRegistryConfiguration.cs
+++ b/test/Ocelot.UnitTests/ConsulRegistryConfiguration.cs
@@ -1,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Ocelot.UnitTests
+{
+    public class ConsulRegistryConfiguration
+    {
+        public string HostName { get; set; }
+        public int? Port { get; set; }
+    }
+}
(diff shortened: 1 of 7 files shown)


TomPallister:
@geffzhang code looks good!

If you look at my latest commits to the service discovery branch Imthinking we should have something like

ServiceProviderFactory.Get(reRoute) this uses the reRoute to build an IServiceProvider.

IServiceProvider would be ConsolServiceProvider / NoServiceProvider (if just using one service from configuration)

Then we have ILoadBalancerFactory.Get(ReRoute) this takes the reroute and builds a ILoadBalancer for that reRoute.

ILoadBalancer is NoLoadBalancer, RandomLoadBalancer, RoundRobinLoadBalancer and LeastConnectionLoadBalancer. 

For me the best load balancer is least connection but that is more complicated because we have to tell the load balancer when we have finished with the connection to the downstream service.

I am thinking we could make a LoadBalancing middleware that is reponsible for getting the HostAndPort of downstream service. Then call next middleware, then release connection. The call to next middleware would need to be try catch which I don't like :(

geffzhang:
@TomPallister 
I read your latest commit, your design is better, I wrote the code completion feature that I very much agree with your design, load balancing is designed as a middleware.

TomPallister:
@geffzhang Im going to close this pull request. If you want to make changes to the load balancers implementation open a pull request. I need to improve the least connection load balancer as it uses a lock which I do not like!

Issue #23, opened 2017-02-02
implements the Circuit Breaker Pattern (using Polly from thepollyproject.org)


geffzhang:
HttpClientHttpRequester need implements the Circuit Breaker Pattern，I using Polly from thepollyproject.org https://github.com/App-vNext/Polly/wiki/Transient-fault-handling%3A-why-Retry-and-Circuit-Breaker%3F

TomPallister:
@geffzhang I know Polly so this should be quite simple. The hard part will be configuring it in the Json. Maybe have a go implementing this with a pull request.

geffzhang:
I go to implement Polly configuration which should be on the ReRoute

geffzhang:
#27 @TomPallister I have refactor polly circuit breaker pattern， config it in json： 
"ExceptionsAllowedBeforeBreaking","DurationOfBreak","TimeoutValue"

TomPallister:
@geffzhang I have reviewed the code looks good, when I merge the changes I will add information to the README to tell people how to use the feature.

TomPallister:
closing this as pull request merged

PR #27, opened 2017-02-03
Circuit breaker pattern

geffzhang:
Refactor CircuitBreaker Pattern and Add Qos Config in ReRoute


TomPallister:
@geffzhang code looks good!

I'm happy to merge this after I have finished the service discovery work.

TomPallister:
Fixes #23 and is ok to merge

geffzhang:
@TomPallister I merge the code ,but AcceptanceTests should_use_service_discovery_and_load_balance_request could not pass

TomPallister:
@geffzhang I will pull the code and look at the tests.

TomPallister:
@geffzhang the test is failing because CreateHttpMessageHandler() is broken when qos.ExceptionsAllowedBeforeBreaking is 0. I think Polly needs greater than 0.

Perhaps it would be better to add the QoS configuration to FileReRoute.cs if this is left null the user does not want to use QoS, else use the settings in the QoS configuration. So create a class like below.

public class FileQoS
{
       public int ExceptionsAllowedBeforeBreaking { get; set; }

        public TimeSpan DurationOfBreak { get;set; }

        public TimeSpan TimeoutValue { get; set; }

        public TimeoutStrategy TimeoutStrategy { get; set; }
}

Then add Qos property to ReRoute, change constructor to take QoS and expose public Getter.

This is then used in FileOcelotConfigurationCreator.SetUpReRoute(FileReRoute fileReRoute, FileGlobalConfiguration globalConfiguration)

We map FileQos to Qos and populate the ReRoute QoS parameter.

Then when we build the Request in HttpRequestCreator we check to see if we should use QoS. If true call WithQos else don't.

Then in HttpClientBuilder we check if we should use QosS before adding a handler.

QoS should be turned off by default and the user can have the option to turn it on?

geffzhang:
@TomPallister I Refactor qos as options，But  another test should_use_global_request_id_and_forward not pass

TomPallister:
@geffzhang I will check again when I get home from work :)

test/Ocelot.AcceptanceTests/RequestIdTests.cs
```diff
@@ -95,8 +95,7 @@ public void should_use_global_request_id_and_forward()
                             DownstreamPort = 51879,
                             DownstreamScheme = "http",
                             DownstreamHost = "localhost",
-                            UpstreamTemplate = "/",
-                            UpstreamHttpMethod = "Get",
+                            UpstreamTemplate = "/",
```
TomPallister: @geffzhang This test needs UpstreamHttpMethod = "Get",

TomPallister:
@geffzhang RequestIdTests.should_use_global_request_id_and_forward needs UpstreamHttpMethod = "Get" adding to FileReRoute

TomPallister:
@geffzhang thank you for helping with Ocelot.

rahulbhat13:
@TomPallister @geffzhang I am new to ocelot , i want to use the x-forwaded filters ,qos and circuit breaker to pivotal cloud foundry so can you please suggest about this .
Qos is just a Json config i believe , x-forwaded i have no idea as of now and how polly will help in pcf . Please suggest

PR #49, opened 2017-02-19
Fix #41 - get cake scripts running on Mac

### LINKED ISSUE #41: Get Cake build working on MacOS
Think the problem is that cake cannot reference variables declared at top of page in private methods on macos...no idea why it can on windows!

binarymash:
Also added compile-time support for OSX.10.12-x64. The acceptance tests still fail on this version of OS X, however, because of the funky stuff going on in TestConfiguration.cs. Fixing this is outside the scope of this issue.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -80,7 +80,7 @@ Task("Version")
 		if (AppVeyor.IsRunningOnAppVeyor)
 		{
 			Information("Persisting version number...");
-			PersistVersion(nugetVersion);
+			PersistVersion(committedVersion, nugetVersion);
 			buildVersion = nugetVersion;
 		}
 		else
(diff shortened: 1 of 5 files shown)


binarymash:
The build now works on my mac and windows. Note that the release notes cannot be generated on osx because that uses an exe that is not cross platform.

I had to add runtimes for osx.10.12-x64 to do the work on my mac. This exposes problems in the acceptance tests - they fail for me because the config file switcheroo that is going on in TestConfiguration.cs assumes osx.10.11-x64 if you're on a mac and I didn't know how to fix. However, I guess you're using osx.10.11-x64 so should be fine for you for now. I remember that this config stuff also caused issues when I set up the original cake scripts, so would be good to fix.

TomPallister:
@binarymash thanks! All looks fine. I'm going to sort that testconfig thing now.

PR #53, opened 2017-02-19
removed test configuration nonsense

### LINKED ISSUE #50: Find a replacement for TestConfiguration.cs
The acceptance tests rely on TestConfiguration to get the path to the configuration file used by the tests. This causes a couple of problems...

1. The build configuration is hard-coded to debug
2. It currently fails if you are using osx.10.12-x64.

We should be able to replace this with dynamic code that calculates the config file path dynamically. Surely we can do something like 

```csharp
using System.IO;

\\...

return Path.Combine(Directory.GetCurrentDirectory(), "configuration.json");
```

TomPallister:
fixed #50


TomPallister:
manually tested on mac

PR #54, opened 2017-02-19
Fix package publishing

binarymash:
Check that the package is published before merging this!

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -282,7 +282,7 @@ private void PersistVersion(string committedVersion, string newVersion)
 }
 
 /// generates release notes based on issues closed in GitHub since the last release
-private void GenerateReleaseNotes(string file)
+private void GenerateReleaseNotes(ConvertableFilePath file)
 {
 	if (!IsRunningOnWindows())
 	{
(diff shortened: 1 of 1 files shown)


binarymash:
So, the CreatePackages step has failed but I'm not sure why because the path looks correct for the releasenote.md. I'm also not sure why this is not failing the whole build.

The build is working fine for me locally when I check out the pull request. @TomPallister can you kick off the build again in Appveyor? I wonder if its just a random failure for some reason.

binarymash:
btw, looks like your feature branches are only publishing their packages for the first commit on the branch. This is because gitversion is currently configured to not generate a unique semver version on this branch (this is the default gitversion behaviour).. It's only the develop branch that generates a unique semver. On my own projects I choose not to publish unstable packages from feature branches..

binarymash:
Good, working now, must have just been a dodgy build in AppVeyor.

TomPallister:
How do you stop it publishing unstable commits? Sorry for noobiness.

binarymash:
I wrote my own code in the cake scripts - have a look at https://github.com/binarymash/responses/blob/develop/build.cake - mostly the same as on this project, but also has a ShouldPublishToUnstableFeed method.

Another reason not to publish all feature branches to the unstable feed is that, depending on name, they might be higher priority than the develop branch. Priority is reverse alphanumeric order, and the develop branch has an "unstable" suffix so if you call your branch "zzz" then this will take higher priority than the most recent develop build, and if you call it "aaa" then it will take lower priority. So, the behaviour is a bit non-deterministic. Pull requests are fine, I think, because they have a "pullrequest" suffix and so will be lower priority than develop.

This is also why I publish my master and release (with a "beta" suffix) branches to the stable feed, and keep unstable develop on it's own feed which people can subscribe to if they want.

TomPallister:
@binarymash stole your publish unstable features code, seemed to work! Thanks.

PR #59, opened 2017-03-05
Remove use of CookieContainer

TomPallister:
I realised we can get rid of the cookie container as cookies are just sent as a header called cookie...


TomPallister:
@geffzhang please look at pull request if you have time. Do you agree with changes?

Issue #66, opened 2017-03-13
ResponderMiddleware will never set a 500 response code


MarcDenman:
[ResponderMiddleware.cs](https://github.com/TomPallister/Ocelot/blob/c09cec67c63a543eefbfbffafff3c2662b356c85/src/Ocelot/Responder/Middleware/ResponderMiddleware.cs#L59) looks as if it can set a 500 to be returned, however, the implementation of [ErrorsToHttpStatusCodeMapper](https://github.com/TomPallister/Ocelot/blob/c09cec67c63a543eefbfbffafff3c2662b356c85/src/Ocelot/Responder/ErrorsToHttpStatusCodeMapper.cs) creates a new OkResponse and does not pass the list of errors through. 

As such, when `ResponderMiddleware` calls [IsError](https://github.com/TomPallister/Ocelot/blob/c09cec67c63a543eefbfbffafff3c2662b356c85/src/Ocelot/Responses/Response.cs#L20) `false` will always be returned.

I am 85% sure what I have said above is correct, as I am still learning the codebase, happy to PR though if I can be pointed in the correct direction.

TomPallister:
@MarcDenman you are correct, I would change the ErrorsToHttpStatusCodeMapper to not return with a wrapped Response<T> value. Would be great if you could make that change otherwise just leave it here and ill get to it. Good spot.

Really grateful for the assistance. I know from experience getting a second pair of eyes on my code will pick up these kind of issues so much appreciated.

MarcDenman:
Excellent, not a problem, happy to help out :) 

I have made a PR with the changes and fixed up the unit tests

TomPallister:
@MarcDenman merged PR closing issue!

PR #67, opened 2017-03-14
Change HttpStatusCodeMapper not to wrap responses

MarcDenman:
As part of #66 we realised that the implementation of
IErrorToHttpStatusCodeMapper would always return a wrapped StatusCode
within an OK response, in turn meaning that ResponderMiddleware would
never fall into the else branch for returning a 500.

This commit removes the wrapping of the status code and removes the unused
logic for generating the 500 status code, giving the mapper full
responsbility for generating the correct status code.


TomPallister:
@MarcDenman looks good thanks for the help!

PR #69, opened 2017-03-14
Properties renamed for consistency with Identity Server

FryDerm:
Renamed ScopeName to ApiName, ScopeSecret to ApiSecret, and AdditionalScopes to AllowedScopes in order to be more consistent with Identity Server naming conventions.


TomPallister:
@FryDerm thank you very much looks great!

Issue #35, opened 2017-02-11
Better logging / Check logging


TomPallister:

TomPallister:
its probably a bit rubbish :(

MarcDenman:
Any hints/suggestions for this? Are there any areas that need a bit more logging than others?

Ocelot looks a lot like something I had been to toying with about 6 months ago except far far better! Probably going to look at the codebase other the weekend so quite happy to start looking at improving the logging if that would be helpful.

TomPallister:
@MarcDenman when I started writing Ocelot I wrote a wrapper around the standard .NET core logging stack for some reason and I can't really remember now which is not good! It might have been because I wanted to allow the user to specify a request id / correlation id that their requests should log to rather than the .NET request id stuff. 

Thinking about this I'm not really happy with how all that worked so I was going to review it and see if the code could be improved.

There is also an issue where some places take a logger factory and create a new logger in the constructor and other places where we pass a logger in. I would probably favour doing the same thing everywhere as it avoids confusion.

Also all the logging statements are directly in code that they are logging about and I was thinking about using the decorator pattern to move the logging out of the class doing logic.

Finally the log messages are pretty sporadic. I would just like to organise it. 

I know how important good logging is and I'm annoyed at myself for letting it slip a bit.

Any help would be appreciated.

MarcDenman:
@TomPallister That is brilliant information, thank you :) 

Unless you have any huge objections, I am going to start by trying to get all the logging injection consistent (probably using the `IOcelotLoggerFactory` for now) and just make sure there are logging statements in all the right places. 

Hopefully after that any refactoring for the wrapper will be a little bit easier.

MarcDenman:
Just made a [commit ](https://github.com/TomPallister/Ocelot/commit/3fb83077fb8e9f27d76a69c47003df35c7955103) for some changes to the responderMiddleware. 

I know it is only a little change but could someone double check and make sure it is roughly what was wanted, didn't realise there were so many middlewares and don't want to get halfway through before realising I have gone down the wrong track.

I am also going to expand the `IOcelotLogger` to also implement LogTrace to try and give a little bit more granularity.

TomPallister:
@MarcDenman I've left a comment on the commits, let me know what you think.

MarcDenman:
Sorry, meant to reply yesterday but forgot. Your comment sounds good, I have added a prop on the Base OcelotMiddelware class which sets the name when it is initialised. 

[...]

PR #70, opened 2017-03-14
Update logging

MarcDenman:
As part of #35 I have been trying to make sure and double check that all the middlewares have roughly consistent logging. This PR has the base changes for that, adding MiddlewareName to the base class, expanding the logger interface to allow a couple extra useful methods, extension method to allow a list of errors to be converted into an error string.

Additionally it contains the first few middlewares from the pipeline (except ExceptionHandlerMiddleware)

### DIFF
--- a/src/Ocelot/Authentication/Middleware/AuthenticationMiddleware.cs
+++ b/src/Ocelot/Authentication/Middleware/AuthenticationMiddleware.cs
@@ -34,7 +34,7 @@ namespace Ocelot.Authentication.Middleware
 
         public async Task Invoke(HttpContext context)
         {
-            _logger.LogTrace($"entered {MiddlwareName}");
+            _logger.TraceMiddlewareEntry();
 
             if (IsAuthenticatedRoute(DownstreamRoute.ReRoute))
             {
@@ -46,6 +46,7 @@ namespace Ocelot.Authentication.Middleware
                 {
                     _logger.LogError($"Error getting authentication handler for {context.Request.Path}. {authenticationHandler.Errors.ToErrorString()}");
                     SetPipelineError(authenticationHandler.Errors);
+                    _logger.TraceMiddlewareCompleted();
                     return;
                 }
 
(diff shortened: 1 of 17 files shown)


TomPallister:
@MarcDenman this is really good, thank you very much!

What do you think about doing something to not have to add {Middleware} everywhere on the logging calls? I don't know if it works in .NET core but in good old fashioned .NET we have CallerMemberName attribute. I don't know if you are familiar with it so forgive me if I'm being patronising. 

We could maybe do this 

```csharp
 CreateLogger([CallerMemberName]string memberName = "")
 {
      //store member name, which should be constructor name, then use it for subsequent logging.
 }
```

What do you think? I'm not sure if this helps but throwing it out there also no idea if it will work!

MarcDenman:
@TomPallister Sorry for the delay in replying, only just seen this.

Nope, not being patronising at all :) I have never come across it, good to learn about hidden gems like that. I will take a look at it and try to see if it helps.

Probably won't get around to it till Saturday, got a busy couple of evenings ahead of me.

TomPallister:
@MarcDenman no worries no rush! Have to fit this stuff in around real life.

MarcDenman:
Started looking at it, [CallerMemberName] will unfortunately not work, as the logger is created within the constructor of the middleware so the value will end up being ".ctor". However, the CreateLogger function has the signature `CreateLogger<T>`, where T is the type you are logging for. I tried to get the type out of the logger, however it doesn't appear to be exposed easily through Microsoft.ILogger.  Instead, I have altered the IOcelotLogger to have a Name property (I want to think of a better name) and altered the constructor for the AspDotNetLogger so the name can be set when it is initialized.

I have also hung a few extension methods(not ideal I know) off IOcelotLogger for a couple common logging statements, i.e. MiddlewareEntered, MiddlewareCompleted

Not sure what to do with the MiddlewareName property, tempted to leave it as it could still be useful  for logging purposes.

Thoughts?

TomPallister:
@MarcDenman looks good. Are you happy to merge this?

MarcDenman:
Yep. happy if you are.

PR #71, opened 2017-03-16
 string.Equals under Centos not expect

geffzhang:
This string.Equals Method Windows is normal, but running under Centos, due to a case of the string problem, seems to be. NET core implementation issues on a Linux system is case sensitive


TomPallister:
@geffzhang sorry its taken me a while to merge this! Looks good.

PR #72, opened 2017-03-18
refactor:  replace AddSingleton with TryAddSingleton

geffzhang:


TomPallister:
@geffzhang what is the difference between TryAddSingleton  and AddSingleton?

geffzhang:
@TomPallister The registration that is, if it detects a DI container which already has a current registration Interface or Service, would not have to register again, not registered will go in. So why here you want to use? It is as if the user has implemented this interface and registered vessel, using the user registration, rather than the middleware itself. Users can very well to customize the functions provided by the middleware, which is polymorphism in OO, this principle is replaced on the Richter scale , 

https://github.com/aspnet/DependencyInjection/commit/8834f1aea5de3a387d291ebd8687c7af2096abdf

TomPallister:
@geffzhang sounds good, thanks for the explanation.

PR #82, opened 2017-04-04
Http client performance

geffzhang:
After stress testing of HttpClient, optimize the Httpclient-cache, use the ConcurrentDictionary and ConcurrentQueue replace MemoryCache


TomPallister:
@geffzhang brilliant! Thank you. I will review this ASAP!

liamyue:
The HttpClient class instance acts as a session to send HTTP requests. An HttpClient instance is a collection of settings applied to all requests executed by that instance. In addition, every HttpClient instance uses its own connection pool, isolating its requests from requests executed by other HttpClient instances.
https://msdn.microsoft.com/en-us/library/system.net.http.httpclient(vs.110).aspx#Anchor_5
For these infomations,should i still need to keep a ConcurrentQueue for the HttpClient?

Issue #84, opened 2017-04-16
Random thoughts on middleware


binarymash:
Hey @TomPallister,

Just been looking through the code trying to understand how it all works. I've got a few thoughts about the middleware implementation...

1. Is there any reason for `ExceptionHandlingMiddleware` to not implement `OcelotMiddleware`?

2. Any reason for not defining `abstract Task Invoke(HttpContext context)` on `OcelotMiddleware`? Adding this would give extra compile-time check that middleware implementors are on the correct path.

3. It took me a while to understand what parts of the code were middleware, and what parts of the code were infrastructure stuff. Given that the middleware are effectively the feature slices of the project, and are where all the magic happens, I feel it could do with being a bit more distinct from the rest of the supporting code. Any opinion on grouping all the middleware together as individual folders under the an `Ocelot.Middleware` namespace? Eg,

```
+-Ocelot
  +-Middleware
  | +-Authentication
  | | +- ...
  | |
  | +-Authorisation
  | | +- ...
  | | 
  | +-...
  |
  +-other stuff
```

This structure would also provide a path to providing a more modular package-based approach to middleware in the future, with a common naming convention: optional middleware could be published in packages; the common `Ocelot.Middleware.Whatever` convention would make it easy to discover these packages; and say someone didn't want to use identityserver then they could choose to avoid pulling this package in. 

If you think any of this is worth doing I'm happy to pick up the work on it.

TomPallister:
@binarymash makes sense, I tried to stuff all the classes below a relevant feature such as Authentication has all the code for authentication and anything cross cutting I just dumped in a catch all..probably called Infrastructure.

I find the code quite hard to follow myself because it is not directly procedural in the sense that you can go this method calls that and that calls this. Due to the fact that anyone can add middleware or do what they want basically it could all fall apart quite quickly!!

More than happy for you to make those changes! You shouldn't have any nightmare merges. 

The only feature Im thinking of doing is storing the config in Consul which gets me away from having to implement that raft consensus algorithm any time soon.

binarymash:
Cool bananas. That's my Easter Sunday afternoon sorted out then :)

I noticed a few naming inconsistencies between the folders and the middleware classes. I'll have a look at making everything consistent.

Oh yeah there's definitely still a need for the cross cutting stuff, no problem with that.

TomPallister:
@binarymash awesome thanks! I'm painting and decorating. It's harder than I thought to select a colour.

[...]

PR #88, opened 2017-04-18
Feature/request mutation

binarymash:
As discussed in #84, here's a refactor of the middleware so that they are not so tightly coupled.

The way the code works now is that there is a new middleware, `DownstreamRequestInitialiserMiddleware`, which runs at the beginning of the pipeline, creates an `HttpRequestMessage` based on the incoming request. Subsequent middleware modify this downstream request object directly, so there is no complex logic at the end of the pipeline to assemble the request and no need for the middleware to store anything on the scoped repo.

I don't believe I've broken anything; the acceptance and integration tests pretty much worked without changes. However, there's quite a lot of changes to unit tests so I can't but can't guarantee 100%. What I think might be a small change in behaviour now is that I think the port will always be added on to the downstream request. Not sure if that always happened or is something I introduced.

The request and requester middleware could probably be merged together now, but I've not done this.

No idea how this new architecture affects performance.


TomPallister:
@binarymash looks good. I'll have a proper look asap! Wasn't around last night.

TomPallister:
@binarymash I think it always added the port if I remember correctly. I took a view that if your downstream service doesn't work taking the port then you could submit a PR with a fix for it -_-

Cant see any performance issues that this would introduce. Potentially need to get the Benchmarking framework set up again as I'm just saying works at this point but not for this PR! 

I agree its a better architecture to create the request at the start and mutate it. I think I wanted to do this and ended up not doing it for no good reason.

Reviewing at the moment! Help much appreciated.

src/Ocelot/Middleware/OcelotMiddlewareExtensions.cs
```diff
@@ -53,6 +53,9 @@ public static class OcelotMiddlewareExtensions
         {
             await CreateAdministrationArea(builder);
 
+            // Initialises downstream request
+            builder.UseDownstreamRequestInitialiser();
+
             // This is registered to catch any global exceptions that are not handled
             builder.UseExceptionHandlerMiddleware();
```
TomPallister: Probably missing something but shouldn't this be the first piece of middleware, or are we assuming the UseDownstreamRequestInitialiser cannot throw an exception and if it does it will be caught!

src/Ocelot/Request/Middleware/DownstreamRequestInitialiserMiddleware.cs
```diff
@@ -0,0 +1,46 @@
...
+            _requestCreator = requestCreator;
+            _qosProviderHouse = qosProviderHouse;
+            _logger = loggerFactory.CreateLogger<DownstreamRequestInitialiserMiddleware>();
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            _logger.LogDebug("started calling request builder middleware");
+
+            var mapper = new Mapper();
```
TomPallister: Should we be injecting this guy? Probably dont need to tbh but thought id better ask.

src/Ocelot/Middleware/OcelotMiddlewareExtensions.cs
binarymash: I agree, not sure why it's not :)

src/Ocelot/Request/Middleware/DownstreamRequestInitialiserMiddleware.cs
binarymash: Yep no problem with doing that.

binarymash:
I'm out today, but will try to find the time to do these changes tonight.

TomPallister:
@binarymash all good dude no rush. We aint getting paid!

binarymash:
Ok, done. I noticed that I didn;t have much testing around the new middleware, so I added a load more. Also now handle errors in the mapper, because I noticed that if the scheme is not http or https then it fails.

btw, I copied most of the mapper code from the existing request middleware. Is there any reason why we reading the content stream only to then create a new one? Looks like we can just say Content = new StreamContent(request.Body), which I presume is more efficient. I've not committed this, but my tests seemed to pass with it in.

TomPallister:
@binarymash have you tried running the acceptance tests with the Content = new StreamContent(request.Body) change? I seem to remember having problems either with the acceptance tests or using it manually. Might be wrong though.

binarymash:
Hmm, interesting, I thought the tests all passed the other day when I tried using StreamContent, but looks like its causing some problems for me now. Definitely something work looking into, I think, but no need to make the change right now, as the code in this commit still uses the original functionality.

TomPallister:
@binarymash merged!

PR #89, opened 2017-04-20
Use final paths in the reroutes of configuration.json file.

juancash:
The idea is the user not have to map all request when use Ocelot like a simple request redirector.

Imagine the following scenario:

1. An intranet with three web apis.
2. An internal gateway like unique enter point to internal requests.
3. An external gateway like unique enter point to external requests.
4. The web apis (of point 1) can interact via internal gateway.
5. Our clients and mobiles apps use the api via external gateway.
6. The external gateway only offers a couple of methods that are mapped in the configuration file. If we need add or quit methods, we edit it´s configuration file.
7. The apis of intranet can changes frecuently. Every day our developers can implement two o three new controllers per api. We dont want have to edit the internal gateway configuration continously.

The configuration.json for internal gateway could be something like this:

```
{
      "UpstreamPathTemplate": "/api/api1/{urlPath}",
      "UpstreamHttpMethod": "Get",
      "DownstreamPathTemplate": "/api/{urlPath}",
      "DownstreamScheme": "http",
      "DownstreamHost": "localhost",
      "DownstreamPort": 5001,
      "QoSOptions": {
        "ExceptionsAllowedBeforeBreaking": 3,
        "DurationOfBreak": 10,
        "TimeoutValue": 10000
      }
    },
    ...
    {
      "UpstreamPathTemplate": "/api/api2/{urlPath}",
      "UpstreamHttpMethod": "Get",
      "DownstreamPathTemplate": "/api/{urlPath}",
      "DownstreamScheme": "http",
      "DownstreamHost": "localhost",
      "DownstreamPort": 5002,
      "QoSOptions": {
        "ExceptionsAllowedBeforeBreaking": 3,
        "DurationOfBreak": 10,
        "TimeoutValue": 10000
      }
    },
    ...
    {
      "UpstreamPathTemplate": "/api/api3/{urlPath}",
      "UpstreamHttpMethod": "Get",
      "DownstreamPathTemplate": "/api/{urlPath}",
      "DownstreamScheme": "http",
      "DownstreamHost": "localhost",
      "DownstreamPort": 5003,
      "QoSOptions": {
        "ExceptionsAllowedBeforeBreaking": 3,
        "DurationOfBreak": 10,
        "TimeoutValue": 10000
      }
    },
    ...
```

If we add new controllers to our web apis, we don´t need edit the internal gateway configuration, because the traffic will be redirected without problems.

Thanks.

### DIFF
--- a/src/Ocelot/DownstreamRouteFinder/UrlMatcher/UrlPathPlaceholderNameAndValueFinder.cs
+++ b/src/Ocelot/DownstreamRouteFinder/UrlMatcher/UrlPathPlaceholderNameAndValueFinder.cs
@@ -18,8 +18,8 @@ namespace Ocelot.DownstreamRouteFinder.UrlMatcher
                     if (IsPlaceholder(upstreamUrlPathTemplate[counterForTemplate]))
                     {
                         var variableName = GetPlaceholderVariableName(upstreamUrlPathTemplate, counterForTemplate);
-                        
-                        var variableValue = GetPlaceholderVariableValue(upstreamUrlPath, counterForUrl);
+
+                        var variableValue = GetPlaceholderVariableValue(upstreamUrlPathTemplate, variableName, upstreamUrlPath, counterForUrl);
 
                         var templateVariableNameAndValue = new UrlPathPlaceholderNameAndValue(variableName, variableValue);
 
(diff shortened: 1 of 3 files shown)


TomPallister:
@juancash Looks good thanks for the effort! Could you write an acceptance test for this so we know it works end to end? If you just add it to the RoutingTests.cs file. This shouldn't be too much work as you can just use the methods already in there! Basically copy a test and change it!

juancash:
@Tom Acceptance test added.

TomPallister:
@juancash thank you very much! I will merge this now.

PR #97, opened 2017-05-09
Updated CharactersDontMatch method

juancash:
I should have done it in the PR #89 but I didn´t notice. Sorry.

At this point the comparer must ignore case sensitive. It´s important for the proper functioning.

### DIFF
--- a/src/Ocelot/DownstreamRouteFinder/UrlMatcher/UrlPathPlaceholderNameAndValueFinder.cs
+++ b/src/Ocelot/DownstreamRouteFinder/UrlMatcher/UrlPathPlaceholderNameAndValueFinder.cs
@@ -70,7 +70,7 @@ namespace Ocelot.DownstreamRouteFinder.UrlMatcher
 
         private bool CharactersDontMatch(char characterOne, char characterTwo)
         {
-            return characterOne != characterTwo;
+            return char.ToLower(characterOne) != char.ToLower(characterTwo);
         }
 
         private bool ContinueScanningUrl(int counterForUrl, int urlLength)


juancash:
@TomPallister 

I´m looking why I didn´t notice this in the tests but seems I don´t understand something with the tests in class CaseSensitiveRoutingTests.

Now I have to leave urgently, but in a few of hours I´ll add a comment explaining my doubt.

TomPallister:
@juancash I think I originally had an option where a re route could be case sensitive (I'm not sure anyone would ever wants this). 

Perhaps we can remove that feature if its confusing.

juancash:
@TomPallister I think it´s not confusing.

The method that I changed in this PR is not related with the property:

`"ReRouteIsCaseSensitive": true`

This method is in the class UrlPathPlaceholderNameAndValueFinder where the target is look for the values in the upstreamUrlPathTemplate. In this point of the code the upstreamUrlPath and upstreamUrlPathTemplate can combine uppercase and lowercase chars. For this reason, the value search must be independent of whether it is upper or lower case.

```
Example 1:

upstreamUrlPathTempalte = "/products/{productId}"
upstreamUrlPath => "/products/1/"
```

```
Example 2:

upstreamUrlPathTempalte = "/products/{productId}"
upstreamUrlPath => "/PRODUCTS/1/"
```

Both examples must work in a real scenario, but I´m trying this and it´s not like that. However with this change the two examples works fine.

My doubt it´s, why the CaseSensitiveRoutingTests don´t detect this?

I´m checking this tests but I can´t understand how it works.

For example, the following test:

```
        [Fact]
        public void should_return_response_200_when_global_ignore_case_sensitivity_set()
        {
            var configuration = new FileConfiguration
            {
                ReRoutes = new List<FileReRoute>
                    {
                        new FileReRoute
                        {
                            DownstreamPathTemplate = "/api/products/{productId}",
                            DownstreamPort = 51879,
                            DownstreamScheme = "http",
                            DownstreamHost = "localhost",
                            UpstreamPathTemplate = "/products/{productId}",
                            UpstreamHttpMethod = "Get",
                        }
                    }
            };

            this.Given(x => x.GivenThereIsAServiceRunningOn("http://localhost:51879/api/products/1", 200, "Some Product"))
                .And(x => _steps.GivenThereIsAConfiguration(configuration))
                .And(x => _steps.GivenOcelotIsRunning())
                .When(x => _steps.WhenIGetUrlOnTheApiGateway("/PRODUCTS/1"))
                .Then(x => _steps.ThenTheStatusCodeShouldBe(HttpStatusCode.OK))
                .BDDfy();
        }
```

If I run this test, pass without problems. All it's OK because the ReRouteIsCaseSensitive property is not defined in the FileReRoute object and it's set to false by default. Now, forget the case sensitive features. If I simply change this:

`.When(x => _steps.WhenIGetUrlOnTheApiGateway("/PRODUCTS/5"))`

the test should not pass, because the service is

`this.Given(x => x.GivenThereIsAServiceRunningOn("http://localhost:51879/api/products/1", 200, "Some Product"))`

However the test is still going on.

What I´m missing here?

Thanks.

TomPallister:
@juancash Im going to merge this PR.

The fake service in the test 

this.Given(x => x.GivenThereIsAServiceRunningOn("http://localhost:51879/api/products/1", 200, "Some Product"))

Will respond to any HttpRequest.

The Ocelot config is set to forward requests from /products/{productId} to /api/products/{productId} on the downstream service. Ocelot will now just pass 5 to the fake service and it will return its response.

PR #99, opened 2017-05-10
Fixes to get build running again on macOS. 

binarymash:
Basically, re-applying changes from #41 which were lost in efec217.

We need to be careful here, because I remember the original changes in #41 broke the release process. However, I think this should good as we're using ConvertableFilePath etc rather than string, which iirc is what the original problem was.

Integration tests failing for me on the mac, but I don't think that's anything to do with any of this.


TomPallister:
@binarymash merged

Issue #103, opened 2017-05-18
loadBalancer Config Error


haptear:
this is my laodbalancer config,but run throw the exception  [System.Exception:“Unable to start Ocelot..configuration, errors were ”]
what wrong with the config?
{
  "ReRoutes": [
    {
      "DownstreamPathTemplate": "/api/web/cms/{urlPath}",
      "DownstreamScheme": "http",
      "DownstreamHost": "localhost",
      "DownstreamPort": 8901,
      "UpstreamPathTemplate": "/cms/{urlPath}",
      "UpstreamHttpMethod": "Get",
      "FileCacheOptions": {
        "TtlSeconds": 15
      },
      "LoadBalancer": "RoundRobin"
    },
    {
      "DownstreamPathTemplate": "/api/web/cms/{urlPath}",
      "DownstreamScheme": "http",
      "DownstreamHost": "localhost",
      "DownstreamPort": 8909,
      "UpstreamPathTemplate": "/cms/{urlPath}",
      "UpstreamHttpMethod": "Get",
      "FileCacheOptions": {
        "TtlSeconds": 15
      },
      "LoadBalancer": "RoundRobin"
    }
  ],
  "GlobalConfiguration": {
    "RequestIdKey": "OcRequestId",
    "ServiceDiscoveryProvider": {
      "Provider": "Consul",
      "Host": "localhost",
      "Port": 8901
    }
  }
}

juancash:
@haptear What version of Ocelot are you using?

geffzhang:
This is a bug, I have fixed , see https://github.com/TomPallister/Ocelot/pull/105

PR #105, opened 2017-06-03
I've applied Ocelot to the product，bug fixed

geffzhang:
I've applied Ocelot to the product , the bug rollup that was fixed during the test：
1、RoundRobin loadblancer cause config error #103
2、Fix Ratelimit every day, every minute, every hour logic bugs
3、refactor: Refactor HttpClientHttpRequester Httpclient queue


TomPallister:
@geffzhang thanks for the PR looks good. I will merge it!

PR #109, opened 2017-06-29
Feature/clear cache

TomPallister:
This PR gives the user the ability to call Ocelot administration API and clear a region (re routes) cache. The region is set in the Ocelot configuration (like everything else :o).

There is a http endpoint for this on OutputCacheController.

Because we are using cachemanager.net if you set up a backplane or use a distributed cache this gives you the ability to cache across a cluster of Ocelots and clear the whole cache by calling one member of the cluster....pretty neat!

At the moment I have only exposed clearing the cache by region but we could easily add a clear all endpoint in the future.


rajeshgithub001:
> This PR gives the user the ability to call Ocelot administration API and clear a region (re routes) cache. The region is set in the Ocelot configuration (like everything else :o).
> 
> There is a http endpoint for this on OutputCacheController.
> 
> Because we are using cachemanager.net if you set up a backplane or use a distributed cache this gives you the ability to cache across a cluster of Ocelots and clear the whole cache by calling one member of the cluster....pretty neat!
> 
> At the moment I have only exposed clearing the cache by region but we could easily add a clear all endpoint in the future.

It doesnt working, pls check issue https://github.com/ThreeMammals/Ocelot/issues/1212

PR #119, opened 2017-08-27
Feature/multi protocol

geffzhang:
In the process of using ocelot, need to support RPC protocols other than HTTP, so refactoring request and requester middleware support multi-protocol, default support HTTP, Https, can build on this code to support other RPC protocol, can extend the request and requester two middleware on the existing code infrastructure。 @TomPallister Please suggest that this can be merged into the code base, mainly the original HTTP code refactoring, according to the schema of the route extension of other protocols


TomPallister:
@geffzhang thanks for the contribution! :) I'll take a look at this as my next priority!

TomPallister:
@geffzhang thanks for the PR I really appreciate the effort. 

I have looked at the code and there is a problem with code coverage..when I run build.ps1 the build is failing because we don't have enough tests for the new code.

`Report generation took 3.3 seconds
We are not running on the build server so we won't publish the coverage report to coveralls.io
Sequence Coverage: 74.20
An error occurred when executing task 'RunUnitTests'.
Error: Code coverage fell below the threshold of 76.4%. You can find the code coverage report at https://coveralls.io/github/TomPallister/Ocelot`

I was also thinking that maybe the design could be improved and performance. With this code we build  another middleware pipeline during the users request based on the scheme from the configuration. However we know the scheme when ocelot starts so we could set this up at runtime rather than during a user request.

If you look at ILoadBalancerHouse and how we use that you can see we set up the infrastructure when Ocelot starts and then get it out of memory when we need it which is faster than building it during the users request.

I appreciate that this is probably only milliseconds worth of difference but its still faster! 

Also I've been looking at converting to .NET Core 2.0 and I'm not sure at the moment we can do things like this yet!

Anyway thanks for the PR the code looks good but can you take a look at my comments and see what you can do?

geffzhang:
@TomPallister Your suggestion is very good, which is also I did not consider the place, I follow your guidance to refine this part of the code

Issue #146, opened 2017-11-09
ReRoute has null CacheOptions when being fetched from cache


hampos:
Hey again @TomPallister 

I tried setting `FileCacheOptions` on my reroute, and have been getting `NullReferenceException` on this line:
https://github.com/TomPallister/Ocelot/blob/1d1a68ff95bb1caf5c279c64e18b5d168e82f7cd/src/Ocelot/Cache/Middleware/OutputCacheMiddleware.cs#L44

On the second time I perform a request on a reroute, and configuration has been cached.

It appears `DownstreamRoute.ReRoute.CacheOptions` is null when fetched from cache, did a little digging there and it seems to be coming from `JsonConvert` being unable to match `fileCacheOptions` with `CacheOptions` (from the JSON) in the constructor of `ReRoute`
https://github.com/TomPallister/Ocelot/blob/1d1a68ff95bb1caf5c279c64e18b5d168e82f7cd/src/Ocelot/Configuration/ReRoute.cs#L22
Tried renaming `fileCacheOptions` to `cacheOptions` and the deserialization happened successfully.

I am onto another error now, not sure if I should open another issue for that.
It's a deserialization error again
```
Exception caught in global error handler, exception message: Unable to find a constructor to use for type System.Net.Http.CurlHandler+CurlResponseMessage. A class should either have a default constructor, one constructor with arguments or a constructor marked with the JsonConstructor attribute. Path 'Version', line 1, position 11., exception stack:    at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateNewObject(JsonReader reader, JsonObjectContract objectContract, JsonProperty containerMember, JsonProperty containerProperty, String id, Boolean& createdFromNonDefaultCreator)
```
when trying to fetch from the cache the cached Response object for that ReRoute.
This is the cached data:
```
{
  "Version": {
    "Major": 1,
    "Minor": 1,
    "Build": -1,
    "Revision": -1,
    "MajorRevision": -1,
    "MinorRevision": -1
  },
  "Content": {
    "Headers": [
      {
        "Key": "Content-Length",
        "Value": [
          "142"
        ]
      },
      {
        "Key": "Content-Type",
        "Value": [
          "application/json; charset=utf-8"
        ]
      }
    ]
  },
  "StatusCode": 200,
  "ReasonPhrase": "OK",
  "Headers": [
    {
      "Key": "Date",
      "Value": [
        "Wed, 08 Nov 2017 13:20:08 GMT"
[...]

PR #151, opened 2017-11-13
Add acceptance test that exposes JSON deserialization bug from #146

hampos:
I have created a test that exposes the deserialization problem from issue #146 where `DownstreamRoute.ReRoute.CacheOptions` is null when fetched from cache.

I created a custom cache handle for CacheManager that mimics the behavior of the `DictionaryHandle` but uses the `ICacheSerializer` to serialize/deserialize values with Json.Net before saving them to an internal Dictionary.
When the handle tries to deserialize the saved configuration, it skips `CacheOptions` key because it mismatches with `fileCacheOptions`.
You can see the test failing if you revert the change on `ReRoute.cs` from 50ec8f17ea3dc6a817ece9a44fd71374c0d4e788

Don't know if this is overkill but couldn't think of any better way to expose this.

### DIFF
--- a/test/Ocelot.AcceptanceTests/Caching/InMemoryJsonHandle.cs
+++ b/test/Ocelot.AcceptanceTests/Caching/InMemoryJsonHandle.cs
@@ -1,0 +1,137 @@
+﻿using CacheManager.Core;
+using CacheManager.Core.Internal;
+using CacheManager.Core.Logging;
+using System;
+using System.Collections.Concurrent;
+using System.Linq;
+using static CacheManager.Core.Utility.Guard;
+
+namespace Ocelot.AcceptanceTests.Caching
+{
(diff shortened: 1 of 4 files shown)


TomPallister:
@hampos only issue I can see is a failing test.

Failed   Ocelot.AcceptanceTests.ConfigurationInConsul.should_return_response_200_with_simple_url
Error Message:
 System.IO.IOException : Failed to bind to address http://127.0.0.1:9500: address already in use.

The consul port for your new test just needs changing to something other than 9500 to get it working..suggest 9502!

hampos:
@TomPallister Changed the port to 9502, tests seem to be passing now.

hampos:
@TomPallister do you have indentation settings for the project? I am under the impression that our settings don't match. Would like to fix that before merging the PR.

PR #153, opened 2017-11-13
Implement mapping of HttpResponseMessage to CachedResponse to fix #152

### LINKED ISSUE #152: HttpResponse can't be deserialized when cached in Redis
This feels like a platform specific issue, I'm running Ocelot in a docker container from the `microsoft/aspnetcore:2.0` image which is a Debian implementation, but haven't tested it directly on Windows.
The issue happens when using a `CacheHandler` with Json Serialization and serializing an `HttpResponse`.
It seems that Json.NET can't deserialize the Unix implementation of HttpResponse `CurlResponseMessage` since it doesn't have a valid constructor.

```
Exception caught in global error handler, exception message: Unable to find a constructor to use for type System.Net.Http.CurlHandler+CurlResponseMessage. A class should either have a default constructor, one constructor with arguments or a constructor marked with the JsonConstructor attribute. Path 'Version', line 1, position 11., exception stack:    at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateNewObject(JsonReader reader, JsonObjectContract objectContract, JsonProperty containerMember, JsonProperty containerProperty, String id, Boolean& createdFromNonDefaultCreator)
```

hampos:
## Closes #152 
- #152 

I have created a test that exposes the deserialization problem from issue #152 where HttpResponse cannot be deserialized from cache.
I used the  `InMemoryJsonHandle` from the pull request https://github.com/TomPallister/Ocelot/pull/151 in order to do this.

I opened the pull request on develop, I think you can change this if you create another branch for this issue.


TomPallister:
@hampos looks good! Thanks for the help.

src/Ocelot/Cache/CachedResponse.cs
```diff
@@ -0,0 +1,21 @@
...
+	public class CachedResponse
+	{
+		public CachedResponse()
+		{
+			StatusCode = HttpStatusCode.OK;
+			Headers = new Dictionary<string, IEnumerable<string>>();
+			Body = "";
+		}
+
+		public HttpStatusCode StatusCode { get; set; }
```
TomPallister: Can we have private setters here please, the project only has public setters for the file configuration stuff because microsofts framework for config cant support private setters.

src/Ocelot/Cache/Middleware/OutputCacheMiddleware.cs
```diff
@@ -65,11 +67,54 @@ public class OutputCacheMiddleware : OcelotMiddleware
...
 
-            _outputCache.Add(downstreamUrlKey, response, TimeSpan.FromSeconds(DownstreamRoute.ReRoute.CacheOptions.TtlSeconds), DownstreamRoute.ReRoute.CacheOptions.Region);
+            _outputCache.Add(downstreamUrlKey, cached, TimeSpan.FromSeconds(DownstreamRoute.ReRoute.CacheOptions.TtlSeconds), DownstreamRoute.ReRoute.CacheOptions.Region);
 
             _logger.LogDebug("finished response added to cache for {downstreamUrlKey}", downstreamUrlKey);
         }
+
+		internal Task<HttpResponseMessage> CreateHttpResponseMessage(CachedResponse cached)
+		{
+			return Task.Run(() =>
```
TomPallister: Why are we spinning up a task here? I dont think we need the overhead

test/Ocelot.AcceptanceTests/Startup.cs
```diff
@@ -43,5 +44,24 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
 
             app.UseOcelot().Wait();
         }
-    }
+	}
+
+	public class StartupWithCustomCacheHandle : Startup
```
TomPallister: This is cool I should do this more!!

src/Ocelot/Cache/Middleware/OutputCacheMiddleware.cs
```diff
@@ -47,7 +48,8 @@ public class OutputCacheMiddleware : OcelotMiddleware
             {
                 _logger.LogDebug("cache entry exists for {downstreamUrlKey}", downstreamUrlKey);
 
-                SetHttpResponseMessageThisRequest(cached);
+				var response = await CreateHttpResponseMessage(cached);
```
TomPallister: spacing seems a little off! Might just be github

TomPallister (changes requested):
Thank you very much for the contribution! Looks good. Please can you check my comments and see what you think?

src/Ocelot/Cache/CachedResponse.cs
hampos: yep, sure no problem!

src/Ocelot/Cache/Middleware/OutputCacheMiddleware.cs
hampos: No it's not github, I probably have different settings on my visual studio and messes things up.
What are your settings on TextEditor -> C# -> Tabs?

src/Ocelot/Cache/Middleware/OutputCacheMiddleware.cs
hampos: Yes you're right, I originally put the Task in the signature to be uniform with CreateCachedResponse, then thought to go premature on it and spinned a task because it does a Base64 conversion.
Should I drop the Task completely? Or return a Task.FromResult() for uniformity?

src/Ocelot/Cache/Middleware/OutputCacheMiddleware.cs
hampos: Maybe we can add a .editorconfig to enforce specific identation settings on the solution?

TomPallister: Yeh good idea!

TomPallister: i use vscode for Ocelot so not sure what I'm using...tbh im a shitty developer who doesnt care about tabs and spaces :E

src/Ocelot/Cache/Middleware/OutputCacheMiddleware.cs
TomPallister: I would just drop the task!

src/Ocelot/Cache/Middleware/OutputCacheMiddleware.cs
TomPallister: Do you have anything you would recommend for a default .editorconfig?

TomPallister (approved):

TomPallister:
@hampos looks good. Are you happy to merge it?

hampos:
Yes if you're ok I'm good for the merge!

Issue #147, opened 2017-11-09
Ocelot doesnt response env in configuration.json


TomPallister:
see file configuration repository

@sharpn spotted this at work! Need to support configuration.dev.json etc the problem is in the FileConfigurationRepository because it hard codes configuration.json in there...sigh at me.

hemantkd:
Hi Tom,
I have a lot to learn but I'd still like to contribute in small ways.

TomPallister:
@hemantkd all contributions are welcome! Would you like some more information about this issue?

dbarkwell:
Are you thinking merge/override configuration files like appsettings or a single configuration based on environment?

TomPallister:
I'm not sure at the moment, will have a think!

hemantkd:
How about something of this form:
        public Response<FileConfiguration> Get(string enviromentName)
        {
            var configFilePath = $"{AppContext.BaseDirectory}/configuration.{enviromentName}.json";
			.
			.
			.
        }
Similarly for the Set method in the FileConfigurationRepository.

Would this work, to at least give an environment based support?

TomPallister:
@hemantkd yep I think it might work! Give it a try if you have time.

dbarkwell:
Could also pass ihostingenvironment into the class and get the environmentname property.

hemantkd:
@TomPallister Cool. Will give it a go as soon as I find some time.
The related unit tests could use some refactoring before I make any changes to the code, I think.

hemantkd:
@dbarkwell not sure if need the dependency on ihostingenvironment when all we will be using really is just one property from it?

dbarkwell:
@hemantkd you'd have to change the interface to implement your solution. would be easier to make the change on the concrete class. just my $0.02.

hemantkd:
@TomPallister what do you say? :)

TomPallister:
try both and see what you prefer..

My guess is ihostingenvironment passed into the constructor would be better and follow the standards.

hemantkd:
Sure, I'm here to play and learn.
Can you please shed some light on what standard is that specifically so that I can read about it?
Trying to unlearn the bad ways and learn the good ways. :)

TomPallister:
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/environments

and then you can use the ihostingenvironment  to get the current environment and append it to any configuration files you want to use! such as appsettings!

hemantkd:
[...]

PR #155, opened 2017-11-16
Initial draft for the issue 147 fix

hemantkd:
Hi @TomPallister,

I have hopefully followed the process that you mentioned correctly.
Have a look at the code fixes for the issue #147 

Thanks for your patience!

### DIFF
--- a/src/Ocelot/Configuration/Repository/FileConfigurationRepository.cs
+++ b/src/Ocelot/Configuration/Repository/FileConfigurationRepository.cs
@@ -8,41 +8,41 @@ namespace Ocelot.Configuration.Repository
 {
     public class FileConfigurationRepository : IFileConfigurationRepository
     {
-        private readonly IHostingEnvironment _hostingEnvironment;
+        private readonly string _configFilePath;
 
         private static readonly object _lock = new object();
 
         public FileConfigurationRepository(IHostingEnvironment hostingEnvironment)
         {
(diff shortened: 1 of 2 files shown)


hemantkd:
Thanks for letting me @TomPallister! :)

Yes, these questions were also in my chain of thoughts.

If _hostingEnvironment.EnvironmentName is null then string interpolation handles it and just uses an empty string. So, the file name becomes configuration..json and everything still works.

The Unit Tests don't have that scenario at the moment.

TomPallister:
Should it be configuration.json not configuration..json?

I think we should have a test for this!

hemantkd:
configuration..json is still a valid file name but I can understand that it looks odd. I can make the change so that it falls back to configuration.json instead and will write a test for it. Thanks.

hemantkd:
Hi @TomPallister,

Let me know if you're happy with the tests.
I tried to keep them similar to the existing ones and created some private methods to re-use some existing code.
I have also changed the code so that if the environment name is null or empty, the file name used is `configuration.json` instead of `configuration..json`
Do you think we need to worry about trimming the whitespaces from the name?

Let me know your thoughts. Thanks

TomPallister:
@hemantkd looks OK to me...I'm not too worried about trimming whitespace. If you are happy to merge it then I am?

hemantkd:
@TomPallister I'm happy too

Issue #176, opened 2017-12-13
Allow Ocelot to store public configuration in Cluster


TomPallister:
I'm putting this here because I had a thought on the train. This potentially removes the need for consul storing the FileConfiguration. However fully understand sticking with consul as it is probably way better than my raft implementation and does a lot of other stuff!

**Happy path**
User uses administration API to manage Ocelot
Configuration is stored in Cluster
Backed by raft

PR #182, opened 2017-12-22
Raft round 2

TomPallister:
So I think my RAFT implementation works and I'm now pulling it into Ocelot.


TomPallister:
fixed #176

PR #200, opened 2018-01-16
Can't set StatusCode after Response has started

kinglionsoft:
As same middleware (like [ClientRateLimitMiddleware](https://github.com/TomPallister/Ocelot/blob/develop/src/Ocelot/RateLimit/Middleware/ClientRateLimitMiddleware.cs#L124)) may has set StatusCode,  resetting StatusCode by ExceptionHandlerMiddleware causes exception:
```
System.InvalidOperationException: Status code cannot be set, response has already started.
```


TomPallister:
@kinglionsoft thanks for the contribution much appreciated.

Looks good to me I will merge!

PR #203, opened 2018-01-18
fix/#98 - command line build on mac

stuartsaltzman:
- Cake v0.25.0, fixes:
-- ToDictionary on Mono causes "The type 'Dictionary<,>' is defined in an assembly that is not referenced"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -250,5 +250,5 @@ _build/
 _static/
 _templates/
 
-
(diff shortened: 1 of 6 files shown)


TomPallister:
@stuartsaltzman thanks for the PR and interest in the project! I will review this ASAP! I'm doing a hackathon at the moment so time is precious!! :)

PR #215, opened 2018-02-01
fix: LF for linux scripts (.sh)

stuartsaltzman:
- change line breaks to LF for unix scripts (\*.sh)


TomPallister:
@stuartsaltzman thanks! Sorry my bad I will remember this.

PR #216, opened 2018-02-01
Feature/issue 209 upstream host based routing

### LINKED ISSUE #209: How to support hostname on upstream path template?
Hi, just wonder how to support hostname on upstream path template since my case we have more than one domain point to same server. 

Thanks

TomPallister:
Fix #209


darting:
wow it is really cool!

Issue #214, opened 2018-02-01
Problem when routing calls to an API that has no response body.


janhabelvik:
Hi
I try to forward calls to a method in an  Web API method that has a "void" return type. What happens is that Ocelot hangs indefinitely. It seems to occur for any verb. Calling the method directly works fine and responds with HTTP/200.

Is this an issue with my hosting or configuration, or could be a another problem.

My Ocelot Gateway Implementation runs under IIS, and calling a method returning a response body works fine.

Thanks

TomPallister:
@janhabelvik I have no idea why this is happening :( I will look into it. I assume you have a controller that returns void? It might be something to do with the fact is doesn't have a response body.

TomPallister:
@janhabelvik also how are you hosting Ocelot? IIS, Kestrel etc?

janhabelvik:
I'm hosting in IIS, but the issue seems to happen independent of the hosting environment.

I downloaded the source and did some debugging, the processing seems to hang on ResponderMiddleware.Invoke on the line:  await _responder.SetResponseOnHttpContext(context, HttpResponseMessage);

This code seems to get stuck when the content.Length is 0. I guess the copy step can be skipped if it's the case.
            using (Stream stream = new MemoryStream(content))
            {
                if (response.StatusCode != HttpStatusCode.NotModified)
                {
                    await stream.CopyToAsync(context.Response.Body);
                }
            }

TomPallister:
@janhabelvik thanks for looking into this. If you want you can submit a PR with the change?

Otherwise I will fix it asap! :D

janhabelvik:
This little fix resolved the issue for me:
            using (Stream stream = new MemoryStream(content))
            {
                if (response.StatusCode != HttpStatusCode.NotModified
                    **&& context.Response.ContentLength != 0**)
                {
                    await stream.CopyToAsync(context.Response.Body);
                }
            }

TomPallister:
fixed in 3.0.2

PR #218, opened 2018-02-02
fix for #214 and some tests for the class it was in but not for the e…

TomPallister:
…rror (cant be arsed to test as have to spin up IIS)


TomPallister:
fixes #214

PR #225, opened 2018-02-10
added 'DownstreamHealthcheck' parameter to FileConfiguration again!

JamesYing:
omg, it's so hard to me!
@TomPallister  please check again, sorry my bro, it's my first pull request and this is a great project！

Next step, I want added a HealthcheckMiddle in project, but I think I haven't a good idea for when check?If check downstream host at middle so it will check action at every request, it's not a good performace.I want add policy handler and cachemanage for health check.

Do you have a idea?


TomPallister:
@JamesYing I'm not sure why you want to add a property but not add the functionality. Maybe wait until your feature is finished before making a pull request :)

I would say why do you want to add this functionality? What would you do if a service failed it's health check? What would happen if there is only one service etc? Also what url on the downstream service would return the health check! I'm not saying this isn't a good idea! It is something that would be nice but it's quite a lot of work! :)

Ocelot already has code that handles retries, docs here http://ocelot.readthedocs.io/en/latest/features/qualityofservice.html

Ocelot integrates with consul for service discovery and I would suggest that rather than adding these features to Ocelot it would make more sense to use Consul for them.

However if this is something you want to add. The services are returned by an interface called IServiceDiscoveryProvider and the type of provider is chosen in ServiceDiscoveryProviderFactory. At the moment Ocleot will return ConfigurationServiceProvider or ConsulServiceDiscoveryProvider. You would need to hook into ConfigurationServiceProvider to do health checks. Maybe a class that polls in the background and doesn't return a service if it is available.

JamesYing:
@TomPallister Thanks for your answer.
In this version, I saw the ocelot added a new configuration property for loadbalancer, when I added more host to configuration and some one has a unkown error(may be is a network problem), at moment the ocelot always request the bad one.Other, if we have health check property, we can add a web view for host manage.
Any way, I'll add  full health check feature to my branch first and then pull it again.

TomPallister:
@JamesYing sounds good! :)

geffzhang:
Healthcheck does not exist before the introduction of independent host configuration, it must rely on Consul, itself has Healthcheck, this feature is very good

TomPallister:
@JamesYing I think the best place for the demos is in 

./samples

This seems to be a standard way of doing things. I am going to add a service fabric sample there soon!

JamesYing:
ok, Tom, Now the branch is test version, I will create new branch for sample

JamesYing:
I close the pr now

TomPallister:
@JamesYing OK cool! Thanks for your help :D

PR #242, opened 2018-02-21
Make Ocelot work with service fabric DNS and naming service for guest exe and stateless

### LINKED ISSUE #238: Support ServiceFabric
What are the steps to configure (Program.cs, Startup.cs, etc.) Ocelot in a Service Fabric stateless service for an on premise deployment?

TomPallister:
doesnt fix


TomPallister:
Needs to check use service disxovery is true and provider name..

Issue #246, opened 2018-02-26
issues about bufferflay 


axzxs2001:
I found a problem with the butterfly test tracking.
I created a new Ocelot Gateway project with a port of 5000, two API items, 5001 and 5002 ports respectively, and in the Configuration.json file for the Ocelot Gateway project, two API project Rerouts settings are " Httphandleroptions ': {' usetracing ': true}.
When I first visited http://localhost:5001/api5001/values, it was normal, and when I visited Http://localhost:5002/api5002/values again, Returns an error with an HTTP status code of 500. When I first visited http://localhost:5002/api5002/values, it was normal. So always the second request returns an HTTP status code of 500 error.
Then I tracked it with Ocelot source code and found an exception in Createhttpmessagehandler method of Httpclientbuilder Clss, Delegatinghandler.innerhandler can be assigned two times on the first request, and only if the Delegatinghandler.innerhandler equals null when assignment, the second assignment will be an error, I changed the code to
if (Delegatinghandler.innerhandler = null)
{
Delegatinghandler.innerhandler = Httpmessagehandler;
}
The problem is solved, but I don't know if it affects other features.

TomPallister:
@axzxs2001 thanks for finding this and letting me know!

Do you think you could submit a PR with the fix? I will check to see if it effects other features.

If you cannot don't worry I will take a look at it myself when I have time.

axzxs2001:
I have submitted a PR

TomPallister:
@axzxs2001 thank you very much!!

TomPallister:
@geffzhang any ideas on this? Is it OK? All tests are passing. 

I will have a look this morning on my way to work.

geffzhang:
@TomPallister This is written on the blog（ http://www.cnblogs.com/axzxs2001/p/8478340.html） and you can look at his examples https://github.com/axzxs2001/Asp.NetCoreExperiment/tree/master/Asp.NetCoreExperiment/OcelotAndButterfly

TomPallister:
@geffzhang thank you! I would like to write a failing test before merging fix.

TomPallister:
@geffzhang @axzxs2001 

I think I have broken tracing with butterfly when I merged the request aggregation feature. I will look into this and fix ASAP. I will also write some tests for the feature so it does not happen again!

TomPallister:
[...]

PR #247, opened 2018-02-26
Update HttpClientBuilder.cs

axzxs2001:
Resolves the issue  #246  of accessing the Second API Project report HTTP status 500 when using butterfly tracing.


axzxs2001:
#246

Issue #260, opened 2018-03-06
CI - Reduce the number of build warnings


binarymash:
## New Feature

Reduce the number of build warnings - work ongoing from #253

## Motivation for New Feature

We want to minimise the number of build warnings, enforce a common coding style and identify any common gotchas in the code

## Steps to Reproduce the Problem

Do a build ;)

## Specifications

n/a

TomPallister:
@binarymash much appreciated, just had a quick look! Only a few things that i commented on and they are my mistakes! Would be awesome if you have time to fix it :p you thought you got away from tidying up my shit code!

TomPallister:
@binarymash thanks for merging chief. Was just in the water poet near Liverpool Street. Was a Huddersfield exciles sticker in the bogs. Exciting.

simonjefford:
Warnings now 0...

Used a pragma to squash the warning about `AddBaseUrl` being obsolete as it was test code (and setup code at that). Also, I switched off package analysis to shut up the warning about the summary text (see https://github.com/NuGet/Home/issues/4587 for more).

TomPallister:
@simonjefford awesome!! Thanks 👍 

AddBaseUrl is some crap that wasn't very well thought out :( I need to get rid of it but focusing on other features at the moment. We have an issue for it!

TomPallister:
@binarymash not sure if you want to re-open this but will leave closed for now as @simonjefford (a friend of mine from real life..yes I have two of them!) has helped out here!

Issue #253, opened 2018-03-02
Static code analysis


binarymash:
Now we're getting a larger number of contributors it might be useful to enforce a few rules around code style. 

I use https://github.com/DotNetAnalyzers/StyleCopAnalyzers on most of my projects, which uses Roslyn analyzers to flag up violations at build time and, if you're using Visual Studio, when you're editing the code.

If we do implement this I want to avoid making it a painful experience for contributors... many of the rules are contentious, so I'd suggest that we start off with most rules turned off and then add in whatever we think will be useful to improve the code quality. 

Roslyn analyzers are not currently supported in vscode (though it looks like there is progress -
 see https://github.com/OmniSharp/omnisharp-vscode/issues/43) so anyone using vscode would only see the failures or warnings when building on the command line, which adds a lot of friction to the dev experience.

I've pushed up a work in progress which has most of the rules turned off - see https://github.com/binarymash/ocelot/tree/feature/AddStyleCopAnalyzers - and in the meantime, we can argue about whether `using` statements should be in alphabetical order ;)

TomPallister:
@binarymash looks good to me. Maybe everything should be a warning instead of none then we can get to fixing them. Once we are at 0 warnings then we can turn them to errors? Not sure what best approach is.

Submit a PR and let's get cracking! I'd love to improve the quality of the code.

binarymash:
Good stuff, I'll try to find some time this weekend. 

I think it'll be cleaner to disable all the violations, then for a particular rule turn it on, fix and commit, and then move on to the next rule. Turning them all to warnings will just generate a *lot* of noise (try building https://github.com/binarymash/Ocelot/commit/df86052a31235d4dd740020dc8ddf574b3b1b81a ...)

I'm very much of the opinion that you shouldn't get any build warnings if you can help it... we've all worked on code that generates 500 warnings, and you just start to ignore them, and the code deteriorates. If there's a good reason for violating a rule, then in those cases you should wrap the code in a `#pragma warning disable` - `#pragma warning restore`

TomPallister:
@binarymash yep makes sense! Agree with everything.

PR #261, opened 2018-03-06
Feature/reduce build warnings

binarymash:
Fixes / New Feature #

#260

## Proposed Changes

Fixes all build warnings except for...

1. The warnings relating to package analysis. These are because of limitations in the current dotnet core package production infrastructure - see https://github.com/NuGet/Home/issues/4587. Not sure what the status is for the fix of this. We could disable package analysis, but that might hide other problems, so suggest we leave it as is for now.

2. CS0618 (using obsolete code) 

I've also enabled errors for some of the rules, which I forgot to do as part of #253. The rules I've chosen are a subset of the rules I use on my own projects. When I turned on errors for these rules it did not result in any errors in the code, so I'm taking that as proof that these rules are pretty unobtrusive and won't cause much hassle for devs unless they are really following bad practices.

### DIFF
--- a/src/Ocelot/Configuration/Repository/FileConfigurationRepository.cs
+++ b/src/Ocelot/Configuration/Repository/FileConfigurationRepository.cs
@@ -18,7 +18,7 @@ namespace Ocelot.Configuration.Repository
             _configFilePath = $"{AppContext.BaseDirectory}/configuration{(string.IsNullOrEmpty(hostingEnvironment.EnvironmentName) ? string.Empty : ".")}{hostingEnvironment.EnvironmentName}.json";
         }
 
-        public async Task<Response<FileConfiguration>> Get()
+        public Task<Response<FileConfiguration>> Get()
         {
             string jsonConfiguration;
 
@@ -29,10 +29,10 @@ namespace Ocelot.Configuration.Repository
 
             var fileConfiguration = JsonConvert.DeserializeObject<FileConfiguration>(jsonConfiguration);
 
-            return new OkResponse<FileConfiguration>(fileConfiguration);
+            return Task.FromResult<Response<FileConfiguration>>(new OkResponse<FileConfiguration>(fileConfiguration));
         }
 
-        public async Task<Response> Set(FileConfiguration fileConfiguration)
+        public Task<Response> Set(FileConfiguration fileConfiguration)
         {
             string jsonConfiguration = JsonConvert.SerializeObject(fileConfiguration);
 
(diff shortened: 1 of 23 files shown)


src/Ocelot/Configuration/Repository/FileConfigurationRepository.cs
```diff
@@ -18,7 +18,7 @@ public FileConfigurationRepository(IHostingEnvironment hostingEnvironment)
             _configFilePath = $"{AppContext.BaseDirectory}/configuration{(string.IsNullOrEmpty(hostingEnvironment.EnvironmentName) ? string.Empty : ".")}{hostingEnvironment.EnvironmentName}.json";
         }
 
-        public async Task<Response<FileConfiguration>> Get()
+        public Task<Response<FileConfiguration>> Get()
```
TomPallister: This method we should probably make 

```csharp
jsonConfiguration = System.IO.File.ReadAllText(_configFilePath);
```

into
```csharp
jsonConfiguration = System.IO.File.ReadAllTextAsync(_configFilePath);
```

Then we dont need the Task.FromResult

src/Ocelot/Configuration/Repository/FileConfigurationRepository.cs
```diff
@@ -45,8 +45,8 @@ public FileConfigurationRepository(IHostingEnvironment hostingEnvironment)
 
                 System.IO.File.WriteAllText(_configFilePath, jsonConfiguration);
```
TomPallister: Same as above make this WriteAllTextAsync and we dont need Task.FromResult

TomPallister:
@binarymash sorry took me ages to look at this, been super busy with the new job!

binarymash:
No worries, I'll take a look at this. Hope the new job is going ok, and you're not missing the Pig too much ;)

binarymash:
Ok, this is a bit trickier than the 2 minute job I thought it would be because you can't use await within a lock :)

I'm assuming there's a lock there to prevent concurrent access to the config file, so I looked at locking the file and accessing the filestream via File.Open(...). But, in doing this I see random test failures around accessing the config files. Pretty sure I've seen these failures before, so not sure if it's anything to do with my changes. However, this highlights that changing to use this approach rather than using the sync lock changes the behaviour slightly - we'll now get an exception on concurrent access. Do we ever expect this to happen at runtime? If so, what should we do?

And, what assumptions can we make about the lifetime of the config file? Will it always already exist before we try to write to it? I don't know the operation of Ocelot well enough to understand the exact behaviour here.

Might be worth splitting this into its own issue. What do you reckon?

TomPallister:
@binarymash sorry that is about the 20th time I’ve forgotten about that lock. Feel free to merge!

PR #267, opened 2018-03-07
Map content specific headers properly

### LINKED ISSUE #263: Content-Encoding headers are not preserved
## Actual Behavior
Downstream requests from Ocelot do not contain Content-Encoding headers even if upstream requests do. This causes rerouted gzip-encoded requests to fail.

## Analysis
Headers are mapped in [`RequestMapper.MapHeaders`](https://github.com/ThreeMammals/Ocelot/blob/a53338312bcc1ca99a6ee901e35ac24b07079d03/src/Ocelot/Request/Mapper/RequestMapper.cs#L72), but they are all added to the downstream request by calling `requestMessage.Headers.TryAddWithoutValidation`. This doesn't work for Content-Encoding,
 though, because .NET understands it as a [content header](https://github.com/dotnet/corefx/blob/3e72ee5971db5d0bd46606fa672969adde29e307/src/System.Net.Http/src/System/Net/Http/Headers/KnownHeaders.cs) (so it should be added via `requestMessage.Content.Headers.TryAddWithoutValidation`).

`RequestMapper` should be updated to properly forward content headers like Content-Encoding and Content-Language. Content-Type is already handled in [`RequestMapper.MapContent`](https://github.com/ThreeMammals/Ocelot/blob/a53338312bcc1ca99a6ee901e35ac24b07079d03/src/Ocelot/Request/Mapper/RequestMapper.cs#L50). Perhaps that method could be updated to check for other content headers.

  1. Attempt to re-route a request with compressed contents (and a gzip Content-Encoding header)
  1. Notice that the downstream request is missing the header and, consequently, the request fails.

## Specifications

  - Version: 3.1.7
  - Platform: Windows 10 x64

TomPallister:
## Fixes #263 
- #263 

## Proposed Changes
  - map all content specific headers to downstream request content property, make sure we dont map them to request specific headers, added a gzip encoding acceptance test

### DIFF
--- a/src/Ocelot/Request/Mapper/RequestMapper.cs
+++ b/src/Ocelot/Request/Mapper/RequestMapper.cs
@@ -47,9 +47,26 @@
 
             var content = new ByteArrayContent(await ToByteArray(request.Body));
 
-            content.Headers.TryAddWithoutValidation("Content-Type", new[] {request.ContentType});
+            content.Headers
+                .TryAddWithoutValidation("Content-Type", new[] {request.ContentType});
 
+            AddHeaderIfExistsOnRequest("Content-Language", content, request);
+            AddHeaderIfExistsOnRequest("Content-Location", content, request);
+            AddHeaderIfExistsOnRequest("Content-Range", content, request);
+            AddHeaderIfExistsOnRequest("Content-MD5", content, request);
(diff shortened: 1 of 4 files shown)


mjrousos (approved):

mjrousos:
Looks good; thanks for the quick fix!

PR #279, opened 2018-03-16
When using cookie container cookies are per request, not shared.

### LINKED ISSUE #274: Investigate - Cookies are not thread safe when using CookieContainer
I have found a Bug about cookie in Ocelot ; let me describe it : I have two pc，both of them will request the  ocelot which deployed in cloud server and both of them  are in the same LAN ; then the website behind the  ocelot  will get the   cookie2 from pc1's request;  in fact the cookie2 belongs to pc2; how can i sove this problem?

TomPallister:
Fixes / New Feature #

Still work in progress
fixes #274 

## Proposed Changes

  - added acceptance tests, need to work out failing unit tests but probably going to be a redesign where we hold a reference to the cookie container and empty it if needed
  -
  -


TomPallister:
close not possible

PR #301, opened 2018-04-05
Override ToString() in DownstreamRequest

FelixBoers:
When executing the downstreamUrlCreatorMittleware the downstream request
url shoud be written to the log. But instead of the url the type name
gets written because the ToString() method wasn't overriden.

Now ToString() internally calls the ToUri() method in order to provide
the url instead of the type name.


TomPallister:
@FelixBoers thanks :) really appreciate the help!

PR #302, opened 2018-04-05
Log downstream templates in middleware

FelixBoers:
Log downstream templates in DownstreamRouteFinderMiddleware.
Concatenate all downstream templates separated by , (colon) in order to write them to the log.


TomPallister:
@FelixBoers thanks for the help! :)

PR #303, opened 2018-04-06
Feature/another look at logging

### LINKED ISSUE #287: New Feature - Provide more detailed log
Hello ,  I encountered some difficulties in configuring error lookups.

Can you provide more detailed content in the logging content?  like the request URL and current error configuration.

This is the current logging content :
```
requestId: no request id, previousRequestId: no previous request id, message: DownstreamRouteFinderMiddleware setting pipeline errors. IDownstreamRouteFinder returned Error Code: UnableToFindDownstreamRouteError Message: UnableToFindDownstreamRouteError
```
and

```
requestId: no request id, previousRequestId: no previous request id, message: 1 pipeline errors found in ResponderMiddleware. Setting error response status code
```

When I have a lot of configuration, it's hard to find out which configuration item caused the error.

Thanks~!

TomPallister:
Fixes / New Feature #

fix #287 
Work in progress

## Proposed Changes

  - refactor logger to only take strings as currently takes params but this is misleading as we just turn them into strings anyway and you dont get structured logging
  - add log methods for structured logging with request ids
  - log errors when they happen rather than at the end of the pipeline, this is misleading the flow
- use asp.net request id if one is not set, i notice in the logs there is never a request id


f2calv:
@TomPallister just wondering why the structured logging has taken a step backwards and is now back to string interpolation? Has there been a performance issue?

PR #317, opened 2018-04-17
Update middlewareinjection.rst

Burnaster:
Class name "OcelotMiddlewareConfiguration" does not exist and "UseOcelot" extension method expects a "OcelotPipelineConfiguration" so i think this documentaiton is outdated.

Fixes#
Documentation update.
## Proposed Changes

  - Example / Snipped referencing the wrong class name.


TomPallister:
@Burnaster thanks for the PR, I will merge now :)

Issue #330, opened 2018-04-25
Incorrect delegate handler being called on request via reroute


Dilsy99:
**Expected**
When attaching a delegate handler I would expect that the correct handler is called based on the reroute request.

**Actual**
When you have two reroutes that have the same upstream url with different verbs that have different delegate handlers attached to them, the first "hit" to one of the upstream url works correctly but when trying to hit the other url (with a different verb), the downstream route is correctly hit but the handler called is the one attached to the other route.

Example config:

{
      "DownstreamPathTemplate": "/api/ClientQuery",
      "DownstreamScheme": "http",
      "DownstreamHostAndPorts": [
        {
          "Host": "localhost",
          "Port": 52709
        }
      ],
      "UpstreamPathTemplate": "/customer",
      "UpstreamHttpMethod": [ "Get" ],
      "DelegatingHandlers": [
        "ClientDefaultHandler"
      ]
    },
{
      "DownstreamPathTemplate": "/api/ClientCommand",
      "DownstreamScheme": "http",
      "DownstreamHostAndPorts": [
        {
          "Host": "localhost",
          "Port": 52709
        }
      ],
      "UpstreamPathTemplate": "/customer",
      "UpstreamHttpMethod": [ "Post" ],
      "DelegatingHandlers": [
        "ClientAddedHandler"
      ]
    },

Call "Get" /customer it will invoke the ClientDefaultHandler
Call "Post" /customer it will also invoke ClientDefaultHandler rather than CLientAddedHandler as expected

**Investigation**
The issue is caused because the cache key used in the HttpClientBuilder does not take the verb into account so requests to the url from any verb will use the same httpclient (and associated handlers) as the "first" request.

A suggested fix is to include the http verb in the cache key so different verbs to the same url will use there own httpclient (and handlers)

```cs
private string GetCacheKey(DownstreamContext request)
        {
            var cacheKey = $"{request.HttpContext.Request.Method}:{request.DownstreamRequest.OriginalString}";

            return cacheKey;
        }
```

***Sorry this is my first time contributing to this type of project but I am happy to make this change and submit a pull request if this is acceptable. 

Many thanks

## Specifications

  - Version: 5.5.5
  - Platform: Windows 10
  - Subsystem: Net Core 2.0 C# Visual Studio 2017

TomPallister:
@Dilsy99 wow this is awesome investigation! Thank you very much. Please submit a PR I will happily accept one :)

Thanks!! 😃

[...]

PR #331, opened 2018-04-25
Develop

Dilsy99:
Fixes # 330

#330 Fix issue with httpclient being cached based on url but not including the http verb. This caused issues because handers were being called on incorrect routes because the httpclient was being shared by multipe routes. Added the http verb to the cache key and added unit test


TomPallister:
@Dilsy99 this is really good! Thanks 😄

PR #367, opened 2018-05-21
made http client work under full .net 46

### LINKED ISSUE #366: Ocelot on full .Net framework 4.6.1 fails to create HttpClient and use

In VS 2017 create .Net Core project but select (at the second screen) .NET Framework, not .Net Core one.
Select in the Project -> properties 4.6.1 framework.
Configure Ocelot and run.
An attempt to get into a route will throw.

## Specifications

  - Version: Win 10
  - Platform: .NET 4.6.1
  - Subsystem:

And here is the code to fix it:

<pre>
--- a/src/Ocelot/Requester/HttpClientBuilder.cs
+++ b/src/Ocelot/Requester/HttpClientBuilder.cs
@@ -42,15 +42,29 @@ namespace Ocelot.Requester
             {
                 return httpClient;
             }
-
-            var httpclientHandler = new HttpClientHandler
+            bool useCookies = context.DownstreamReRoute.HttpHandlerOptions.UseCookieContainer;
+            HttpClientHandler httpclientHandler;
+            // Dont' create the CookieContainer if UseCookies is not set ot the HttpClient will complain
+            // under .Net Full 46
+            if (useCookies)
             {
-                AllowAutoRedirect = context.DownstreamReRoute.HttpHandlerOptions.AllowAutoRedirect,
-                UseCookies = context.DownstreamReRoute.HttpHandlerOptions.UseCookieContainer,
-                CookieContainer = new CookieContainer()
-            };
+                httpclientHandler = new HttpClientHandler
+                {
+                    AllowAutoRedirect = context.DownstreamReRoute.HttpHandlerOptions.AllowAutoRedirect,
+                    UseCookies = context.DownstreamReRoute.HttpHandlerOptions.UseCookieContainer,
+                    CookieContainer = new CookieContainer()
+                };
+            }
+            else
+            {
+                httpclientHandler = new HttpClientHandler
+                {
+                    AllowAutoRedirect = context.DownstreamReRoute.HttpHandlerOptions.AllowAutoRedirect,
+                    UseCookies = context.DownstreamReRoute.HttpHandlerOptions.UseCookieContainer,
+                };
+            }

-            if(context.DownstreamReRoute.DangerousAcceptAnyServerCertificateValidator)
+            if (context.DownstreamReRoute.DangerousAcceptAnyServerCertificateValidator)
             {
                 httpclientHandler.ServerCertificateCustomValidationCallback = (request, certificate, chain, errors) => true;

diff --git a/src/Ocelot/Requester/HttpClientHttpRequester.cs b/src/Ocelot/Requester/HttpClientHttpRequester.cs
index c7914c9..60d4899 100644
--- a/src/Ocelot/Requester/HttpClientHttpRequester.cs
+++ b/src/Ocelot/Requester/HttpClientHttpRequester.cs
@@ -32,7 +32,17 @@ namespace Ocelot.Requester

             try
             {
-                var response = await httpClient.SendAsync(context.DownstreamRequest.ToHttpRequestMessage());
+                var message = context.DownstreamRequest.ToHttpRequestMessage();
+
+                // Remove Content from the message if not put or post or the http client will choke
+                // under .Net Full 46
+                if (message.Method != HttpMethod.Post &&
+                    message.Method != HttpMethod.Put)
+                {
+                    message.Content = null;
+                }
+                _logger.LogDebug(string.Format("Sending {0}", message));
+                var response = await httpClient.SendAsync(message);
                 return new OkResponse<HttpResponseMessage>(response);
             }
             catch (TimeoutRejectedException exception)
</pre>

tsirkin:
Fixes #366

### DIFF
--- a/src/Ocelot/Requester/HttpClientBuilder.cs
+++ b/src/Ocelot/Requester/HttpClientBuilder.cs
@@ -45,7 +45,7 @@ namespace Ocelot.Requester
             bool useCookies = context.DownstreamReRoute.HttpHandlerOptions.UseCookieContainer;
             HttpClientHandler httpclientHandler;
             // Dont' create the CookieContainer if UseCookies is not set ot the HttpClient will complain
-            // under .Net Full 46
+            // under .Net Full Framework
             if (useCookies)
             {
                 httpclientHandler = new HttpClientHandler
--- a/src/Ocelot/Requester/HttpClientBuilder.cs
+++ b/src/Ocelot/Requester/HttpClientBuilder.cs
@@ -42,15 +42,29 @@ namespace Ocelot.Requester
             {
                 return httpClient;
             }
-
-            var httpclientHandler = new HttpClientHandler
+            bool useCookies = context.DownstreamReRoute.HttpHandlerOptions.UseCookieContainer;
+            HttpClientHandler httpclientHandler;
+            // Dont' create the CookieContainer if UseCookies is not set ot the HttpClient will complain
+            // under .Net Full 46
+            if (useCookies)
             {
-                AllowAutoRedirect = context.DownstreamReRoute.HttpHandlerOptions.AllowAutoRedirect,
-                UseCookies = context.DownstreamReRoute.HttpHandlerOptions.UseCookieContainer,
-                CookieContainer = new CookieContainer()
-            };
+                httpclientHandler = new HttpClientHandler
+                {
+                    AllowAutoRedirect = context.DownstreamReRoute.HttpHandlerOptions.AllowAutoRedirect,
+                    UseCookies = context.DownstreamReRoute.HttpHandlerOptions.UseCookieContainer,
+                    CookieContainer = new CookieContainer()
+                };
+            }
+            else
+            {
+                httpclientHandler = new HttpClientHandler
+                {
+                    AllowAutoRedirect = context.DownstreamReRoute.HttpHandlerOptions.AllowAutoRedirect,
+                    UseCookies = context.DownstreamReRoute.HttpHandlerOptions.UseCookieContainer,
+                };
+            }
 
-            if(context.DownstreamReRoute.DangerousAcceptAnyServerCertificateValidator)
+            if (context.DownstreamReRoute.DangerousAcceptAnyServerCertificateValidator)
             {
                 httpclientHandler.ServerCertificateCustomValidationCallback = (request, certificate, chain, errors) => true;
 
--- a/src/Ocelot/Requester/HttpClientHttpRequester.cs
+++ b/src/Ocelot/Requester/HttpClientHttpRequester.cs
@@ -33,11 +33,17 @@ namespace Ocelot.Requester
             try
             {
                 var message = context.DownstreamRequest.ToHttpRequestMessage();
+                /** According to https://tools.ietf.org/html/rfc7231
+                 * GET,HEAD,DELETE,CONNECT,TRACE
+                 * Can have body but server can reject the request.
+                 * And MS HttpClient in Full Framework actually rejects it.
+                 * see #366 at issue 
+                **/
 
-                // Remove Content from the message if not put or post or the http client will choke
-                // under .Net Full 46
-                if (message.Method != HttpMethod.Post &&
-                    message.Method != HttpMethod.Put)
+                if (message.Method == HttpMethod.Get ||
+                    message.Method == HttpMethod.Head ||
+                    message.Method == HttpMethod.Delete ||
+                    message.Method == HttpMethod.Trace)
                 {
                     message.Content = null;
                 }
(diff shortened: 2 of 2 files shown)


src/Ocelot/Requester/HttpClientHttpRequester.cs
```diff
@@ -32,7 +32,17 @@ public class HttpClientHttpRequester : IHttpRequester
 
             try
             {
-                var response = await httpClient.SendAsync(context.DownstreamRequest.ToHttpRequestMessage());
+                var message = context.DownstreamRequest.ToHttpRequestMessage();
+
+                // Remove Content from the message if not put or post or the http client will choke
```
TomPallister: should this be in a preprocessor directive if the problem only exists in .net 461? You can read more about them [here](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-if)

src/Ocelot/Requester/HttpClientBuilder.cs
```diff
@@ -42,15 +42,29 @@ public IHttpClient Create(DownstreamContext context)
             {
                 return httpClient;
             }
-
-            var httpclientHandler = new HttpClientHandler
+            bool useCookies = context.DownstreamReRoute.HttpHandlerOptions.UseCookieContainer;
+            HttpClientHandler httpclientHandler;
+            // Dont' create the CookieContainer if UseCookies is not set ot the HttpClient will complain
```
TomPallister: should this be in a preprocessor directive if the problem only exists in .net 461? You can read more about them [here](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-if) might make more sense rather than the comment

tsirkin:
I am aware of preprocessor directives.
But I am not sure about the exact versions of .net effected by this.So I am not sure what the exact directive should be.
I have used 462 because what I was doing was incorporating sustainys saml lib into ocelot and that requires 462.
And, I am trying to avoid preprocessing as I think that it breaks code readabiliy and debuging,but that's just me.

TomPallister:
@tsirkin pre processing is pretty typical if you are doing stuff specific to a version of .net.

I'm not 100% because I'm sat on a train with bad internet but I think that net standard 2.0 only works for net461+ so you would just need the directive to account for greater than 461 I guess.

Also what do you mean the httpclient will chose under 461 if you have content and it isnt put or post? That seems weird given that GET supports content since http1 spec :o I havent tested it myself but cant find anything after a quick google search!

tsirkin:
@TomPallister Thanks you for response.

1. I tested Ocelot on 4.7.2 and the problem is still there.
Removing my patch from HttpClientBuilder and running an app I am getting

fail: Ocelot.Errors.Middleware.ExceptionHandlerMiddleware[0]
      requestId: 0HLE0HKRQ0Q8T:00000001, previousRequestId: no previous request id, message: Exception caught in global error handler, exception message: The UseCookies property must be set to 'true' to use this property., exception stack:    at System.Net.Http.HttpClientHandler.set_CookieContainer(CookieContainer value)

fixing this I am getting

warn: Ocelot.Requester.Middleware.HttpRequesterMiddleware[0]
      requestId: 0HLE0HLQJ5UNS:00000001, previousRequestId: no previous request id, message: Error making http request, exception: System.Net.ProtocolViolationException: Cannot send a content-body with this verb-type.
         at System.Net.HttpWebRequest.CheckProtocol(Boolean onRequestStream)
         at System.Net.HttpWebRequest.BeginGetRequestStream(AsyncCallback callback, Object state)
         at System.Net.Http.HttpClientHandler.StartGettingRequestStream(RequestState state)
         at System.Net.Http.HttpClientHandler.PrepareAndStartContentUpload(RequestState state)
      --- End of stack trace from previous location where exception was thrown ---
         at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
         at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
         at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()

2. I looked into using preprocessor directive again.
Sorry, but I can't figure out how *exactly* you want to apply it here.
Is this what you have in mind:
```
#if NET46 || NET461 || NET462 || NET47 || NET471 || NET472
 code here
#endif
```
(Please remember that it have to be a boolean query
[https://docs.microsoft.com/en-us/dotnet/standard/frameworks](https://docs.microsoft.com/en-us/dotnet/standard/frameworks)
[https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-if](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-if))
Also, those are run at pre compilation, as opposite to run time, are you going to ship different nuget packages for different frameworks?
That is possible but it is for you to decide to do this work. 

As for the body in Get, seems to me that this is not the spec's fault, it is the MS http client that forbids it.

What *might* be better is to test against !Get method vs == Post || == Put as there are other http methods like Patch, Webdav and God knows what else.
What do you think?
I am here to help with the patch.

TomPallister:
@tsirkin that all makes sense, I will merge this later!

tsirkin:
done

TomPallister:
@tsirkin looks good, just left one comment let me know what you think.

PR #379, opened 2018-06-01
Hard-code coveralls.net version 

binarymash:
Coveralls.net 1.0.0 seems to only support .net core 2.1, so for now we need to hard-code the version as 0.7.0. We should probably revisit this once Ocelot is updated to 2.1.


binarymash:
?? No idea why coveralls is reporting a drop in test coverage

TomPallister:
@binarymash thanks! I noticed this the other day on a different PR. Bit weird.

Issue #390, opened 2018-06-07
Allow users to set useproxy on httpclient


fablacky:
Although my gateway is quite fast on development, once it gets published, every request takes about 2-3 seconds longer to complete no matter what. I did some digging and apparently it's a very well known issue involving the HttpClient. 
The following url helped me a lot:
[https://github.com/dotnet/corefx/issues/6180](url) 
I edited your HttpClientBuilder set the UseProxy propert to false when instantiating the HttpClientHandler.
![image](https://user-images.githubusercontent.com/17791461/41103735-234185ac-6a6a-11e8-9152-57b80c8c9bd1.png)

As a result all my requests are as fast as on develop environment. Do you think this qualifies as an added entry in the configuration file? 
Thank you so much for your attention and support.

TomPallister:
@fablacky interesting that no one else has run into this problem! I was not aware of it. Is this not because there is some kind network configuration issue as per the last comment in the issue you have linked to this?

Regardless it is probably a valuable option to allow users to configure! I would happily accept a PR for this if you feel like taking the issue on.

TomPallister:
Released in 7.0.5

PR #391, opened 2018-06-09
Add ability to specify whether to UseProxy or not on ReRoutes (#390)

marcoaam:
New Feature #390

Adds the possibility to specify whether to UseProxy or not as mentioned on the issue #390 . I kept the value defaulted to true which is the current default for the HttpHandler and that way the change is less invasive. A follow up feature would be to specify the proxy but that would be separate from this. Hope this is usefull. Let me know your thoughts.

Thanks.

### DIFF
--- a/src/Ocelot/Configuration/Creator/HttpHandlerOptionsCreator.cs
+++ b/src/Ocelot/Configuration/Creator/HttpHandlerOptionsCreator.cs
@@ -18,7 +18,7 @@ namespace Ocelot.Configuration.Creator
             var useTracing = _tracer.GetType() != typeof(FakeServiceTracer) && options.UseTracing;
 
             return new HttpHandlerOptions(options.AllowAutoRedirect,
-                options.UseCookieContainer, useTracing);
+                options.UseCookieContainer, useTracing, options.UseProxy);
         }
     }
 }
--- a/src/Ocelot/Configuration/File/FileHttpHandlerOptions.cs
+++ b/src/Ocelot/Configuration/File/FileHttpHandlerOptions.cs
@@ -6,6 +6,7 @@
         {
             AllowAutoRedirect = false;
             UseCookieContainer = false;
+            UseProxy = true;
         }
 
         public bool AllowAutoRedirect { get; set; }
@@ -13,5 +14,7 @@
         public bool UseCookieContainer { get; set; }
 
         public bool UseTracing { get; set; }
+
+        public bool UseProxy { get; set; }
     }
 }
--- a/src/Ocelot/Configuration/HttpHandlerOptions.cs
+++ b/src/Ocelot/Configuration/HttpHandlerOptions.cs
@@ -6,11 +6,12 @@
     /// </summary>
     public class HttpHandlerOptions
     {
-        public HttpHandlerOptions(bool allowAutoRedirect, bool useCookieContainer, bool useTracing)
+        public HttpHandlerOptions(bool allowAutoRedirect, bool useCookieContainer, bool useTracing, bool useProxy = true)
         {
             AllowAutoRedirect = allowAutoRedirect;
             UseCookieContainer = useCookieContainer;
             UseTracing = useTracing;
+            UseProxy = useProxy;
         }
 
         /// <summary>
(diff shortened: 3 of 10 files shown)


TomPallister:
@marcoaam wow thanks for your contribution! I will review this early next week.

fablacky:
@marcoaam Wow, thank you so much! I was ready to get onto this myself but finding out the job was already done was an amazing surprise! Cheers!

src/Ocelot/Configuration/Creator/HttpHandlerOptionsCreator.cs
```diff
@@ -15,10 +15,10 @@ public HttpHandlerOptionsCreator(IServiceTracer tracer)
 
         public HttpHandlerOptions Create(FileHttpHandlerOptions options)
         {
-            var useTracing = _tracer.GetType() != typeof(FakeServiceTracer) && options.UseTracing;
+            var useTracing = _tracer.GetType() != typeof(FakeServiceTracer) && options.UseTracing;
 
             return new HttpHandlerOptions(options.AllowAutoRedirect,
-                options.UseCookieContainer, useTracing);
+                options.UseCookieContainer, useTracing, options.UseProxy);
```
TomPallister: looks good

src/Ocelot/Configuration/File/FileHttpHandlerOptions.cs
```diff
@@ -5,13 +5,16 @@ public class FileHttpHandlerOptions
         public FileHttpHandlerOptions()
         {
             AllowAutoRedirect = false;
-            UseCookieContainer = false;
+            UseCookieContainer = false;
+            UseProxy = true;
```
TomPallister: 👍

src/Ocelot/Configuration/File/FileHttpHandlerOptions.cs
```diff
@@ -5,13 +5,16 @@ public class FileHttpHandlerOptions
...
         }
 
         public bool AllowAutoRedirect { get; set; }
 
         public bool UseCookieContainer { get; set; }
 
-        public bool UseTracing { get; set; }
+        public bool UseTracing { get; set; }
+
+        public bool UseProxy { get; set; }
```
TomPallister: 👍

src/Ocelot/Configuration/HttpHandlerOptions.cs
```diff
@@ -6,11 +6,12 @@
     /// </summary>
     public class HttpHandlerOptions
     {
-        public HttpHandlerOptions(bool allowAutoRedirect, bool useCookieContainer, bool useTracing)
+        public HttpHandlerOptions(bool allowAutoRedirect, bool useCookieContainer, bool useTracing, bool useProxy = true)
```
TomPallister: Not sure that we need this to default here as we default it in the FileConfiguration but I dont have a big problem with it

src/Ocelot/Configuration/HttpHandlerOptions.cs
```diff
@@ -23,9 +24,14 @@ public HttpHandlerOptions(bool allowAutoRedirect, bool useCookieContainer, bool
...
+        /// <summary>
         /// Specify is handler has to use a opentracing
         /// </summary>
-        public bool UseTracing { get; private set; }
+        public bool UseTracing { get; private set; }
+
+        /// <summary>
+        /// Specify if handler has to use a proxy
+        /// </summary>
+        public bool UseProxy { get; private set; }
```
TomPallister: 👍

src/Ocelot/Configuration/HttpHandlerOptions.cs
marcoaam: I would have also liked to avoid this but the main reason to default it here as well was to make the parameter optional and that way keep the change as small as possible, but I am happy to remove it from here and update all the references (mainly tests, about 25 references) if needed.

TomPallister: @marcoaam if you are happy to do it that would be awesome! I don't mind merging and doing it myself if you don't have the time! Let me know what you prefer.

marcoaam: Thanks for the quick response. No worries @TomPallister, It should only take a minute. I will get that done now.

TomPallister:
@fablacky it is set to true by default so should be fine unless this problem is more subtle!

Can I ask when you manually change the code yourself and build Ocelot what version of .net do you target when you build it?

fablacky:
@TomPallister 
On my environment it targets .NET Standard 2.0.

TomPallister:
@fablacky I have no idea why this doesnt work for you then, you running windows, mac or linux? :(

PR #401, opened 2018-06-13
Feature/investigate rate limiting

### LINKED ISSUE #372: RateLimitCore.ProcessRequest. Is it a bug?
RateLimitCore.ProcessRequest, I think it's a bug. when requests exceed Limit in Period,the next request doesn't have to wait for PeriodTimespan,just wait for a time longer than Period.

**# RateLimitCore.ProcessRequest,  I think it's a bug. when  requests exceed Limit in Period,the next request doesn't have to wait for PeriodTimespan,just wait for a time longer than Period.**

TomPallister:
Fixes / New Feature #
removed periodtimespan because i dont think we need it, just use period 
fixes #372

## Proposed Changes

  -
  -
  -

### DIFF
--- a/docs/features/ratelimiting.rst
+++ b/docs/features/ratelimiting.rst
@@ -13,14 +13,12 @@ OK so to get rate limiting working for a ReRoute you need to add the following j
         "ClientWhitelist": [],  
         "EnableRateLimiting": true,  
         "Period": "1s",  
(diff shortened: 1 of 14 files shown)


stacklesshaha:
why remove periodtimespan? In my opinion, having periodtimespan  is better

Issue #407, opened 2018-06-15
Catch all type route doesn't work as expected on a /api/{everything} reroute


marcoaam:
## Expected Behavior

Route with /anypath/.something (note the *"."* on the path) should work and return the downstream service content.

## Actual Behavior

Using IdentityServer as an example it returns 404 and the downstream service doesn't receive the request when using the configuration specified on the steps to reproduce and requesting /identityserver/.well-known/openid-configuration.

It works and returns the expected content by adding a route like the following:

```json
    {
      "DownstreamPathTemplate": "/.well-known/openid-configuration",
      "DownstreamScheme": "http",
      "DownstreamHostAndPorts": [
        {
          "Host": "localhost",
          "Port": 5000
        }
      ],
      "UpstreamPathTemplate": "/identityserver/.well-known/openid-configuration",
      "UpstreamHttpMethod": [ "Get" ]
    }
```

Also works when adding a route with a catch all without a prefix in the path like:

```json
    {
      "DownstreamPathTemplate": "/{everything}",
      "DownstreamScheme": "http",
      "DownstreamHostAndPorts": [
        {
          "Host": "localhost",
          "Port": 5000
        }
      ],
      "UpstreamPathTemplate": "/{everything}",
      "UpstreamHttpMethod": [ "Get" ]
    }
```

## Steps to Reproduce the Problem

1 - Use IdentityServer as a downstream service and add the following config to the ocelot gateway:

```json
    {
      "DownstreamPathTemplate": "/{everything}",
      "DownstreamScheme": "http",
      "DownstreamHostAndPorts": [
        {
          "Host": "localhost",
          "Port": 5000
        }
      ],
      "UpstreamPathTemplate": "/identityserver/{everything}",
      "UpstreamHttpMethod": [ "Get" ]
    }
```

2 - Hit the /identityserver/.well-known/openid-configuration endpoint.

## Specifications

  - Version: 7.0.5
  - Platform: Windows 10 x64
  - Subsystem: .NET core 2.1

TomPallister:
@marcoaam all good, I will have a look at this and see what's up then feedback!

TomPallister:
released in nuget package 7.0.7 will be available on nuget soon!

raman-m:
@AShaker01 Hi Ahmed!
Please open question discussion in [Discussions](https://github.com/ThreeMammals/Ocelot/discussions) space!

PR #411, opened 2018-06-18
Fix catch all route on UpstreamTemplatePatternCreator regex to match …

marcoaam:
Fixes issue #407

### DIFF
--- a/src/Ocelot/Configuration/Creator/UpstreamTemplatePatternCreator.cs
+++ b/src/Ocelot/Configuration/Creator/UpstreamTemplatePatternCreator.cs
@@ -6,7 +6,7 @@ namespace Ocelot.Configuration.Creator
 {
     public class UpstreamTemplatePatternCreator : IUpstreamTemplatePatternCreator
     {
-        private const string RegExMatchEverything = "[0-9a-zA-Z].*";
+        private const string RegExMatchOneOrMoreOfEverything = ".+";
         private const string RegExMatchEndString = "$";
         private const string RegExIgnoreCase = "(?i)";
         private const string RegExForwardSlashOnly = "^/$";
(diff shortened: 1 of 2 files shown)


TomPallister:
@marcoaam there is a problem with one of the acceptance tests so we need to fix this before we merge the PR.

marcoaam:
Sorry @TomPallister. I thought I did a full run but I must have missed that test. I will have a look.

marcoaam:
@TomPallister I fixed the test but after looking at it more closely I decided to remove it as it doesn't seem to me a valid scenario. This route should not return a 404 as it is not any different from the other 2 tests in that file like "should_return_response_200_with_forward_slash_and_placeholder_only" and "should_return_response_200_with_nothing_and_placeholder_only" . 

I think a route with /somepath/{everything} should be fine to get called with just /somepath/ or /somepath. Feel free to let me know if this is not the case and I can revert that commit.  Let me know what you think and if this would cause any issues. Thanks

marcoaam:
The other option could be get all of those 3 scenarios to return 404 as the placeholder is empty and to have specific routes for "/" and "/somepath/" separate from the catch all routes. But that probably means some breaking changes for some people as it will impact the current routes.

TomPallister:
@marcoaam the test is valid so we need to put it back in. The user has a reroute set up /products/{id}. Requesting /products/ to the api should match not that route, a request on /products/1 should match.

The solution is to make the regex not match .* and only valid url characters until the next slash or end of string whichever comes first.

marcoaam:
Thanks for getting back to me so quickly. No problem. I will get that done.

Maybe something for another time but I still think that maybe it is worth reviewing those 2 test scenarios that I mentioned as they are very similar to this scenario but they expect 200 instead of not found and that seems to me inconsistent with this current one.

TomPallister:
@marcoaam looks good. I will merge and we can get it in the next release!

PR #420, opened 2018-06-19
Incorrect routing when UpstreamHost is specified and UpstreamHttpMethod is empty

### LINKED ISSUE #419: Incorrect routing when UpstreamHost is specified and UpstreamHttpMethod is empty
## Expected Behavior
When a reroute rule specifies `UpstreamHost`, it should only match requests with that host name. There's a bug where this check is ignored if the rule has an empty or null `UpstreamHttpMethod`

## Actual Behavior
`UpstreamHost` is ignored

## Specifications

  - Version: 7.0.5
  - Platform: all

I will submit a PR shortly that addresses this.

davidni:
Fixes #419


TomPallister:
@davidni this looks good, pulled the code so I could understand it better. These tests are not very easy to understand which is my fault! I have made a note to look at them and see if I can separate the functionality out so they dont need so much setup one day!

Issue #374, opened 2018-05-29
why not use "blocking queries" with consul?


SnailDev:
## Expected Behavior / New Feature

> Many endpoints in Consul support a feature known as "[blocking queries](https://www.consul.io/api/index.html#blocking-queries)". A blocking query is used to wait for a potential change using long polling. Not all endpoints support blocking, but each endpoint uniquely documents its support for blocking queries in the documentation.

blocking queries  could notify ocelot the lastest configuration immediately (when store configuration in consul) and  the lastest health services immediately(when service discovery and load balancer).

SnailDev:
I have read source of ocelot, with test,  and I found if use service discovery and load balancer, ocelot will get health reroute server every request and it will affect performance. So I think memorycache and block queries would be useful.

TomPallister:
I have merged a new feature that will allow users to specify if they want to poll consul

TomPallister:
Released in 7.0.5

SnailDev:
@TomPallister I have read your solution. but there are two problems:
1. PollingConsulServiceDiscoveryProvider init on the first request and _timer poll services from consul, so the first request will be fail.
2. _timer poll just run once. I logged in poll method and I found the log print just on first request.
```C#
private async Task Poll()
{
    _logger.LogInformation("Started polling consul");
    _services = await _consulServiceDiscoveryProvider.Get();
     _logger.LogInformation("Finished polling consul");
}
```

please confirm it, thank you very much.

PR #426, opened 2018-06-21
block queries on servicediscovery with consul

SnailDev:
Fixes / New Feature #
block queries on servicediscovery with consul

issue #374 
PollingInterval default is 10000 ms. May be someone should add this introduction in doc.

## Proposed Changes
change query to block query could solve the problem of polling interval


SnailDev:
@TomPallister DiskFileConfigurationRepositoryTests/should_return_file_configuration_if_environment_name_is_unavailable failed.

PR #427, opened 2018-06-22
block queries on servicediscovery with consul

SnailDev:
Fixes / New Feature #
block queries on servicediscovery with consul

issue #374
PollingInterval default is 10000 ms. May be someone should add this introduction in doc.

## Proposed Changes
change query to block query could solve the problem of polling interval

### DIFF
--- a/src/Ocelot/ServiceDiscovery/Providers/PolingConsulServiceDiscoveryProvider.cs
+++ b/src/Ocelot/ServiceDiscovery/Providers/PolingConsulServiceDiscoveryProvider.cs
@@ -93,27 +93,6 @@ namespace Ocelot.ServiceDiscovery.Providers
             await Poll();
         }
 
-        public async Task<List<Service>> GetWithBlockQueries(ulong waitIndex, int timeOut)
-        {
-            var queryResult = await _consul.Health.Service(_config.KeyOfServiceInConsul, string.Empty, true, new QueryOptions() { WaitIndex = waitIndex, WaitTime = TimeSpan.FromMilliseconds(timeOut) });
-
-            var services = new List<Service>();
-
-            foreach (var serviceEntry in queryResult.Response)
-            {
-                if (IsValid(serviceEntry))
-                {
-                    services.Add(BuildService(serviceEntry));
-                }
-                else
-                {
-                    _logger.LogWarning($"Unable to use service Address: {serviceEntry.Service.Address} and Port: {serviceEntry.Service.Port} as it is invalid. Address must contain host only e.g. localhost and port must be greater than 0");
-                }
-            }
-
-            return services.ToList();
-        }
-
         private Service BuildService(ServiceEntry serviceEntry)
         {
             return new Service(
--- a/src/Ocelot/ServiceDiscovery/Providers/PolingConsulServiceDiscoveryProvider.cs
+++ b/src/Ocelot/ServiceDiscovery/Providers/PolingConsulServiceDiscoveryProvider.cs
@@ -15,42 +15,130 @@ namespace Ocelot.ServiceDiscovery.Providers
 {
     public class PollingConsulServiceDiscoveryProvider : IServiceDiscoveryProvider
     {
+        private readonly ConsulRegistryConfiguration _config;
         private readonly IOcelotLogger _logger;
-        private readonly IServiceDiscoveryProvider _consulServiceDiscoveryProvider;
-        private readonly Timer _timer;
-        private bool _polling;
+        private readonly IConsulClient _consul;
+        private const string VersionPrefix = "version-";
         private List<Service> _services;
-        private string _keyOfServiceInConsul;
 
-        public PollingConsulServiceDiscoveryProvider(int pollingInterval, string keyOfServiceInConsul, IOcelotLoggerFactory factory, IServiceDiscoveryProvider consulServiceDiscoveryProvider)
-        {;
+        private int _pollingInterval;
+        private ulong _waitIndex;
+
+        public PollingConsulServiceDiscoveryProvider(int pollingInterval, ConsulRegistryConfiguration config, IOcelotLoggerFactory factory, IConsulClientFactory clientFactory)
+        {
+            ;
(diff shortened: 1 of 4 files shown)


SnailDev:
@TomPallister 
image

TomPallister:
@SnailDev thanks for this I will have a look at it later!

src/Ocelot/ServiceDiscovery/Providers/PolingConsulServiceDiscoveryProvider.cs
```diff
@@ -15,42 +15,109 @@ namespace Ocelot.ServiceDiscovery.Providers
...
 
-        public PollingConsulServiceDiscoveryProvider(int pollingInterval, string keyOfServiceInConsul, IOcelotLoggerFactory factory, IServiceDiscoveryProvider consulServiceDiscoveryProvider)
-        {;
+        private int _pollingInterval;
+        private ulong _waitIndex;
+
+        public PollingConsulServiceDiscoveryProvider(int pollingInterval, ConsulRegistryConfiguration config, IOcelotLoggerFactory factory, IConsulClientFactory clientFactory)
+        {
+            ;
+            _pollingInterval = pollingInterval == 0 ? 10000 : pollingInterval;
```
TomPallister: The default pollinginterval is set by ServiceProviderConfigurationCreator.cs as 0, not sure we want to override it here as that can get confusing!

src/Ocelot/ServiceDiscovery/Providers/PolingConsulServiceDiscoveryProvider.cs
```diff
@@ -15,42 +15,109 @@ namespace Ocelot.ServiceDiscovery.Providers
...
+        {
+            if (_services == null)
+            {
+                _services = await GetService();
+            }
+
+            return _services;
+        }
+
+        public async Task<List<Service>> GetService()
```
TomPallister: This should be private method because nothing else calls it.

src/Ocelot/ServiceDiscovery/Providers/PolingConsulServiceDiscoveryProvider.cs
```diff
@@ -15,42 +15,109 @@ namespace Ocelot.ServiceDiscovery.Providers
...
+
+            if (_waitIndex == 0 || _pollingInterval == 0)
+            {
+                queryResult = await _consul.Health.Service(_config.KeyOfServiceInConsul, string.Empty, true);
+            }
+            else
+            {
+                // block queries
+                queryResult = await _consul.Health.Service(_config.KeyOfServiceInConsul, string.Empty, true,
+                    new QueryOptions() { WaitIndex = _waitIndex, WaitTime = TimeSpan.FromMilliseconds(_pollingInterval) });
```
TomPallister: Does this waittime block the recursive call to Poll() in Poll()? I will pull the code and test this but I though I would ask to save some time

src/Ocelot/ServiceDiscovery/Providers/PolingConsulServiceDiscoveryProvider.cs
```diff
@@ -15,42 +15,109 @@ namespace Ocelot.ServiceDiscovery.Providers
...
-            _keyOfServiceInConsul = keyOfServiceInConsul;
-            _consulServiceDiscoveryProvider = consulServiceDiscoveryProvider;
-            _services = new List<Service>();
-        
-            _timer = new Timer(async x =>
+
+            _config = config;
+            _consul = clientFactory.Get(_config);
+
+            Task.Factory.StartNew(async () =>
```
TomPallister: What happens if Poll() throws a transient exception e.g. Consul is down for a few seconds when it makes it's request? Will it start again? I think timer might be better if it starts the process again. However I don't know how this would work at the moment because the timer doesnt try catch or anything!

src/Ocelot/ServiceDiscovery/Providers/PolingConsulServiceDiscoveryProvider.cs
SnailDev: yes, waittime block the recursive call to Poll() in Poll().

src/Ocelot/ServiceDiscovery/Providers/PolingConsulServiceDiscoveryProvider.cs
SnailDev: >In addition to index, endpoints that support blocking will also honor a wait parameter specifying a maximum duration for the blocking request. This is limited to 10 minutes. If not set, the wait time defaults to 5 minutes. This value can be specified in the form of "10s" or "5m" (i.e., 10 seconds or 5 minutes, respectively). A small random amount of additional wait time is added to the supplied maximum wait time to spread out the wake up time of any concurrent requests. This adds up to wait / 16 additional time to the maximum duration.

2000 may be too little? I'll set the default pollinginterval in ServiceProviderConfigurationCreator.cs

src/Ocelot/ServiceDiscovery/Providers/PolingConsulServiceDiscoveryProvider.cs
SnailDev: >When this is provided, the HTTP request will "hang" until a change in the system occurs, or the maximum timeout is reached. A critical note is that the return of a blocking request is no guarantee of a change. It is possible that the timeout was reached or that there was an idempotent write that does not affect the result of the query.

when change service config, client should update at once. but use timer, there have some interva of two consecutive modificationsl. if i use try..catch.. in Poll(), would it be better?

TomPallister:
@SnailDev are you happy for me to merge this? If you are then I will! :)

SnailDev:
@TomPallister  Of course happy for you to merge it. I'm also not sure why all checks have failed. UnitTests in my pc all passed.

thiagoloureiro:
@TomPallister  need help in this one ? We will still merge this? Thanks!

thiagoloureiro:
closing the Pr after inactivity

PR #560, opened 2018-08-18
Addressing issue #559 (Infinite configuration file growth when ASPNETCORE_ENVIRONMENT is defined)

### LINKED ISSUE #559: Infinite configuration file growth when ASPNETCORE_ENVIRONMENT is defined
## Expected Behavior

Ocelot should be able to read and merge multiple configuration files (ocelot.*.json).

## Symptom

Merged configuration files are growing in size (in an arithmetic progression) on each API Gateway restart.
2nd and all subsequent API Gateway app runs have broken configuration.

## Problem

**When**

_ASPNETCORE_ENVIRONMENT_ environment variable is defined

**and** 
Ocelot has at least one dedicated configuration file initially deployed (Ocelot.Something.json)

**then**

_On first run_ Ocelot produces "Ocelot.{env.EnvironmentName}.json" file containing initial content of "Ocelot.Something.json".

_On second run_ Ocelot uses "Ocelot.{env.EnvironmentName}.json" and "Ocelot.Something.json" as input for config merge operation and produces new "Ocelot.{env.EnvironmentName}.json" file that contains concatenation of existing "Ocelot.{env.EnvironmentName}.json" and "Ocelot.Something.json". The resulting file gets "Ocelot.Something.json" content duplicated two times (since that moment API Gateway configuration is broken).

_On third run_ ... Ocelot does the same operation (again). The merged configuration file becomes 3x of "Ocelot.Something.json"

The problem seems to be caused by interference of business-logic of two classes:
Configuration\Repository\DiskFileConfigurationRepository.cs
and
DependencyInjection\ConfigurationBuilderExtensions.cs

In **DiskFileConfigurationRepository.cs** file the configuration file name is built from current IHostingEnvironment.EnvironmentName value:

```
public DiskFileConfigurationRepository(IHostingEnvironment hostingEnvironment)
{
     _configFilePath = $"{AppContext.BaseDirectory}/{ConfigurationFileName}{(string.IsNullOrEmpty(hostingEnvironment.EnvironmentName) ? string.Empty : ".")}{hostingEnvironment.EnvironmentName}.json";
}
```

then FileConfiguration is saved like so:

```
public Task<Response> Set(FileConfiguration fileConfiguration)
{
    string jsonConfiguration = JsonConvert.SerializeObject(fileConfiguration, Formatting.Indented);

    lock(_lock)
    {
        if (System.IO.File.Exists(_configFilePath))
        {
            System.IO.File.Delete(_configFilePath);
        }

        System.IO.File.WriteAllText(_configFilePath, jsonConfiguration);
    }

    return Task.FromResult<Response>(new OkResponse());
}
```

**ConfigurationBuilderExtensions.cs** 
contains _AddOcelot()_ extension method which knows nothing about IHostingEnvironment and assumes that any file matching pattern "(?i)ocelot\\.([a-zA-Z0-9]*)(\\.json)$" is perfectly valid candidate for merge (including the file created by DiskFileConfigurationRepository).

## Solution

The quick fix I can see here is to to pass IHostingEnvironment to AddOcelot() extension method to let it filter out ocelot.{ASPNETCORE_ENVIRONMENT}.json during the merge cycle.

## Specifications

  - Version: 10.x and down to 7.x (at least)
  - Platform: Any
  - Subsystem: Ocelot configuration

## Comments

This issue actually breaks Continuous Integration / Continuous Deployment cycle when micro-services ecosystem is being developed and tested on IIS instances.

At the project I'm working on we have multiple .NET Core micro-services connected by API Gateway (which uses Ocelot). Each micro-service comes with it's own Gateway configuration (re-routes and aggregates). Some deployment environments have unique capabilities so we have as many ASPNETCORE_ENVIRONMENT variations as the number of unique environments.

When a micro-service is being deployed, the deployment agent replaces corresponding Ocelot.ServiceName.json file at the API Gateway root by the latest gateway configuration file from microservice bundle and restarts the API Gateway.
The problem is that this only works once :)
Currently our script explicitly deletes ocelot.{ASPNETCORE_ENVIRONMENT}.json before starting back the API Gateway service... but if API Gateway is being shutdown and restarted by IIS (by any reason) it breaks the Gateway service.

snb83:
## Fixes #559
- #559 

Made `IHostingEnvironment` optional for `AddOcelot()` to avoid breaking change.

## Proposed Changes

Filter out `ocelot.ASPNETCORE_ENVIRONMENT.json` when merging configs to avoid recurrent merge loop.


src/Ocelot/DependencyInjection/ConfigurationBuilderExtensions.cs
```diff
@@ -28,37 +29,42 @@ public static IConfigurationBuilder AddOcelotBaseUrl(this IConfigurationBuilder
...
+                .Where(fi => reg.IsMatch(fi.Name) && (fi.Name != excludeConfigName))
                 .ToList();
 
             var fileConfiguration = new FileConfiguration();
 
             foreach (var file in files)
             {
-                // windows and unix sigh...
-                if(files.Count > 1 && (Path.GetFileName(file) == "ocelot.json"))
+                if(files.Count > 1 && file.Name.Equals(primaryConfigFile, StringComparison.OrdinalIgnoreCase))
```
TomPallister: could you not just say and the file isnt ocelot.{env}.json here and then no other code would need to change?

src/Ocelot/DependencyInjection/ConfigurationBuilderExtensions.cs
```diff
@@ -28,37 +29,42 @@ public static IConfigurationBuilder AddOcelotBaseUrl(this IConfigurationBuilder
...
                 {
                     continue;
                 }
 
-                var lines = File.ReadAllText(file);
+                var lines = File.ReadAllText(file.FullName);
```
TomPallister: What does FullName do? :o

TomPallister (changes requested):
This looks good just requesting a few changes, also would it be possible for you to update the documentation to say you need to pass the environment into the method? Let me know if you cant be bothered and I can just merge this then make the changes.

src/Ocelot/DependencyInjection/ConfigurationBuilderExtensions.cs
```diff
@@ -28,37 +29,42 @@ public static IConfigurationBuilder AddOcelotBaseUrl(this IConfigurationBuilder
             return builder;
         }
 
-        public static IConfigurationBuilder AddOcelot(this IConfigurationBuilder builder)
+        public static IConfigurationBuilder AddOcelot(this IConfigurationBuilder builder, IHostingEnvironment env = null)
```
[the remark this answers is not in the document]
snb83: I agree... optional arguments usually the fastest, but not the best way to solve the problem. Plus the API looks a bit messy in that case. I'd prefer overloaded versions of the method.
But in this case (as far as I understand) you think it would be better to always provide IHostingEnvironment instance (even if user may never use env specific configuration files).
Anyway, I'll make env a required argument in next commit.

src/Ocelot/DependencyInjection/ConfigurationBuilderExtensions.cs
```diff
@@ -28,37 +29,42 @@ public static IConfigurationBuilder AddOcelotBaseUrl(this IConfigurationBuilder
...
         {
-            return builder.AddOcelot(".");
+            return builder.AddOcelot(".", env);
         }
 
-        public static IConfigurationBuilder AddOcelot(this IConfigurationBuilder builder, string folder)
+        public static IConfigurationBuilder AddOcelot(this IConfigurationBuilder builder, string folder, IHostingEnvironment env = null)
```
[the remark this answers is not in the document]
snb83: Same thoughts here... will fix that in next commit.

src/Ocelot/DependencyInjection/ConfigurationBuilderExtensions.cs
snb83: Well I thought that using more strict FileInfo type instead of storing file path can simplify the code and thus unix/windows specific path handling can be omitted in this case.
https://msdn.microsoft.com/ru-ru/library/system.io.fileinfo_properties(v=vs.110).aspx

But if you want me to minimize the code impact I can just return it to initial state and deal with strings.

TomPallister: no thats fine, I just didnt know what it was :)

Issue #607, opened 2018-09-08
Always 200 response code in middlewares


allrameest:
I'm trying to catch errors both with the built in middlewares and custom middlewares. But `context.Response.StatusCode` is always 200.

This issue seems to be about the same thing and should be solved: https://github.com/ThreeMammals/Ocelot/issues/370
But for me it doesn't work.

I've reproduced it in this repository:
https://github.com/allrameest/OcelotErrorHandling
If you start it and browse to http://localhost:7000/error or /whatever, the correct status code is sent to the client. But in the middleware it is 200.

TomPallister:
@allrameest thanks for this, change was made before and it looks like ive somehow just steamrolled it :/ no idea why.

Will release this to NuGet ASAP.

TomPallister:
fixed in 11.0.3, any problems let me know

PR #380, opened 2018-06-04
Fix incorrect response StatusCode for middleware added before Ocelot (#370)

AlexKuriatnyk:
#370

PR #608, opened 2018-09-10
Correct status code available for middleware

allrameest:
Status code is set directly instead of in OnStarting so that wrapping middlewares see the correct status code and can act on it. Fixes: #607


TomPallister:
@allrameest looks good, this should already be changed :( I will see why it isnt

allrameest:
@TomPallister This PR changed it in SetErrorResponseOnContext, but this is in another path.
https://github.com/ThreeMammals/Ocelot/pull/380

Thanks for merging!

TomPallister:
ahh yeh

Issue #639, opened 2018-09-27
Issue with Sec-WebSocket-Protocol on WS middleware


vasicvuk:
## Expected Behavior / New Feature

I wanted to add downstream WS to Active MQ Artemis STOMP.

I added reroute configuration.
Reroute configuration:
```
{
      "DownstreamPathTemplate":"/stomp",
      "DownstreamScheme":"ws",
      "DownstreamHostAndPorts":[
        {
          "Host":"broker",
          "Port":61623
        }
      ],
      "UpstreamPathTemplate":"/broker/stomp"
    }
```

It is expected to work like this but.

## Actual Behavior / Motivation for New Feature

Actual behavior is that i get in browser:
Error during WebSocket handshake: Sent non-empty 'Sec-WebSocket-Protocol' header but no response was received.

Response code is: 101

Headers that are sent are:
Sec-WebSocket-Extensions: permessage-deflate; client_max_window_bits
Sec-WebSocket-Key: hMNbcPOBCOP9t4S9Xb4C6g==
Sec-WebSocket-Protocol: v10.stomp, v11.stomp, v12.stomp
Sec-WebSocket-Version: 13

Headers received:
Sec-WebSocket-Accept: 2koEJ/JywLlfry8DM/eMQDZgJD8=

But there is no Sec-WebSocket-Protocol in response. 

I guess the issue is here:
https://github.com/ThreeMammals/Ocelot/blob/89f0cc786aa0895a9945eeea9a8ba6ad5ba21fb9/src/Ocelot/WebSockets/Middleware/WebSocketsProxyMiddleware.cs#L36

If  Sec-WebSocket-Protocol is not redirected then it will not give a response of supported protocol

There is no error i log

## Specifications

  - Version: 12
  - Platform: Kastrel
  - Subsystem: Linux

vasicvuk:
I've debugged a code a little and i found out that it is hanging at line:
```
await wsToDownstreamService.ConnectAsync(uri, CancellationToken.None);
```
I am not sure what is problem why it won't connect since. Maybe it is issue with ClientWebSocket. With nginx same works.

vasicvuk:
I created a pull request solving this issue #642 .

[...]

PR #642, opened 2018-09-28
Update to support Subprotocols. Solves #639

vasicvuk:
Ocelot websocket middleware did not work for STOMP over websocket. After investigation i found out that the issue was with subprotocol and headers that are send and filtered. 

I the end i used ASP.Net core proxy as a reference to solve the issue here:

https://github.com/aspnet/Proxy/blob/3015029f5169b699dfa16bcf85cfc7ca0aaddae7/src/Microsoft.AspNetCore.Proxy/ProxyAdvancedExtensions.cs

So i modified the code to use the way ASP.Net proxy handles this.

Fixes / New Feature #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/src/Ocelot/WebSockets/Middleware/WebSocketsProxyMiddleware.cs
+++ b/src/Ocelot/WebSockets/Middleware/WebSocketsProxyMiddleware.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
 using System.Net.WebSockets;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


TomPallister:
@vasicvuk thanks for the PR, I will check it out ASAP!!

TomPallister:
@vasicvuk this looks OK to me and all the tests are passing, will merge.

PR #665, opened 2018-10-16
Add bool in configuration to choose between http and https schemas for service discovery provider

ismael-soriano:
Add boolean `UseHttpsScheme` in `ServiceProviderConfiguration` .

## Proposed Changes
This change is intended so the user can choose through configuration if he is using an http or an https port for the service discovery provider connection.

````
"ServiceDiscoveryProvider": {
    "Host": "localhost",
    "Port": 5443,
    "Token": "footoken",
    "Type": "Consul",
    "UseHttpsScheme": true
}
````

> **Note**: Once the changes have been merged I'll make another PR on `Ocelot.Provider.Consul` repository to apply the scheme depending on its value.

### CODE REVIEW

**Comment by TomPallister** [comment]
@ismael-soriano thanks for your PR! Can you explain the scenario that this covers? We already have DownstreamScheme on ReRoute and Global level so I assume it is when using the dynamic routing and you don't want to use Global?

In this case rather than have another boolean flag in Ocelot I was thinking of removing the       DownstreamScheme concept and the user will always specify http:// or https:// in their Host setting which I would then change the name of to something like BaseUrl (cant think of anything better atm). This is obviously a breaking change but given how many places (3) now can have varying http or https I think this is the best change.

**Comment by ismael-soriano** [comment]
The changes I was trying to make were specifically for the uri that the Consul client will use for its calls for service discovery.

Actually on the `Ocelot.Providers.Consul` package, the uri is only conformed with the `http` schema, so I cannot use an SSL port because the calls will not function properly.

```c#
return new ConsulClient(c =>
{
    c.Address = new Uri($"http://{config.Host}:{config.Port}");

    if (!string.IsNullOrEmpty(config?.Token))
    {
        c.Token = config.Token;
    }
});
```
To change this I've come to think of adding a bool through configuration to specifically state if the schema was `http` or `https`, because on configuration you are only stating the host not the full URI.

In case we provided the full URI through configuration the boolean flag would not be necessary at all.

So, what would you think will be the best approach to fix this?

thiagoloureiro:
closing the Pr after inactivity

PR #672, opened 2018-10-25
Grammar

EdwinVanRooij:
Changed `allow's` to `allows`, typo probably.


briansantura:
Good catch @EdwinVanRooij

PR #698, opened 2018-11-26
Fixed error message when reroute configuration is not found #650

### LINKED ISSUE #650: Possibly better logging messages when failing to find/match upstream routes

Currently, if I incorrectly enter an upstream route into Postman/curl I get the following logging message: `Unable to find downstream route for path: /searchProducts, verb: GET`

e.g.
```
[40m[1m[33mwarn[39m[22m[49m: Ocelot.DownstreamRouteFinder.Middleware.DownstreamRouteFinderMiddleware[0]
      requestId: 0HLH882EDUUIH:00000006, previousRequestId: no previous request id, message: Unable to find downstream route for path: /searchProducts, verb: GET
Ocelot.DownstreamRouteFinder.Middleware.DownstreamRouteFinderMiddleware:Warning: requestId: 0HLH882EDUUIH:00000006, previousRequestId: no previous request id, message: Unable to find downstream route for path: /searchProducts, verb: GET
[40m[1m[33mwarn[39m[22m[49m: Ocelot.Responder.Middleware.ResponderMiddleware[0]
```

For a long time while I was first learning/experimenting with this framework, I kept reading that message as:
- Successfully mapped your upstream template to one of the ReRoute configuration sections.
- Attempted to do an `HTTPClient` request to the mapped downstream template/url .. but failed.

After a lot of trial and error, I think I'm seeing this error message as a general/generic message that _something_ has failed with the mapping/calling.

I would hope/expect that the logging message be slightly more accurate and hopefully suggest that **the error is because the upstream route provided failed to match any reroute configuration upstream-template.**

Logging for unmapped/failed-to-map an upstream route to an upstream-template is a slightly confusing/misleading logging message.

- Have no re-route configuration sections
- goto `http://<whatever>:<whatever-port>/pewpew` (an upstream route that hasn't been defined)
- 404 returned (correct)
- logging message: `Unable to find downstream route for path: /pewpew, verb: GET`

## Specifications

  - Version: 11.0.3

rubenmamo:
Fixes #650

## Proposed Changes
I changed the error message to be clearer that the issue was due to not finding the ReRoute configuration for the upstream path.

I'm a bit conflicted about this change as it seems to simple :) Having said so, it seems that this error is only used for this case.

In addition, whilst looking into this. I got confused by this part in the [DownstreamRouteFinder.cs](https://github.com/ThreeMammals/Ocelot/blob/66b68fc6859af2cf1268083e19cdec49c1631e27/src/Ocelot/DownstreamRouteFinder/Finder/DownstreamRouteFinder.cs).
```
            if (downstreamRoutes.Any())
            {
                var notNullOption = downstreamRoutes.FirstOrDefault(x => !string.IsNullOrEmpty(x.ReRoute.UpstreamHost));
                var nullOption = downstreamRoutes.FirstOrDefault(x => string.IsNullOrEmpty(x.ReRoute.UpstreamHost));

                return notNullOption != null ? new OkResponse<DownstreamRoute>(notNullOption) : new OkResponse<DownstreamRoute>(nullOption);
            }
```
Am I right that this can be replaced with?
```
            if (downstreamRoutes.Any())
            {
                return new OkResponse<DownstreamRoute>(downstreamRoutes.OrderBy(x => !string.IsNullOrEmpty(x.ReRoute.UpstreamHost)).FirstOrDefault());
            }
```

### DIFF
--- a/src/Ocelot/DownstreamRouteFinder/Finder/UnableToFindDownstreamRouteError.cs
+++ b/src/Ocelot/DownstreamRouteFinder/Finder/UnableToFindDownstreamRouteError.cs
@@ -5,7 +5,7 @@ namespace Ocelot.DownstreamRouteFinder.Finder
     public class UnableToFindDownstreamRouteError : Error
     {
         public UnableToFindDownstreamRouteError(string path, string httpVerb) 
-            : base($"Unable to find downstream route for path: {path}, verb: {httpVerb}", OcelotErrorCode.UnableToFindDownstreamRouteError)
+            : base($"Failed to match ReRoute configuration for upstream path: {path}, verb: {httpVerb}.", OcelotErrorCode.UnableToFindDownstreamRouteError)
         {
         }
     }


rubenmamo:
I don't think the Travis CI build failures are caused by something I did. Can I do something to fix it?

TomPallister:
@rubenmamo thanks for the PR I will make this change as it is clearer.

Unfortunately we cannot change the code you suggested. Try changing it on your local and running the tests. You will see which scenarios need this code. I agree its super confusing :/

PR #702, opened 2018-11-29
Web Sockets: work around .NET FX bug

### LINKED ISSUE #701: WebSockets broken on .NET framework (unhandled ArgumentException when setting outgoing headers)
## Expected Behavior
Web Sockets support should work when running Ocelot under .NET Framework. It doesn't , due to a bug in .NET Framework (the bug doesn't occur in .NET Core).

## Actual Behavior
Web Socket connection fails to upgrade because an exception is unhandled while preparing the web socket client for the target service. Setting restricted headers such as `User-Agent` results in an `ArgumentException` bubbling out of method `ThrowOnRestrictedHeader` in class `WebHeaderCollection`. This has been acknowledged as a .NET Framework bug, and there is no other way to set such headers other than rewriting the implementation of those classes. My work-around is to simply catch these validation errors and move on. `User-Agent` and other restricted headers won't be forwarded, but at least the E2E works and the web socket connection is proxied.

  1. Create a simple SignalR service hosted behind Ocelot
  1. Run Ocelot under .NET Framework 4.6.2
  1. Attempt to connect through Ocelot using a SignalR browser client (npm package `@aspnet/signalr`)

## Specifications

  - Version: 12.0.1
  - Platform: Windows 10 + .NET Framework 4.6.2

davidni:
Fixes #701.

### DIFF
--- a/src/Ocelot/WebSockets/Middleware/WebSocketsProxyMiddleware.cs
+++ b/src/Ocelot/WebSockets/Middleware/WebSocketsProxyMiddleware.cs
@@ -90,7 +90,16 @@ namespace Ocelot.WebSockets.Middleware
             {
                 if (!NotForwardedWebSocketHeaders.Contains(headerEntry.Key, StringComparer.OrdinalIgnoreCase))
                 {
-                    client.Options.SetRequestHeader(headerEntry.Key, headerEntry.Value);
+                    try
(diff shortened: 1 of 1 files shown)


TomPallister:
@davidni thanks for this!

Issue #678, opened 2018-11-07
Wrong string representation of the List<T>


dmitriifilonenko:
In method Ocelot.Authorisation.Authorise(...):
if claims do not contain required value then an ErrorResponse will be created:

```
return new ErrorResponse<bool>(new ClaimValueNotAuthorisedError(
                                $"claim value: {values.Data} is not the same as required value: {required.Value} for type: {required.Key}"));
```
But values.Data is of type List<string>. So implicit convertation of values.Data to string gives _"System.Collections.Generic.List`1[System.String]"_ instead of sequence of items in the list like "Item1, Item2, Item3".

As a result I see the following error in the log whenever claims in request do not contain a claim with the name "role":

> claim value: System.Collections.Generic.List`1[System.String] is not the same as required value: Admin for type: role

## Specifications

  - Version: 12.0.1

briansantura:
@dmitriifilonenko I have created a pull request for this issue.

raman-m:
Fixed by #711

PR #711, opened 2018-12-07
Claim values are now being shown when constructing. ClaimValueNotAuthorisedError.

briansantura:
Fixes issue #678 

## Proposed Changes
ClaimsAuthoriser.cs
` return new ErrorResponse<bool>(new ClaimValueNotAuthorisedError(
                                $"claim value: {string.Join(", ", values.Data)} is not the same as required value: {required.Value} for type: {required.Key}"));`


TomPallister:
@briansantura looks good, any change of a test for this to check the string in the error response is what we expect, sorry I missed this when I wrote it.

PR #715, opened 2018-12-12
add support for custom ocelot pipeline

WeihanLi:
Fixes / New Feature #

add support for custom ocelot pipeline

sometimes we may wanna custom pipeline, but when I wanna add a custom ocelot middleware, it's not so easy to update

### DIFF
--- a/samples/OcelotCustomMiddleware/OcelotCustomMiddleware.csproj
+++ b/samples/OcelotCustomMiddleware/OcelotCustomMiddleware.csproj
@@ -1,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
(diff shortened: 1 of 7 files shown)


TomPallister:
@WeihanLi thanks for the PR, I'm not sure what functionality it adds to Ocelot? Can you please explain the use case? You can just bring the individual Ocelot middlewares into any project and call them as required rather than this approach.

WeihanLi:
I wanna customize the ocelot middlewares, sometimes rateLimit or header/claim transform maybe not needed for me, sometimes I wanna add a custom middleware.
I find that it's not so easy to customize the ocelot middlewares, so I made this PR to add a method for custom the middlewares by developers a little easier.

philproctor:
@WeihanLi Let me give this some thought. I understand the need for this and it makes sense, however this approach in this PR is a bit dangerous for third parties -- especially with package updates. There is already a mechanism for customizing the pipeline, however it is a bit on the rigid side and could stand the be a bit more flexible so I get what you're doing here.

WeihanLi:
@philproctor  Okay, thanks for your reply.

Issue #718, opened 2018-12-14
Websocket authentication token


Rakiah:
Hello, I'm currently using the websocket proxying feature which is working correctly but I would require my system to run a bit of code to check for token validity and token existence, from what I understand this is not implemented, where should I start to implement this kind of feature myself, or is someone already working on something similar ?

I was planning to use a query parameter retrieved by the HttpContext to check for this (I can already do that at micro-service level, but it would be better to do it at gateway level)

Rakiah:
Merge request opened for this issue

raman-m:
Will be fixed in #1040

raman-m:
Duplicate of #1040

PR #720, opened 2018-12-17
added authentication middleware for websockets in ocelot pipeline

Rakiah:
Just added the same line of codes that is used in the HTTP pipeline to allow users to specify an Authentication Middleware but in websockets

This was basically an answer to my question at #718


src/Ocelot/Middleware/Pipeline/OcelotPipelineExtensions.cs
```diff
@@ -36,6 +36,12 @@ public static class OcelotPipelineExtensions
                     app.UseDownstreamRouteFinderMiddleware();
                     app.UseDownstreamRequestInitialiser();
                     app.UseLoadBalancingMiddleware();
+
```
TomPallister: This should be right after the DownstreamRouteFinderMiddleware() call if you are going to use it.

thiagoloureiro: @Rakiah please can you take a look into @TomPallister  comment?

PR #766, opened 2019-01-26
Typo querystring => query string

DeonHeyns:
Fixes / New Feature #

## Proposed Changes

  -
  -
  -


DeonHeyns:
Instead of making these PRs as I read through the docs, perhaps a better approach is to send a single PR with all proposed changes.

PR #790, opened 2019-02-22
IOcelotBuilder opens the IMvcCoreBuilder property for easy customization

AClumsy:
IOcelotBuilder opens the IMvcCoreBuilder property for easy customization


AClumsy:
@TomPallister The opening of IMvcCoreBuilder can help Ocelot add more Mvc functions.
If you need to extend the custom controller in Ocelot.Administration, you can't expand now.

geffzhang:
agree

geffzhang:
what is self-requested a review ？

PR #810, opened 2019-03-12
If the registered Consul node is unexpectedly down and not restarted immediately, other services should continue to find the registered service.

philiahe:


thiagoloureiro:
@philiahe your PR looks ok, but wanted to understand better the difference  between ServiceEntry  and CatalogEntry ? Thanks!

philiahe:
> @philiahe I want to understand better the difference between ServiceEntry and CatalogEntry ? Thanks!
> 
> And also looks like we have some failing tests

1. In cluster mode:
image

2. Client node unexpectedly down
image

3. Client node is failed
image

4. Other nodes in the cluster are not accessible
image

5. You can fix this either way：
use catalog （See PR）
image

drop  passing=true
image
image

philiahe:
**Here are some official introductions:**

https://www.consul.io/docs/agent/basics.html
image

image

thiagoloureiro:
@philiahe reverted due failing tests, can you check the tests with your changes please?
 Then create new tests if necessary.

philiahe:
There was a problem with the PR, which I fixed in https://github.com/ThreeMammals/Ocelot/pull/825

PR #819, opened 2019-03-19
remove duplicate `IHttpRequester` register

WeihanLi:
Fixes

## Proposed Changes

`IHttpRequester` registered twice in OcelotBuilder

image

see details:
- <https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/DependencyInjection/OcelotBuilder.cs#L101>
- <https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/DependencyInjection/OcelotBuilder.cs#L110>


thiagoloureiro:
What do you think guys? @TomPallister  @philproctor 

Im a bit confused with the breaking tests always, can't see if there are a breaking change or the tests are failing like before

PR #825, opened 2019-03-20
Allows access to unpass node

philiahe:
Fixes 
If the registered Consul node is unexpectedly down and not restarted immediately, other services should continue to find the registered service.

There was a problem with the previous PR, which I fixed this time

https://github.com/ThreeMammals/Ocelot/pull/810

### DIFF
--- a/src/Ocelot.Provider.Consul/Consul.cs
+++ b/src/Ocelot.Provider.Consul/Consul.cs
@@ -27,7 +27,7 @@
 
         public async Task<List<Service>> Get()
         {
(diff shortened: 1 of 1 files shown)


thiagoloureiro:
I can't see how this can be good, If I return a node which is unhealthy or not passing it will break anyway no ?

philiahe:
> I can't see how this can be good, If I return a node which is unhealthy or not passing it will break anyway no ?

I mentioned the details in this post：
https://github.com/ThreeMammals/Ocelot/pull/810

Both two ways can solve this problem, but it is too troublesome for the last PR to change unit tests synchronously.

PR #899, opened 2019-05-22
Fix/consul poller option

aliprogrammer69:
Fixes / consul poller option#

## Proposed Changes

pollingInterval property of serviceDiscoveryProvider setting is now effective

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,14 @@
+<Project>
+  <PropertyGroup>
+    <RepositoryType>git</RepositoryType>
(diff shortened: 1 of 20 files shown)


thiagoloureiro:
Hello, 
1) Do we need to update documentation to explain how this feature/fix works? 
2) We need unit test the new code, please can you do it?
3) Integration tests if it's the case too.

aliprogrammer69:
Hello.
this fix don't need to update documentations.
i created an unit test for "ConsulFileConfigurationPollerOptionTest"

PR #909, opened 2019-05-31
Updated Consul discovery to resolve hostname

Hypernoodle:
Consul services fail SSL certificate validation since they are only associated with address instead of hostname.

Consul service discovery now registers discovered service with hostname resolved with Consul node list.


daisen:
can you help me how to resolve the node name can not be resolved by dns after this 'pull request'. Do I lost somethings to do

raman-m:
I see the major issue here! 
[Docs](https://github.com/ThreeMammals/Ocelot/blob/develop/docs/features/servicediscovery.rst) were not updated in this PR. Now, the community is experiencing other problems like #954

Issue #939, opened 2019-06-24
Support dynamic routing configuration in Service Fabric


RamjotSingh:
## New Feature
Having to deploy Ocelot every time routing changes is tedious. Ocelot allows for modifying the routes using json reload but without deployment it is not really possible to do this either. This increases the cost of managing services in general. The target of this feature is to modify\extend Ocelot in a way that once deployed, ocelot can adapt to changes in services behind the scene.

## Motivation for New Feature
[Traefik supports something similar](https://docs.traefik.io/configuration/backends/servicefabric/) and that is the real motivation behind this feature. Today we are using Traefik but a few unrelated reasons are forcing us to move away from it.

raman-m:
Hi Ramjot!
Are you still with Ocelot?

RamjotSingh:
No, since this wasn't picked up and we absolutely needed dynamic discovery, we went with Traefik.

raman-m:
Oops! @RamjotSingh

I didn't know you are an engineer at Microsoft 😄 I won't say how much I love Microsofting, but I do like pro contributors, and I definitely don't like the situation when such a pro dev like you has not received enough attention.

Since I am working on the mad release of [.NET 10](https://github.com/ThreeMammals/Ocelot/milestone/13), I have looked into the code, and the idea you expressed in #940 is getting me excited. Now that we have released version [24.1](https://github.com/ThreeMammals/Ocelot/releases/tag/24.1.0) and the overall situation with [Dynamic Routing](https://ocelot.readthedocs.io/en/latest/features/servicediscovery.html#sd-dynamic-routing) has improved, I believe it is time to reopen this feature proposal.

If you intend to contribute, I expect great adventures while developing this feature. Imagine having to refactor a number of core classes, plus it seems we would need a separate [Service Fabric provider](https://ocelot.readthedocs.io/en/latest/features/servicefabric.html) (currently it's built-in, but that may change), and we must write tons of tests. But I hope everything will be alright in the end, with a successful contribution and a great feature inside.

### My question is:
When you walk with your kids, what ice cream pack will you buy and eat, considering three products (not mammals) with the slogans "Ocelot Micro Soft" or "YARP Mini Hard" or "Traefik Pink Sparkles"? 😆

-------
### P.S. 

> No, since this wasn't picked up and we absolutely needed dynamic discovery, **we went with Traefik**.

Who's "we"? A team at Microsoft? LoL.
[...]

PR #940, opened 2019-06-24
Support dynamic routing configuration in Service Fabric

RamjotSingh:
Fixes / New Feature #939 

## Proposed Changes
Adding a new extension of Ocelot for Service Fabric. More details are in the issue. The TL;DR version is that this extension allows you to change routing without touching Ocelot at all.

The routes are no longer put in Ocelot.json instead each service declares it's routes in their Service Manifest. This looks something like this

```xml
<?xml version="1.0" encoding="utf-8"?>
<ServiceManifest Name="Web1Pkg"
                 Version="1.0.0"
                 xmlns="http://schemas.microsoft.com/2011/01/fabric"
                 xmlns:xsd="http://www.w3.org/2001/XMLSchema"
                 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <ServiceTypes>
    <!-- This is the name of your ServiceType. 
         This name must match the string used in RegisterServiceType call in Program.cs. -->
    <StatelessServiceType ServiceTypeName="Web1Type">
      <Extensions>
        <Extension Name="Ocelot.Routing">
          <Labels xmlns="http://schemas.microsoft.com/2015/03/fabact-no-schema">
            <Label Key="ocelot.routing.template">/api/{value},/EquipmentInterfaces:/api2/testing</Label>
          </Labels>
        </Extension>
      </Extensions>
    </StatelessServiceType>
  </ServiceTypes>
```

Service fabric cluster services are probed every 30 seconds and routes are recalculated. If a change occurs, configuration change event is fired and Ocelot reloads the components and routing is changed.

I have tested on Windows only at the moment. I have completed testing on UnSecure cluster. Testing on Secure cluster.

### Sample usage
```cs
config
    .SetBasePath(hostingContext.HostingEnvironment.ContentRootPath)
    .AddJsonFile("appsettings.json", true, true)
    .AddJsonFile("Ocelot.json")
    .AddServiceFabricExtension(new ServiceFabricClientFactoryOptions
    {
        ClusterManagementEndpoint = "http://localhost:19080"
    })
    .AddEnvironmentVariables();
```

### Pending
- [X] Windows UnSecure cluster
- [x] Windows Secure cluster
- [ ] Unit Tests

### DIFF
--- a/Ocelot.sln
+++ b/Ocelot.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2036
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29001.49
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{5CFB79B7-C9DC-45A4-9A75-625D92471702}"
 EndProject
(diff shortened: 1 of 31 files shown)


thiagoloureiro:
Plans for unit/integration tests and when we can have this done for merging? :)

RamjotSingh:
@thiagoloureiro I first wanted to have your opinion on if you liked the design. If people are in general ok with the design I will start working on Unit tests.

thiagoloureiro:
@RamjotSingh Oh, you're right mate. Give me some time to check this out then.
Thanks!

RamjotSingh:
The reason I had changed namespace was coz the providers were proving ServiceDiscovery while this is providing RouteDiscovery. But if you feel they should be the same, I can change it. Do you want me to rename the assembly too?

P.S. I was thinking of having a Provider for SF at a later point which does not use SF's own ReverseProxy but directly reaches service instance. But that is still up in the air and I am not super sure it is a good idea. My idea was to name that assembly (and namespace) as Ocelot.Provider.ServiceFabric

RamjotSingh:
@thiagoloureiro , @TomPallister Did you get a chance to look at this?

RamjotSingh:
@thiagoloureiro , @TomPallister Can you please take a look.

geffzhang:
> The reason I had changed namespace was coz the providers were providing ServiceDiscovery while this is providing RouteDiscovery. But if you feel they should be the same, I can change it. Do you want me to rename the assembly too?
> 
> P.S. I was thinking of having a Provider for SF at a later point which does not use SF's own ReverseProxy but directly reaches service instance. But that is still up in the air and I am not super sure it is a good idea. My idea was to name that assembly (and namespace) as Ocelot.Provider.ServiceFabric

ok

RamjotSingh:
@thiagoloureiro Any thoughts?

RamjotSingh:
@thiagoloureiro, @TomPallister Any plans for taking a look at this PR?

TomPallister:

raman-m:
@RamjotSingh @ggnaegi 
It is interesting! Would you like to reopen this PR?

PR #943, opened 2019-06-26
Allow to add delegating handlers by type

margaale:
I needed to add a lot of handlers and was aiming to do it by reflection so i added this method to allow me to do that.


margaale:
Also a question, when was bumped the core version to 2.2? Is it absolutely necessary?

thiagoloureiro:
> Also a question, when was bumped the core version to 2.2? Is it absolutely necessary? Aren't you breaking semver with this? also, is the dockerfile building? cause its based on 2.1 sdk

Some libraries and components was so outdated, the examples weren't working too.
I created a new PR (https://github.com/ThreeMammals/Ocelot/pull/950) with the dockerfile upgraded.

PR #950, opened 2019-07-01
Upgraded to .NET Core 2.2 Runtime/Sdk

thiagoloureiro:
Fixes / New Feature #

## Proposed Changes

  - Dockerfile Upgrade to 2.2 Runtime/SDK


margaale:
Sorry. I do not agree updating the core version. For example in our case we should upgrade all our servers to 2.2 cause we are not using docker there

Issue #961, opened 2019-07-10
Add ability to register custom load balancers


DavidLievrouw:
## New Feature
Add ability to register custom load balancers.

## Motivation for New Feature
We are using the Consul provider. We would like to add a custom load balancer that decides which service to route to, based on our custom metrics that are known in Consul.

E.g. we would like to give preference to services that have no "critical" or "warn" health checks. Or give preference to service instances that have a certain tag. Stuff like that...

This functionality is currently not possible with the built-in load balancers (CookieStickySessions, NoLoadBalancer, RoundRobin or LeastConnection). So I would like to be able to add load balancers, and specify them during DI composition.

This would require reworking the LoadBalancerFactory class, because only the built-in load balancers can be created there (switch statement).

(Possibly related to #870?)

## Specifications

  - Version: 13.5.2
  - Platform: Windows-x64
  - Target framework: netcoreapp2.2

TomPallister:
fixed by #1155

raman-m:
Fixed by https://github.com/ThreeMammals/Ocelot/pull/1155

PR #962, opened 2019-07-10
Add ability to register custom load balancers

DavidLievrouw:
New Feature #961

## Proposed Changes

  - Add ability to register custom ILoadBalancer implementations during DI composition.
  - Refactor the LoadBalancerFactory class, so that it can create more than only the built-in load balancers.
  - Extend the IOcelotBuilder with AddCustomLoadBalancer() helper methods.


src/Ocelot/DependencyInjection/OcelotBuilder.cs
```diff
@@ -82,6 +84,10 @@ public OcelotBuilder(IServiceCollection services, IConfiguration configurationRo
             Services.TryAddSingleton<IFileConfigurationRepository, DiskFileConfigurationRepository>();
             Services.TryAddSingleton<IFileConfigurationSetter, FileAndInternalConfigurationSetter>();
             Services.TryAddSingleton<IServiceDiscoveryProviderFactory, ServiceDiscoveryProviderFactory>();
+            Services.AddSingleton<ILoadBalancerCreator, NoLoadBalancerCreator>();
```
thiagoloureiro: Why not TryAddIngleton ?

test/Ocelot.UnitTests/DependencyInjection/OcelotBuilderTests.cs
```diff
@@ -289,5 +351,18 @@ private void ThenAnExceptionIsntThrown()
         {
             _ex.ShouldBeNull();
         }
+
+        private class FakeCustomLoadBalancer : ILoadBalancer
+        {
+            public Task<Response<ServiceHostAndPort>> Lease(DownstreamContext context)
+            {
+                throw new NotImplementedException();
```
thiagoloureiro: Missing implementation or will be like this?

src/Ocelot/DependencyInjection/OcelotBuilder.cs
DavidLievrouw: Because ```LoadBalancerFactory``` expects an ```IEnumerable<ILoadBalancerCreator>```, not a single one.
When registering ```ILoadBalancerCreator``` implementations using ```TryAddSingleton```, only the first one in ```OcelotBuilder``` would be registered.

See unit test ```should_add_custom_load_balancer_creators_by_di_factory_method_with_arguments```.

test/Ocelot.UnitTests/DependencyInjection/OcelotBuilderTests.cs
DavidLievrouw: Not relevant for ```OcelotBuilder``` tests.
I added a comment to the fake class.

TomPallister:

TomPallister:
fixed by #1155

PR #968, opened 2019-07-23
[New feature] Support claims to path transformation

vmartos:
Add the option to transform a claim to the downstream path. Similar to the ClaimsToQueryString forwarding. We are starting to use Ocelot in our backend and we need this feature, to transform the user id from the claims to the downstream backend. This way we can do things like this:

Get /api/users/me (defined only in the api gateway)
Get /api/users/{userId} (defined in the users microservice backend)
the request contains the user_id in the claims, so it can then change the downstream path with the user_id to the users microservice backend /api/users/{userId}

## Proposed Changes

  - Add a new middleware (ClaimsToDownstreamPathMiddleware), that using the new ChangeDownstreamPathTemplate substitutes the placeholder that matches the ClaimToThings existing key.
- Add unit tests and acceptance tests

An example usage (ocelot.json):

`{
      "UpstreamPathTemplate": "/api/users/me",
      "DownstreamHostAndPorts": [
        {
          "Host": "localhost",
          "Port": 63048
        }
      ],
      "DownstreamPathTemplate": "/api/users/{userId}",
      "DownstreamScheme": "http",
      "AuthenticationOptions": {
        "AuthenticationProviderKey": "AuthJWTKey",
        "AllowedScopes": []
      },
      "ChangeDownstreamPathTemplate": {
        "userId": "Claims[user_id] > value"
      }
    }`

if the DownstreamPathTemplate contains a placeholder that matches ChangeDownstreamPathTemplate existing key (in the example userId) it substitutes the value so will end up calling the downstream host with /api/users/123.

if ChangeDownstreamPathTemplate  contains a claim to things, but there is no match in the DownstreamPathTemplate it will respond with an error.

The middleware basically adds or changes a PlaceholderNameAndValue from the DownstreamContext.TemplatePlaceholderNameAndValues, so when the DownstreamUrlCreatorMiddleware executes it will replace the placeholder with the values from the claims.

### DIFF
--- a/src/Ocelot/Configuration/Builder/DownstreamReRouteBuilder.cs
+++ b/src/Ocelot/Configuration/Builder/DownstreamReRouteBuilder.cs
@@ -19,6 +19,7 @@ namespace Ocelot.Configuration.Builder
         private Dictionary<string, string> _routeClaimRequirement;
         private bool _isAuthorised;
         private List<ClaimToThing> _claimToQueries;
+        private List<ClaimToThing> _claimToDownstreamPath;
         private string _requestIdHeaderKey;
         private bool _isCached;
         private CacheOptions _fileCacheOptions;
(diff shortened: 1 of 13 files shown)


vmartos:
any update on this? it has sense to merge or do you thing it is useful?

djrhodes:
Hi, thanks for adding this functionality, we need exactly this in our project. Do you have any idea when this will be merged in and released?

thiagoloureiro:
@vmartos just merged, a new package will be generated as pre-release, @djrhodes  you can try from your project too using latest pre-release and then we will release the latest stable soon.

Issue #296, opened 2018-03-29
New Feature - Merge configuration files


AaronFixer:
## Expected Behavior / New Feature
This may already be a feature but I see no mention of it in the documentation and it isn't working in my project. 

It would be good to support multiple .json files, potentially with better naming conventions. I have an app with a number of services and routes, and it would be nice to separate them out into dedicated files rather then have one huge `configuration.json`. 

For example: 

 * AccountRoutes.json <- Would have all routes for the account controller/service
 * ComponentActions.json  <- Would have all the routes for component actions controller/service

etc. etc. 

Is this already possible? It seems that when I define multiple `.AddJsonFile()` it only picks up the routes in the last file it pinks up with the `ReRoutes` field.

TomPallister:
@AaronFixer thanks for your interest in the project! 

This feature isn't supported yet I will but it on the backlog but if you would like to have a go yourself please feel free to submit a PR.

There are a couple of considerations here.

1. Ocelot uses configuration.json as a database so a solution would need to take multiple files and merge them into configuration.json or have a second file that ocelot uses as a database (not sure I like this).

2. Been thinking about this for a while completely change how configuration works. I was thinking of making it more like identity server where it is provided statically in c# / database. 

3. You could just do this yourself in ConfigureAppConfiguration with something like below and it would not need to be part of Ocelot.

```csharp
.ConfigureAppConfiguration((hostingContext, config) =>
                {
                    merger.Merge();
      
                    config
                        .SetBasePath(hostingContext.HostingEnvironment.ContentRootPath)
                        .AddJsonFile("configuration.json");
                })
```

I think 1 and 3 are better options at the moment for me but that could change!

Anyway let me know if you want to give this a try and I will be of as much assistance as I can...otherwise I'm afraid you will have to wait a bit for the feature :(

AaronFixer:
For me this is definitely something I think I (and many others) could potentially benefit from with large projects wanting to use a Gateway! 

I'll have to do this in my spare time, although I don't imagine it should take too long. 

1 and 3 both work. 3 is something I could just use for the time being as a workaround. I'll see if I get a chance to work on a PR for 1. in the meantime.

Thanks for getting back to me!

TomPallister:
Yep I agree this would be a useful feature! If you don't get round to it its on my list and I expect to do it in the next months or so.

TomPallister:
[...]

PR #970, opened 2019-07-24
Fix to AddOcelot(), Config File Merge

rjwester:
Fixes / New Feature #296

@CesarD recently brought up a bug in a reply to the old #296  issue. I ran into this issue myself in my own project and made some adjustments to accommodate.

## Proposed Changes

  - AddOcelot() now includes the correct config file with the specified environment
     - This was done by removing the excludeConfigName variable and altering the regex
     - The files that will be included would look something like this (assuming a Development env) ocelot.json, ocelot.api.json, ocelot.api.Development.json, etc.


CesarD:
Thanks for taking care of this!

thiagoloureiro:
Had to revert, issues with tests, can you double check please?

csnance:
Any update on this? Merging the wrong set of config files by default is a significant issue for me.

Issue #913, opened 2019-06-04
AllowedScopes does not match string array in JWT scope claim


fsmullinslc:
## Expected Behavior / New Feature
Expected AllowedScopes to parse scope claim and match one or more values in string array similar to RouteClaimsRequirement.  For example JWT scope claim value has two scopes, space delimited by whitespace: `Values.Read Values.Write`

## Actual Behavior / Motivation for New Feature
`ScopesAuthoriser` compares the entire claim value to the AllowedScopes array causing it to fail to match either scope.  It looks like you addressed a similar issue in RouteClaimsRequirement.  Is that the workaround for this issue? Or, can AllowedScopes be modified to use that behavior?

## Steps to Reproduce the Problem

  1.  Generate a JWT with "scp" claim containing two or more scopes delimited by **whitespace**
  2.  Send JWT to Ocelot for route with a single AllowedScope
  3.  Allowed scope is not matched even though the scope exists in the claim

## Specifications
  - Version: Ocelot 13.5.0 and **19.0.2**
  - Platform: .NET Core 2.2 and **.NET 7**

jords1987:
Hi, @fsmullinslc @marctalary  will this feature be supported in upcoming versions?

fsmullin:
Hi @jords1987 @marctalary, I am not part of the development team that manages the pull requests.  I worked around the issue by using dependency injection.  I added a function that is called from ConfigureServices() 

`        protected void CustomizeOcelot(IServiceCollection services)
        {
            services.RemoveAll<IScopesAuthoriser>();
            services.TryAddSingleton<IScopesAuthoriser, DelimitedScopesAuthorizer>();
        }
`

Here is the source to the class that will parse the delimited scopes.  Hopefully this helps you.

`using System.Collections.Generic;
using System.Linq;
using System.Security.Claims;
using Ocelot.Authorisation;
using Ocelot.Infrastructure.Claims.Parser;
using Ocelot.Responses;

namespace MyGateway.Custom.Claims
{
    public class DelimitedScopesAuthorizer : IScopesAuthoriser
    {
        private readonly IClaimsParser _claimsParser;
        private readonly string _scope = "scope";
        private readonly string _msScope = "http://schemas.microsoft.com/identity/claims/scope";

        public DelimitedScopesAuthorizer(IClaimsParser claimsParser)
        {
            _claimsParser = claimsParser;
        }

[...]

PR #971, opened 2019-07-25
Support multiple scopes

marctalary:
Fixes / New Feature #913 

## Proposed Changes

  - Fix to support multiple which are in a single scope claim space delimited.
  https://openid.net/specs/openid-connect-basic-1_0.html#Scopes 
"Multiple scope values MAY be used by creating a space-delimited, case-sensitive list of ASCII scope values."


TomPallister:

Issue #993, opened 2019-08-19
403 - Http Options - AllowedScopes


andersondepaiva:
## Expected Behavior / New Feature
Return 200 - OK, in the cases that http method is OPTIONS in the route with AllowedScopes configured.

## Actual Behavior / Motivation for New Feature
Returned 403 - Forbbiden

## Steps to Reproduce the Problem
1. Ocelot-Confg
{
      "UpstreamPathTemplate": "/api/service-test/{everything}",
      "DownstreamPathTemplate": "/{everything}",
      "DownstreamScheme": "http",
      "DownstreamHostAndPorts": [
        {
          "Host": "localhost",
          "Port": 9012
        }
      ],
      "AuthenticationOptions": {
        "AuthenticationProviderKey": "IdentityApiKey",
        "AllowedScopes": [
          "service-test-portal"
        ]
      }
    }

2. Request OPTIONS
OPTIONS - http://localhost:8000/api/service-test/items

## Specifications

  - Version: 13.5.2

## Observations:
After modifications in the class Ocelot.Authorisation.Middleware.AuthorisationMiddleware for:
namespace Ocelot.Authorisation.Middleware
```c#
{
    using Configuration;
    using Logging;
    using Ocelot.Middleware;
    using Responses;
    using System.Threading.Tasks;

    public class AuthorisationMiddleware : OcelotMiddleware
    {
        private readonly OcelotRequestDelegate _next;
        private readonly IClaimsAuthoriser _claimsAuthoriser;
        private readonly IScopesAuthoriser _scopesAuthoriser;

        public AuthorisationMiddleware(OcelotRequestDelegate next,
            IClaimsAuthoriser claimsAuthoriser,
            IScopesAuthoriser scopesAuthoriser,
            IOcelotLoggerFactory loggerFactory)
            : base(loggerFactory.CreateLogger<AuthorisationMiddleware>())
        {
            _next = next;
            _claimsAuthoriser = claimsAuthoriser;
            _scopesAuthoriser = scopesAuthoriser;
        }

        public async Task Invoke(DownstreamContext context)
        {
            if (context.HttpContext.Request.Method.ToUpper() != "OPTIONS" && IsAuthenticatedRoute(context.DownstreamReRoute))
            {
                Logger.LogInformation("route is authenticated scopes must be checked");

                var authorised = _scopesAuthoriser.Authorise(context.HttpContext.User, context.DownstreamReRoute.AuthenticationOptions.AllowedScopes);

[...]

PR #996, opened 2019-08-23
Implementation for resolve issue #993

andersondepaiva:
Fixes / New Feature #

## Proposed Changes

  - Added IF to check if HTTP method is not OPTIONS
  - Issue described: https://github.com/ThreeMammals/Ocelot/issues/993

### DIFF
--- a/src/Ocelot/Authorisation/Middleware/AuthorisationMiddleware.cs
+++ b/src/Ocelot/Authorisation/Middleware/AuthorisationMiddleware.cs
@@ -25,7 +25,7 @@
 
         public async Task Invoke(DownstreamContext context)
         {
-            if (IsAuthenticatedRoute(context.DownstreamReRoute))
+            if (!IsOptionsHttpMethod(context) && IsAuthenticatedRoute(context.DownstreamReRoute))
             {
                 Logger.LogInformation("route is authenticated scopes must be checked");
 
@@ -52,7 +52,7 @@
                 }
             }
 
-            if (IsAuthorisedRoute(context.DownstreamReRoute))
+            if (!IsOptionsHttpMethod(context) && IsAuthorisedRoute(context.DownstreamReRoute))
             {
                 Logger.LogInformation("route is authorised");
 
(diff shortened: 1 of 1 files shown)


vmartos:
Hello, we are having a similar problems with the OPTIONS request. By default angular seems to add the OPTIONS before any other request, and since they are failing because of the authentication middleware it fails to do the final request. OPTIONS should not be handled by the authorization middleware.

andersondepaiva:
Ok, based me in the Authentication class implementation, that follow same solution 
```c#
public class AuthenticationMiddleware : OcelotMiddleware
    {
        private readonly OcelotRequestDelegate _next;

        public AuthenticationMiddleware(OcelotRequestDelegate next,
            IOcelotLoggerFactory loggerFactory)
            : base(loggerFactory.CreateLogger<AuthenticationMiddleware>())
        {
            _next = next;
        }

        public async Task Invoke(DownstreamContext context)
        {
            if (context.HttpContext.Request.Method.ToUpper() != "OPTIONS" && IsAuthenticatedRoute(context.DownstreamReRoute))
            {
                Logger.LogInformation($"{context.HttpContext.Request.Path} is an authenticated route. {MiddlewareName} checking if client is authenticated");

                var result = await context.HttpContext.AuthenticateAsync(context.DownstreamReRoute.AuthenticationOptions.AuthenticationProviderKey);

                context.HttpContext.User = result.Principal;

                if (context.HttpContext.User.Identity.IsAuthenticated)
                {
                    Logger.LogInformation($"Client has been authenticated for {context.HttpContext.Request.Path}");
                    await _next.Invoke(context);
                }
                else
                {
                    var error = new UnauthenticatedError(
                        $"Request for authenticated route {context.HttpContext.Request.Path} by {context.HttpContext.User.Identity.Name} was unauthenticated");

                    Logger.LogWarning($"Client has NOT been authenticated for {context.HttpContext.Request.Path} and pipeline error set. {error}");

                    SetPipelineError(context, error);
                }
            }
            else
            {
                Logger.LogInformation($"No authentication needed for {context.HttpContext.Request.Path}");

                await _next.Invoke(context);
            }
        }

        private static bool IsAuthenticatedRoute(DownstreamReRoute reRoute)
        {
            return reRoute.IsAuthenticated;
        }
    }
```

Thank's

PR #997, opened 2019-08-27
Add ConfigAwarePlaceholders class

donaldgray:
This allows placeholder values to be sourced from IConfiguration in
addition to set values.

New Feature

## Proposed Changes

I found myself writing a few `DelegatingHandlers` for appending headers for downstream calls using values from `IConfiguration`. This PR contains a new implementation of `IPlaceholders`. 

This decorates and replaces the `IPlaceholders` instaces in the IoC container to allow the existing placeholders to be used (it checks these first). The idea is to have config like:

```json
"UpstreamHeaderTransform": {
    "x-api-key": "{DownstreamService:ApiKey}"
}
```

I've added `AddConfigPlaceholders()` to `IOcelotBuilder` but I wasn't sure if that was the correct location, or if somewhere else would work.

### DIFF
--- a/src/Ocelot/DependencyInjection/IOcelotBuilder.cs
+++ b/src/Ocelot/DependencyInjection/IOcelotBuilder.cs
@@ -24,5 +24,7 @@ namespace Ocelot.DependencyInjection
 
         IOcelotBuilder AddTransientDefinedAggregator<T>()
             where T : class, IDefinedAggregator;
+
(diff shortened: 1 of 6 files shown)


donaldgray:
A quick note on this - I've not tested it in a 'real' app. I thought it best to get an opinion on it and I'll fully test it tomorrow.

PR #1015, opened 2019-09-17
Adding support for filtering on UpstreamScheme.

OrionDevelopment:
This PR adds support to filter on UpstreamScheme, just as you can already filter on UpstreamHost.
Allows for the routing of http upstream traffic to a specific downstream http handler, as well as upstream https traffic to a specific downstream https handler, by combining the UpstreamScheme and DownstreamScheme configuration entries for a given reroute.

Added the same unit tests as UpstreamHost has, and its logic is entirely similar.


TomPallister:

Issue #1001, opened 2019-08-28
Ocelot is not working with .NET Core 3


jsantanders:
## Expected Behavior 

the following method  extension
```
public static IOcelotBuilder AddOcelot(this IServiceCollection services);
public static IOcelotBuilder AddOcelot(this IServiceCollection services, IConfiguration configuration);
```
add the Ocelot Gateway to the AspNetCore project.

## Actual Behavior 
throws an exception 

```
System.TypeLoadException
  HResult=0x80131522
  Message=Could not load type 'Microsoft.Extensions.DependencyInjection.MvcJsonMvcCoreBuilderExtensions' from assembly 'Microsoft.AspNetCore.Mvc.Formatters.Json, Version=3.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'.
  Source=Ocelot
  StackTrace:
   at Ocelot.DependencyInjection.OcelotBuilder..ctor(IServiceCollection services, IConfiguration configurationRoot)
   at Ocelot.DependencyInjection.ServiceCollectionExtensions.AddOcelot(IServiceCollection services, IConfiguration configuration)
   at OcelotApiGw.Startup.ConfigureServices(IServiceCollection services) in <path to source code file> line 73
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at Microsoft.AspNetCore.Hosting.ConfigureServicesBuilder.InvokeCore(Object instance, IServiceCollection services)
   at Microsoft.AspNetCore.Hosting.ConfigureServicesBuilder.<>c__DisplayClass9_0.<Invoke>g__Startup|0(IServiceCollection serviceCollection)
   at Microsoft.AspNetCore.Hosting.StartupLoader.ConfigureServicesDelegateBuilder`1.<>c__DisplayClass15_0.<BuildStartupServicesFilterPipeline>g__RunPipeline|0(IServiceCollection services)
   at Microsoft.AspNetCore.Hosting.ConfigureServicesBuilder.Invoke(Object instance, IServiceCollection services)
   at Microsoft.AspNetCore.Hosting.ConfigureServicesBuilder.<>c__DisplayClass8_0.<Build>b__0(IServiceCollection services)
[...]

PR #1019, opened 2019-09-19
Fix TypeLoadException with AspNetCore 3.0

jkotalik:
Fixes https://github.com/ThreeMammals/Ocelot/issues/1001
 
Call AddJsonFormatters via reflection to avoid TypeLoadException. I verified this locally with a 3.0 app.


jkotalik:
cc @pranavkm @rynowak

PR #1033, opened 2019-10-11
Add  description fields to globalConfiguration and reRoutes

ChinaQin:
## Proposed Changes
-Add a description field to facilitate management of gateways and routes, and distinguish between gateways and routes displayed by the WebUI. Otherwise, the routes cannot be effectively distinguished when there are too many routes.


TomPallister:

PR #1046, opened 2019-10-30
Update InMemoryCache 

catcherwong:
Using `ConcurrentDictionary` to replace `Dictionary`


TomPallister:
@catcherwong I'm not sure if it needs to be concurrent. Why do you want this? 

I can see that there is potential bug, is concurrent dictionary slower than dictionary? I don't want to make performance worse.

catcherwong:
> I'm not sure if it needs to be concurrent. Why do you want this?

For cache, maybe concurrent will be more better than not concurrent.

> is concurrent dictionary slower than dictionary? 

Yes.

PR #1054, opened 2019-11-08
Feature/method transformer

pitming:
Fixes / New Feature #
Add the posibility to modify the HttpAction

## Proposed Changes

  - This is the first step from a developpement I'm working on. The idea is to be able to parse request and response with Liquid Template
  -
  -


TomPallister:
I dont understand this PR the method is always set from the http request, if the objective is to allow an upstream request to be a POST and then turn it into a PUT this would need another approach. I'm going to close this for now.

pitming:
That's exactly the point.
We are in the business of integrating a very poorly designed "Rest" api where all method are POST. 

Actually, we use Ocelot with that allow use to rewrite all the call.

But we wanted to be able to change Action and then generate body through Liquid templates.

We actually post ponned the developement, but all suggestion are welcome

TomPallister:
merged into master

Issue #603, opened 2018-09-07
Respond with an error on no connection to downstream service


Deilan:
## New Feature
Respond with a reasonable error message in a case when _No connection could be made because the target machine actively refused it_. And add an extension point to override the message and status code (e. g. 502 Bad Gateway) for this case of course.

The error usually looks like the following in logs:
```
warn: Ocelot.Requester.Middleware.HttpRequesterMiddleware[0]
      requestId: 0HLGKQK19POEE:00000002, previousRequestId: no previous request id, message: Error making http request, exception: System.Net.Http.HttpRequestException: No connection could be made because the target machine actively refused it ---> System.Net.Sockets.SocketException: No connection could be made because the target machine actively refused it
         at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)
         --- End of inner exception stack trace ---
         at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)
         at System.Threading.Tasks.ValueTask`1.get_Result()
         at System.Net.Http.HttpConnectionPool.CreateConnectionAsync(HttpRequestMessage request, CancellationToken cancellationToken)
         at System.Threading.Tasks.ValueTask`1.get_Result()
         at System.Net.Http.HttpConnectionPool.WaitForCreatedConnectionAsync(ValueTask`1 creationTask)
         at System.Threading.Tasks.ValueTask`1.get_Result()
         at System.Net.Http.HttpConnectionPool.SendWithRetryAsync(HttpRequestMessage request, Boolean doRequestAuth, CancellationToken cancellationToken)
         at System.Net.Http.DiagnosticsHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
         at System.Net.Http.HttpClient.FinishSendAsyncBuffered(Task`1 sendTask, HttpRequestMessage request, CancellationTokenSource cts, Boolean disposeCts)
         at Ocelot.Requester.HttpClientHttpRequester.GetResponse(DownstreamContext context)
```

## Motivation for New Feature
More smooth UX.

## Steps to Reproduce the Problem
Reroute to a host/port, which doesn't run a web server.

TomPallister:
@Deilan what would you suggest for the error message?

Deilan:
@TomPallister 
"_No connection could be made because the target machine actively refused it_" would be pretty good,
[...]

PR #1074, opened 2019-12-04
Feature/603 bad gateway

jlukawska:
Fixes / New Feature #603 Respond with an error on no connection to downstream service

## Proposed Changes

  - return 502 Bad Gateway status code when request can not be completed
  -
  -


TomPallister:
@jlukawska thanks for the PR, I'm not sure this is the correct behaviour because the error will be raised after making a http request and getting any exception from HttpClient.SendAsync which might not be caused by the service being called! What do you think?

jlukawska:
Yes, you're right. So I've changed this to look for HttpRequestException only.

TomPallister:
@jlukawska should Ocelot just not return the error from the service? Given that Ocelot is just a proxy? I'm really not sure :(

jlukawska:
This post: https://github.com/ThreeMammals/Ocelot/issues/603#issuecomment-441811416 and my little research convinced me that it makes sense, but the decision is yours of course.

TomPallister:

PR #1080, opened 2019-12-12
Do not crash host on Dispose

jmezach:
During the .NET Core 3.0 upgrade I introduced this Dispose method on the FileConfigurationPoller. It turns out that the implementation was a bit problematic, because of the timer hasn't been created yet (since that happens in the StartAsync) and the instance is being disposed it throws an exception here.

In our case, this happened because Consul wasn't available at the time our Ocelot based application was starting. This caused an exception during startup, but the actual exception was being masked by this bug since the host was being disposed due to a startup error, but it couldn't be properly disposed due to this bug. This failed spectacularly in our production environment because our API gateway failed to start.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -1,4 +1,4 @@
-﻿#tool "nuget:?package=GitVersion.CommandLine"
+﻿#tool "nuget:?package=GitVersion.CommandLine&version=5.0.1"
 #tool "nuget:?package=GitReleaseNotes"
 #addin nuget:?package=Cake.Json
 #addin nuget:?package=Newtonsoft.Json
--- a/src/Ocelot/Configuration/Repository/FileConfigurationPoller.cs
+++ b/src/Ocelot/Configuration/Repository/FileConfigurationPoller.cs
@@ -105,7 +105,8 @@ namespace Ocelot.Configuration.Repository
 
         public void Dispose()
         {
-            _timer.Dispose();
+            _timer?.Dispose();
+            _timer = null;
         }
     }
 }
--- a/test/Ocelot.UnitTests/Configuration/FileConfigurationPollerTests.cs
+++ b/test/Ocelot.UnitTests/Configuration/FileConfigurationPollerTests.cs
@@ -41,14 +41,14 @@ namespace Ocelot.UnitTests.Configuration
             _internalConfigCreator = new Mock<IInternalConfigurationCreator>();
             _internalConfigCreator.Setup(x => x.Create(It.IsAny<FileConfiguration>())).ReturnsAsync(new OkResponse<IInternalConfiguration>(_internalConfig));
             _poller = new FileConfigurationPoller(_factory.Object, _repo.Object, _config.Object, _internalConfigRepo.Object, _internalConfigCreator.Object);
-            _poller.StartAsync(new CancellationToken());
         }
 
         [Fact]
         public void should_start()
         {
-            this.Given(x => ThenTheSetterIsCalled(_fileConfig, 1))
-                 .BDDfy();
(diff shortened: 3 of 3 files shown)


jmezach:
@TomPallister I think this is a rather critical fix. Would you mind taking a look here?

jmezach:
Interesting, the Travis builds are failing due to an issue with GitVersion. I'm guessing something has changed there, and we're not pinning the version if GitVersion.CommandLine we want to use, so I guess this could happen.

The AppVeyor build succeeded on my first PR, and now tests are failing that are unrelated to my change. I'm also unable to reproduce those test failures locally. Maybe a flaky test?

jmezach:
Okay, looks like pinning the version of GitVersion.CommandLine solved the issue and builds are both green now. Looks like there's a flaky test in there somewhere though.

jmezach:
@thiagoloureiro  Do you have time to take a look here? This is an important fix since the bug has the ability to completely crash the host leaving it in an unknown state, causing ANCM timeouts when hosting in IIS.

jmezach:
@TomPallister Thanks for merging this. How are things going with this project by the way? I would love to be a bit more involved here. I've been building an API gateway based on Ocelot for my company which is very close to going live.

TomPallister:
@jmezach after about a year away from the project I'm returning. That sounds awesome I hope it hasn't been to difficult!!!

I'm just making the build pipeline a bit easier to use and going through open PRs.

I think Ocelot's code base needs some TLC, I also think the project might benefit from a major version upgrade where the names of some things are considered. I don't think ReRoute is a very good name for example! I literally just chose it when I was hacking away 3-4 years ago. A lot of this stuff is pretty standard in other projects. I also think we can make Ocelot a lot faster for some basic use cases with the new features since I last used c# & dotnet! Pretty excited to try some stuff out :)

jmezach:
That is great news. And yes, I would love to be invited to the team. Honestly it's been a great experience using Ocelot to build our API gateway.

I think Ocelot serves its purpose as a great foundation to build an API gateway on top off. There's definitely room for improvement though, as you mentioned. Configuration is something that has been bothering me personally, since it can become quite verbose as you put more services behind it. I think I suggested some kind of templating system before in an issue somewhere.

And of course there are quite a couple of new things in .NET Core 3.0 that could benefit Ocelot. Out of the top of my head, things like the IHttpClientFactory, Endpoint routing and probably the new low allocation pipelines stuff as well.

Plenty of ideas I guess. Curious as to what ideas you have ;)

TomPallister:
Sounds good, I have sent you an invite. Yeh the configuration system is a bit meh, you are basically just saying what you want the classes to look like in json. Out of interest are you using docker and some kind of orchestration system? I'm wondering if we should make Ocelot more "cloud native"!

jmezach:
@TomPallister Cool, I have accepted the invite. Do you have a place where the team hangs out to discuss things? Or is it all on GitHub?

As for your question, at the moment we're still running on Windows VM's natively. This is something we would like to change, but a big hurdle for us is our authentication mechanism. At the moment we're still using the old Windows Identity Foundation stuff. We would like to replace that, but that would mean having to do a big bang across our entire landscape of services. This is exactly why we're using Ocelot to build an API gateway to handle the authentication stuff, which can then delegate to downstream services using a more modern mechanism, while keeping the existing mechanism in place from the outside. Eventually we can then move over everything to something like OAuth2. At least, that's the plan. In the mean time, we're also looking at Docker and Kubernetes.

TomPallister:
OK cool sounds good, there is a slack organisation in the readme (just added) and we also communicate on the core team sometimes. Other than that just issues etc.

PR #1110, opened 2020-01-21
Update README.md

jmezach:
Fix the broken link to Slack. It was a relative link, now it is an absolute one.


jmezach:
@TomPallister By the way, how do I sign-up to that Slack workspace? Or do you send out invites? I believe Slack has the ability to allow users to sign-up themselves, which can then be approved by a team admin or something like that.

shahargl:
@jmezach how do I get invite for the slack?

Issue #1129, opened 2020-02-12
CookieStickySessions is not working with Kubernetes service discovery


navinv789:
## Steps to Reproduce the Problem

I am using Kubernetes service discovery to discover the pods and  CookieStickySessions load balancer. 
CookieStickySessions  is working fine if we are adding pod ip's in DownstreamHostAndPorts but when we are using Kubernetes service discovery the traffic is not been send to the desired pods. pods are selected randomly. 

Due to some business requirement we cant use redish cache or other alternatives.

## Specifications

  - Version:13.5.2
  - Platform: dot net core 2.2
  - Subsystem:

ussamoo:
I think it's related to this issue #1126

TomPallister:
fixed by #1134

Issue #1126, opened 2020-02-10
Question about Ocelot LoadBalancer and Kubernetes Service Discovery Provider


ussamoo:
## Expected Behavior / New Feature

When using Kubernetes service discovery and LoadBalancer of type RoundRobin, Ocelot balance traffic beteween running pods.

## Actual Behavior / Motivation for New Feature

Ocelot send traffic to the same pod without balancing. 

## Steps to Reproduce the Problem

  1. Use the configuration below 
   {
  "GlobalConfiguration": {
    "ServiceDiscoveryProvider": {
      "Namespace": "my-service-name",
      "Type": "Kube"
    }
  },
  "ReRoutes": [
    {
      "DownstreamPathTemplate": "/{everything}",
      "DownstreamScheme": "http",
      "ServiceName": "my-service-name",
      "UpstreamPathTemplate": "/api/{everything}",
      "LoadBalancerOptions": {
        "Type": "RoundRobin"
      },
      "UpstreamHttpMethod": []
    }
  ]
}
  2. Run 3 instance of a pod having as name "my-pod-name"
  3. Create a service named "my-service-name" having as Selector "my-pod-name"

## Specifications

  - Version: 14.0.5
  - Platform: Linux
  - Subsystem:

Ocelot KubeProvider gets the information (IP and Port) correctly about the service named "my-service-name" and sends traffic TO THE SERVICE using the Ip/port.

The first issue here is that the service sends traffic always to the same pod and the second issue is that the Load Balancing is not controlled by Ocelot.

Isn't better to discover pods endpoints based on the service name and send traffic directly to pod endpoint ??

I tested this scenario with Ocelot and it works well, but i'm not sure if it's a good idea.

But doing so means that the Load Balancing can be controlled by Ocelot.

enriko-riba:
I know this is old and you already provided a PR but just to clarify the issue. 

There is nothing wrong in sending the traffic to the virtual service IP but above the TCP/IP routing is the HTTP layer with the connection concept. 
Basically once you establish a HTTP connection with an endpoint you are stuck with the same connection/endpoint. The connection is kept alive until the pod resets it (which is not going to happen) or the connection times out (http linger, keepalive). IIRC the timeout is like 4 minutes.

[...]

PR #1134, opened 2020-02-16
Update Kubernetes Service Discovery Provider to use pods endpoints

ussamoo:
Fixes / New Feature #

This PR adress this two issues : #1126 and #1129 

## Proposed Changes

Updated the KubeProvider to build the list of services based on the endpoints of a given kubernetes service in a specific namespace.

This will fix the issues related to the load balancing since Ocelot will send traffic directly to the pod endpoint instead of the service.

### DIFF
--- a/src/Ocelot.Provider.Kubernetes/KubeApiClientExtensions/EndPointClientV1.cs
+++ b/src/Ocelot.Provider.Kubernetes/KubeApiClientExtensions/EndPointClientV1.cs
@@ -1,0 +1,41 @@
+﻿using HTTPlease;
+using KubeClient;
+using KubeClient.Models;
+using KubeClient.ResourceClients;
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Ocelot.Provider.Kubernetes.KubeApiClientExtensions
+{
+    public class EndPointClientV1 : KubeResourceClient
+    {
+        public EndPointClientV1(IKubeApiClient client) : base(client)
+        {
+        }
+
+        public async Task<EndpointsV1> Get(string serviceName, string kubeNamespace = null, CancellationToken cancellationToken = default)
+        {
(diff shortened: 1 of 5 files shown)


enriko-riba:
:+1:

TomPallister:
@ussamoo sorry it has taken me so long to merge this. 

Thanks for your help with Ocelot!!!!!

chazt3n:
Are we sure this is right? We are having the exact behavior this _sounds_ like it was supposed to fix.

The k8s service endpoint exists specifically to distribute load. Just trying to understand because we've looked up and down the issues/docs here and can't tell 

a) what ocelot is actually supposed to do and
b) why are we routing traffic to the same pod during burst traffic scenarios (the only time we care about balancing load)

enriko-riba:
> Are we sure this is right? We are having the exact behavior this _sounds_ like it was supposed to fix.
> 
Do you think the implementation is wrong or just asking about the behavior?
The original issue with load balancing was that once a client connects and starts sending requests, all requests are delivered to the same pod due to the persistent HTTP connection.
Persistent connections are default since HTTP/1.1 but even with 1.0 there is a "keep-alive" option so there is no way around except closing the connection client side every time. Closing the connection tremendously impacts performance since more time is spent establishing connections than transferring data while leaving it open causes a single pod to serve all traffic from a given client.
The PR is supposed to enable load balancing across all endpoints behind a service object as returned by K8s API.

I haven't tested the new behavior but please share your observations and how you tested.

chazt3n:
Hey there, we found that we had to add the load balancer options and specify round robin on each route to get load balancing - after that, it does appear to be working.

Previously, it did _some_ load balancing, but during burst traffic it would single out a pod. I hope this helps

We tested in AWS EKS and use cloudwatch logs to see which pods were accepting the requests. Pretty low tech, but that's how we spotted the issue and it was fine enough

PR #1148, opened 2020-03-10
Feature/1115 find available port in acceptance tests

### LINKED ISSUE #1115: SocketException on binding to localhost using several ports when running Ocelot acceptance tests

All acceptance tests can be run successfully independently from Windows configuration.

Since I installed Docker and Hyper-V on my machine, several acceptance tests crash on binding to localhost. The exception is:

```
System.IO.IOException : Failed to bind to address http://localhost:51388.
    ---- System.AggregateException : One or more errors occurred. (An attempt was made to access a socket in a way forbidden by its access permissions.) (An attempt was made to access a socket in a way forbidden by its access permissions.)
    -------- System.Net.Sockets.SocketException : An attempt was made to access a socket in a way forbidden by its access permissions.
```
    
First, I thought that something opened these ports, but they are not opened (netstat does not list them as used). It turned out that they are reserved by Hyper-V. And what worse, the reserved port ranges can change after system reboot. 

Here you can find a discussion about this problem and possible workarounds on Docker github: https://github.com/docker/for-win/issues/3171

I would like to start a discussion about that issue and possible solutions. Maybe ports to use in the acceptance tests should be randomly generated  and checked if the binding to them is possible. If the problem with binding to the port exists, a new port should be generated. What do you think?

I think this is quite important, because there is high probability that when you have Hyper-V enabled you will not be able to run all the tests successfully.

Maybe this issue is the reason for changes made in the following commit 07df311094a25329153dffa451d63c8036c4d77c ?

  1. Enable Hyper-V in Windows
  1. Run all Ocelot acceptance tests

## Specifications

  - Platform: Windows 10 Pro

jlukawska:
Fixes #1115 SocketException on binding to localhost using several ports when running Ocelot acceptance tests 

## Proposed Changes
I've implemented the solution that I suggested in the issue: all acceptance tests use randomly generated ports which are first checked to see if they can be used. 
This will prevent the tests from crashing on binding to the ports which are unavailable because of hyper-v reservation or for other reasons.


TomPallister:
@jlukawska looks good thank you!!! Will merge ASAP.

TomPallister:
I merged this manually into master

PR #1149, opened 2020-03-11
Feature/603 bad gateway

jlukawska:
Fixes / New Feature #603 Respond with an error on no connection to downstream service

## Proposed Changes

  - return 502 Bad Gateway status code when request can not be completed
  -
  -


TomPallister:
Thanks @jlukawska can you merge master into your branch? The build should pass!

src/Ocelot/Requester/HttpExeptionToErrorMapper.cs
```diff
@@ -28,6 +30,11 @@ public Error Map(Exception exception)
                 return new RequestCanceledError(exception.Message);
             }
 
+            if (type == typeof(HttpRequestException))
+            {
+                return new ConnectionToDownstreamServiceError(exception);
+            }
```
kjellegafvelin: Maybe it would be good to check if the inner exception is of type **SocketException** and the **SocketErrorCode** is **HostNotFound**.

jlukawska: I considered checking the inner exception and ConnectionRefused error code (the error code from the issue #603), but when I analyzed this issue more, I decided to generalize these conditions to catch other situations where there is no connection to the downstream service.

See: https://github.com/ThreeMammals/Ocelot/issues/603

kjellegafvelin: Ok, sounds reasonable.

TomPallister:
@jlukawska thanks for this feature. Sorry it has taken me so long to merge!

PR #1162, opened 2020-03-22
Added UpstreamHost placeholder to identify host from which request originated

samirsyed:
New Feature #
- Identify host from which request originated, if upstream host based routing is used

## Proposed Changes
- Added "{UpstreamHost}" key in placeholders
- Added unit test case in PlaceholdersTests


TomPallister:
@samirsyed thanks for the PR :) sorry it took me so ages to merge it! I added a few extra tests and updated the docs :)

PR #1273, opened 2020-06-26
The RequestAborted property is not copied in MultiplexingMiddleware

RayRoderic:
## Fixes
The `RequestAborted`  property needs to be copied in target object on `Copy` method,  so that the request can be canceled.

## Proposed Changes
  - The `RequestAborted` property of `HttpRequest` is copied from source to target request object


raman-m (approved):
Thank you for the PR!

raman-m (changes requested):
No tests to cover this case!

raman-m:
Duplicate of #1367

PR #1300, opened 2020-07-25
When the downstream service is not found, Consul provider throws an exception

HonghuiHu:
### Case 1: favicon.ico request
`ResponderMiddleware` generates warning for the route `/favicon.ico`

### Case 2: empty route in service discovery setup by Consul provider
`ExceptionHandlerMiddleware` generates exception for the route `/` in Consul setup.

## Fixes
???

## Proposed Changes
???

### DIFF
--- a/src/Ocelot/LoadBalancer/LoadBalancers/RoundRobin.cs
+++ b/src/Ocelot/LoadBalancer/LoadBalancers/RoundRobin.cs
@@ -1,12 +1,9 @@
+using Microsoft.AspNetCore.Http;
+using Ocelot.Responses;
+using Ocelot.Values;
 ﻿using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
-
-using Microsoft.AspNetCore.Http;
-
-using Ocelot.Responses;
-
-using Ocelot.Values;
 
 namespace Ocelot.LoadBalancer.LoadBalancers
 {
--- a/src/Ocelot/LoadBalancer/LoadBalancers/RoundRobin.cs
+++ b/src/Ocelot/LoadBalancer/LoadBalancers/RoundRobin.cs
@@ -26,8 +26,8 @@ namespace Ocelot.LoadBalancer.LoadBalancers
             {
                 if (services.Count < 1)
                 {
-                    //When the downstream service is not found, LeastConnection prompts Warn and RoundRobin throws an exception。eg:/favicon.ico
-                    return new ErrorResponse<ServiceHostAndPort>(new ServicesAreEmptyError($"services were empty for {httpContext.Request.Path}"));
+                    // When the downstream service is not found, LeastConnection prompts Warn and RoundRobin throws an exception, eg: /favicon.ico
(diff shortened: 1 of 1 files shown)


HonghuiHu:
When the downstream service is not found, LeastConnection prompts Warn and RoundRobin throws an exception。
image

src/Ocelot/LoadBalancer/LoadBalancers/RoundRobin.cs
```diff
@@ -27,6 +24,12 @@ public async Task<Response<ServiceHostAndPort>> Lease(HttpContext httpContext)
             var services = await _services();
             lock (_lock)
             {
+                if (services.Count < 1)
+                {
+                    // When the downstream service is not found, LeastConnection prompts Warn and RoundRobin throws an exception, eg: /favicon.ico
```
raman-m, line 29: Truly speaking this case of icon should not be processed by this method.
Can we filter such cases in upper contexts?

src/Ocelot/LoadBalancer/LoadBalancers/RoundRobin.cs
```diff
@@ -27,6 +24,12 @@ public async Task<Response<ServiceHostAndPort>> Lease(HttpContext httpContext)
...
+                    return new ErrorResponse<ServiceHostAndPort>(new ServicesAreEmptyError($"Services were empty for {httpContext.Request.Path}"));
```
raman-m, line 30: Does this error returning fix the bug with exception generation by global exception handler?
How will `ErrorResponse` prevent exception generating in upper contexts?

raman-m (changes requested):
Hi Honghui!
Thanks for your interest in Ocelot!

Let's start the discussion...
Currently I am not sure the PR contains right solution for extra requests coming from browsers. 😄

raman-m (changes requested):
Missing unit tests and at least one acceptance test for the user's scenario!

raman-m:
@ggnaegi 
Look at this PR please!
What do you think?..

ggnaegi (commented):
@raman-m I think the problem was that round robin behaved differently than other load balancers.
We would expect ``` new ErrorResponse<ServiceHostAndPort>(new ServicesAreEmptyError("There were no services in LoadBalancer")); ``` if no services are available. It seems to me that this is now fixed in the develop branch and this PR could be closed.

``` c#
 var services = await _servicesDelegate?.Invoke() ?? new List<Service>();

            if (services?.Count != 0)
            {
                lock (_lock)
                {
                    if (_last >= services.Count)
                    {
                        _last = 0;
                    }

                    var next = services[_last++];
                    return new OkResponse<ServiceHostAndPort>(next.HostAndPort);
                }
            }

            return new ErrorResponse<ServiceHostAndPort>(new ServicesAreEmptyError($"There were no services in {nameof(RoundRobin)} during {nameof(Lease)} operation."));
```
https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/LoadBalancer/LoadBalancers/RoundRobin.cs

raman-m:
@HonghuiHu
This issue is outdated, and it is already fixed in **develop** branch.
Please, sync your forked repo, and/or download Ocelot v[20.0](../releases) NuGet package

PR #1403, opened 2020-12-28
Override the `ClaimsToHeadersMiddleware` by the `OcelotPipelineConfiguration` settings

kesskalli:
## New Feature
The idea is to allow to use custom claims to headers transformation logic after the authorization process is done.
The transformation using a JSON file requires to handle everything by route. The file will become very verbose in this case.
In some cases, the internal HTTP services can expect some HTTP "X-" Headers to get more context about the request (to support multi-tenancy for instance), the override will allow to add them automatically.

## Proposed Changes
  - Allowed to override the ClaimsToHeaders middleware
  - The override can be "injected" in the Ocelot config on start-up


EngRajabi (approved):
ok

ImiyaPurnima (approved):

raman-m:
@kesskalli commented [on Dec 28, 2020](#issue-775322959):
> ## Proposed Changes
>  - Allowed to override the ClaimsToHeaders middleware

I see these changes. The community will review them soon.

---
>  - The override can be "injected" in the Ocelot config on start-up 

Yeah, but in theory. And I **don't** see these changes in files.
It would be a good idea to add a [sample web app](../tree/develop/samples) to demonstrate how to override this middleware.

---

Also, don't forget to write [tests](#pullrequestreview-1524990460) please!

raman-m (approved):

PR #1424, opened 2021-01-27
Second parameter of Substring is wrong in HeaderFindAndReplaceCreator

fku-:
Second parameter of `Substring` was wrong: calculation of value for the `int length` parameter was wrong.
Existing implementation works only for replacements at the beginning of a string `startOfPlaceholder = 0`.

## Proposed Changes

  - Allow replacements inside a header, not only at the beginning
  - **Update 1** (Sep 30, 2023). Allow replacements inside of placeholder pattern, not only at the beginning
  - Add unit test to check this fix

### DIFF
--- a/src/Ocelot/Configuration/Creator/HeaderFindAndReplaceCreator.cs
+++ b/src/Ocelot/Configuration/Creator/HeaderFindAndReplaceCreator.cs
@@ -71,7 +71,7 @@ namespace Ocelot.Configuration.Creator
 
         private Response<HeaderFindAndReplace> Map(KeyValuePair<string, string> input)
         {
-            var findAndReplace = input.Value.Split(",");
+            var findAndReplace = input.Value.Split(',');
 
             var replace = findAndReplace[1].TrimStart();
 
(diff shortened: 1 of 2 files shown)


raman-m:
Related to
- #1659

src/Ocelot/Configuration/Creator/HeaderFindAndReplaceCreator.cs
```diff
@@ -81,7 +80,7 @@ private Response<HeaderFindAndReplace> Map(KeyValuePair<string, string> input)
             {
                 var endOfPlaceholder = replace.IndexOf("}", startOfPlaceholder, StringComparison.Ordinal);
```
RaynaldM, line 81:
```suggestion
                var endOfPlaceholder = replace.IndexOf('}', startOfPlaceholder, StringComparison.Ordinal);

```

RaynaldM: and you could do same thing at line 74

RaynaldM (approved):

src/Ocelot/Configuration/Creator/HeaderFindAndReplaceCreator.cs
```diff
@@ -1,7 +1,9 @@
 using Ocelot.Configuration.File;
 using Ocelot.Infrastructure;
 using Ocelot.Logging;
-using Ocelot.Responses;
+using Ocelot.Responses;
+using System;
+using System.Collections.Generic;
```
raman-m, line 6: Redundant!
```suggestion
```

raman-m (changes requested):

raman-m (dismissed):

raman-m (approved):

PR #1433, opened 2021-02-08
Disallow request body for HEAD, GET, TRACE web methods

### LINKED ISSUE #1432: Disallow body in HEAD, GET, TRACE

Downstream HEAD, GET & TRACE requests should not allow a request body.

If you mistakenly send a request body on HEAD, GET, TRACE requests they forward the body downstream.

More importantly if you map an upstream POST, PUT, etc with a request body to a downstream method that should not support a request body, the body is still sent.

  1. Configure an upstream POST to a downstream GET
  2. Tests in the https://github.com/ThreeMammals/Ocelot/pull/1433 PR fail in the current develop branch
  
## Specifications

  - Version: 17.0.0

abriening:
## Closes #1432 
- #1432 

## Proposed Changes
  - Ignores request body for requests with upstream HEAD, GET, TRACE (spec **must** not have body)
  - Ignores request body for requests with downstream HEAD, GET, TRACE; allows safe mapping from requests with body to requests without
  - Removes reading content as `ByteArray`, use `StreamContent`
  - `IRequestMapper.Map` is synchronous since it just forwards the input stream
  - Only forward `Content-*` headers if body is forwarded, `Content-Length` is assumed correct from upstream since we're no longer reading the stream


abriening:
Build error seems unrelated to changes:
```
/root/project/build.cake(98,43): error CS1739: The best overload for 'DeleteDirectory' does not have a parameter named 'recursive'
Error: Error(s) occurred when compiling build script:
/root/project/build.cake(98,43): error CS1739: The best overload for 'DeleteDirectory' does not have a parameter named 'recursive'
Makefile:4: recipe for target 'build' failed
```

abriening:
Can rebase against https://github.com/ThreeMammals/Ocelot/pull/1436 to fix CI.

abriening:
Can rebase against https://github.com/ThreeMammals/Ocelot/pull/1436 to fix CI.

WeihanLi:
Maybe add `DELETE`, and I think some apps are using `GET` with body, maybe we could have a switch in the configuration if we should check http method strictly before copy body

abriening:
`DELETE` may have bodies per http 1.1 spec, `GET` must not. Two main goals were shortcut on the body copy code when there shouldn't be a body and don't create weird non-compliant requests when your upstream & downstream methods don't match.

I could see a case for the `BodyIsSupported` having configurable methods. I can take a stab at that if you think it's worthwhile @WeihanLi.

WeihanLi:
Cool, as for the `BodyIsSupported` configuration, maybe we look for others consideration, especially project owner @TomPallister (I could not merge the PR)

src/Ocelot/Request/Mapper/RequestMapper.cs
```diff
@@ -41,29 +42,35 @@ public async Task<Response<HttpRequestMessage>> Map(HttpRequest request, Downstr
             }
         }
 
-        private static async Task<HttpContent> MapContent(HttpRequest request)
+        private static bool BodyIsSupported(string method)
         {
-            if (request.Body == null || (request.Body.CanSeek && request.Body.Length <= 0))
+            return method != HttpMethod.Get.Method
```
RaynaldM, line 47: I don't agree, you can put a body in a get request, it's not standard & recommended, but it's possible.
Ocelot must remain open to this kind of "exotic" case.

RaynaldM (changes requested):

src/Ocelot/Request/Mapper/RequestMapper.cs
raman-m: Ray,
Will be a good idea to provide config option (settings) to ignore (to include) bodies of mentioned web methods to follow (not follow) web standards?

src/Ocelot/Request/Mapper/RequestMapper.cs
```diff
@@ -41,29 +42,35 @@ public async Task<Response<HttpRequestMessage>> Map(HttpRequest request, Downstr
...
+                && method != HttpMethod.Head.Method
+                && method != HttpMethod.Trace.Method;
+        }
+
+        private static HttpContent MapContent(string downstreamMethod, HttpRequest request)
+        {
+            if (!BodyIsSupported(downstreamMethod)
+                || !BodyIsSupported(request.Method))
             {
```
WeihanLi, line 56: Should we log an error or warning when the request body not supported

src/Ocelot/Request/Mapper/RequestMapper.cs
WeihanLi: Maybe this should be configured from the configuration so that we could change it easily if needed

src/Ocelot/Request/Mapper/RequestMapper.cs
raman-m: It depends on implementation! 😄 
But logging warning is better.
Logging error is wrong because we can still make a request to downstream. Only web server decides what to do with a body. So, in case of web server returned error status we can find the warning in Ocelot's log and tune the route to make the route stable.

src/Ocelot/Request/Mapper/RequestMapper.cs
RaynaldM: One more configuration option for a false problem.
Ocelot is designed to meet a wide range of needs (for example, we're using it to deal with an old API that didn't respect any standards or rules), so it's already complex enough to add a new option.
That's just my opinion.

ggnaegi: https://www.rfc-editor.org/rfc/rfc7231#section-4.3.1

As for the GET with body, we could mention elasticsearch...

raman-m: @RaynaldM and @ggnaegi 
As I see, we have a compromise to allow body for GET requests. In this case the PT doesn't have much sense...
Hmmm... Let me think a bit more...

test/Ocelot.AcceptanceTests/HttpTests.cs
```diff
@@ -103,7 +103,7 @@ public void should_return_response_200_when_using_http_two_point_zero()
                         DownstreamPathTemplate = "/{url}",
                         DownstreamScheme = "https",
                         UpstreamPathTemplate = "/{url}",
-                        UpstreamHttpMethod = new List<string> { "Get" },
+                        UpstreamHttpMethod = new List<string> { "Put" },
```
raman-m, line 106: What are doing, man? @abriening 
Current test should remain unchanged because it was written for GET case.

raman-m (changes requested):
@abriening
Current tests for GET case should not be changed with PUT.
You have to add absolutely new tests to cover the feature!

abriening:
Closing. Just use YARP.

raman-m:
😢

PR #1502, opened 2021-08-03
fix OcelotBasic sample to make it work correctly

chenbojian:
# Fixes
The OcelotBasic sample do not work because the `Routes` was `ReRoutes` in the previous version. Upgrading to use the latest version make it works.


raman-m:
@chenbojian  Hi Chen!
This change is included to my PR #1663
Going to close your micro PR as duplicate with merge conflicts.

raman-m:
Fixed in #1663

PR #1505, opened 2021-08-19
Ignore port when using UpstreamHost option

grssmnn:
# Fixes
I wanted to use the `UpstreamHost` option and found out that Ocelot is handling the port as part of the hostname (when running on a different port than :80/:443). For my understanding this is not the expected behaviour since any other software like nginx or traefik ignore the port when talking about the hostname.

### DIFF
--- a/src/Ocelot/DownstreamRouteFinder/Middleware/DownstreamRouteFinderMiddleware.cs
+++ b/src/Ocelot/DownstreamRouteFinder/Middleware/DownstreamRouteFinderMiddleware.cs
@@ -27,7 +27,10 @@ namespace Ocelot.DownstreamRouteFinder.Middleware
 
             var upstreamQueryString = httpContext.Request.QueryString.ToString();
 
-            var upstreamHost = httpContext.Request.Headers["Host"].ToString().Split(":")[0];
+            var hostHeader = httpContext.Request.Headers["Host"].ToString();
+            var upstreamHost = hostHeader.Contains(':')
+                ? hostHeader.Split(':')[0]
+                : hostHeader;
 
             Logger.LogDebug($"Upstream url path is {upstreamUrlPath}");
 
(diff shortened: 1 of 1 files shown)


raman-m:
Hi Kevin!

What issue is this PR related to?

raman-m:
@grssmnn 
https://ocelot.readthedocs.io/en/latest/features/routing.html#upstream-host
#209 
The links above are the context.

I cannot get your proposal!.. You compare nginx & traefik but they are different networking tools than Ocelot.
If you want to ignore the port for Ocelot deploying scenario then just use fix from #209 via specifying pure domain name in [Upstream Host](https://ocelot.readthedocs.io/en/latest/features/routing.html#upstream-host) property of the route.
If your Ocelot hosting scenario uses a port, just do not specify Upstream Host property.
If your Ocelot hosting scenario doesn't use a port, just specify the host value of registered domain name. If you have localhost scenario, just register extra domain via OS functionality and web server.
But I cannot imagine multiple Ocelot gateways at one web hosting without special domain name registration.

Could you describe your user scenario please?
It seems it can be a new feature of Ocelot, but with rare usage.

raman-m:
>@grssmnn commented [on Aug 19, 2021](#issue-974477649)

Specifying different ports for Upstream of routes will be completely different Ocelot's feature. Would you like to open new discussion as an issue? 😉 

But your small fix cares about the hostname...
Is it the problem?
How does it influence to your user scenario?

src/Ocelot/DownstreamRouteFinder/Middleware/DownstreamRouteFinderMiddleware.cs
```diff
@@ -34,7 +34,7 @@ public async Task Invoke(HttpContext httpContext)
 
             var upstreamQueryString = httpContext.Request.QueryString.ToString();
 
-            var upstreamHost = httpContext.Request.Headers["Host"];
+            var upstreamHost = httpContext.Request.Headers["Host"].ToString().Split(":")[0];
```
raman-m, line 37: Better to test for colon sign `:` first to ensure that a port is specified. 👇 

```suggestion
            var hostHeader = httpContext.Request.Headers["Host"].ToString();
            var upstreamHost = hostHeader.Contains(':')
                ? hostHeader.Split(':')[0]
                : hostHeader;
```

RaynaldM (approved):

raman-m (approved):

PR #1520, opened 2021-09-22
Documentation: Updated text about AllowedScopes from "all of " to "at least one of"

roseeng:
In the docs it is said that if you configure AllowedScopes for a Route, the caller needs to have all scopes. 
Current code, unless I am completely mistaken, tests that you have at least one of the configured scopes.
Which anyway imho makes more sense.

``` AuthorizationMiddleware.cs
            var matchesScopes = routeAllowedScopes.Intersect(userScopes).ToList();

            if (matchesScopes.Count == 0)
            {
                return new ErrorResponse<bool>(
                    new ScopeNotAuthorizedError($"no one user scope: '{string.Join(",", userScopes)}' match with some allowed scope: '{string.Join(",", routeAllowedScopes)}'"));
            }
```

### DIFF
--- a/docs/features/authentication.rst
+++ b/docs/features/authentication.rst
@@ -172,6 +172,6 @@ NOTE: In order to get Ocelot to view the scope claim from Okta properly, you hav
 Allowed Scopes
 ^^^^^^^^^^^^^
 
-If you add scopes to AllowedScopes Ocelot will get all the user claims (from the token) of the type scope and make sure that the user has all of the scopes in the list.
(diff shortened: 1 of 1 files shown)


raman-m:
Hi Göran!

[ScopesAuthorizer.cs | Line 39](../blob/develop/src/Ocelot/Authorization/ScopesAuthorizer.cs#L39)
The code has changed a little bit! :)
But I hope the semantics remains the same.

raman-m (approved):
```diff
+ Approved
```

raman-m:
@roseeng
Thank you, Göran Roseen, for this text improvement!

PR #1555, opened 2022-01-18
Feat updatepackges 5. fluent validation is not backward compatibility v9 and v10

EngRajabi:
Feat update microsoft packages net 5.
for microsfot hotfix
for fluent validation is not backward compatibility v9 vs v10


raman-m:
@EngRajabi  Hi Mohsen!

Could you resolve all merge conflicts please?
Pay attention to the fact that during latest release 19.0.2 most of all NuGet packages were updated. Now Ocelot is available for targeting .NET 7. And I believe all packages are updated, including your required package.
Just try version [19.0.2](https://www.nuget.org/packages/Ocelot) and report if it helped plz!

PR #1582, opened 2022-06-15
.net 6 not support AuthenticationProviderKey.

ravanzli:
## Support AuthenticationProviderKey in .net 6!

I used .net 5 AuthenticationProviderKey without any problems. Then, .net 6 could not use that AuthenticationProviderKey.


raman-m:
@TomPallister 
I believe this PR should be converted to an issue! After that we can close this PR because it is kindly outdated, and the [feat/more-config](https://github.com/ThreeMammals/Ocelot/commits/feat/more-config) branch as target is wrong. Do you agree?

raman-m:
@ravanzli Hello Ravan!
I know you had reported in 2020, 3 years ago.
Let us know if your `AuthenticationProviderKey` issue still persists.
So, Ocelot was released with [19.0.2](https://www.nuget.org/packages/Ocelot/19.0.2) version 2 days ago. Could you try to upgrade to this version and check once again please?

Finally, could you share more details on the issue please?
What exact package version and target .NET version?
Can you write a unit/integration test(s) or show real case which will indicate to the problem plz?

We have some acceptance tests concerning authentication: [AuthenticationTests.cs](https://github.com/ThreeMammals/Ocelot/blob/develop/test/Ocelot.AcceptanceTests/AuthenticationTests.cs). Could you check these tests cover your case plz? If Not you need to write new test(s).

raman-m:
@ravanzli In future, if you've decided to indicate the problem, you have to create [an issue](../issues) with a description.
If you have some fixing source code and/or some enhancements I will be happy to help you to create valid PR to the [develop](../commits/develop) branch.

PR #1628, opened 2022-12-24
Incorrect port number

onurkanbakirci:
In websocket config docs, port number have to be `5001` as described in text explanation instead of `50000`.

### DIFF
--- a/docs/features/websockets.rst
+++ b/docs/features/websockets.rst
@@ -66,7 +66,7 @@ Then in your ocelot.json add the following to proxy a Route using SignalR. Note 
       "DownstreamHostAndPorts": [
         {
           "Host": "localhost",
-          "Port": 50000
+          "Port": 5000
         }
       ],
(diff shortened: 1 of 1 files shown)


raman-m (approved):

raman-m:
> In websocket config docs, port number have to be `5001` as described in text explanation instead of `50000`.

Hi Onurkan!
Yes, you are right, ports do not match now in [SignalR](https://ocelot.readthedocs.io/en/latest/features/websockets.html#signalr) paragraph.
But it is still not clear: should it be 5001 or 50000 actually?
Let's ask Tom, and he will decide.

onurkanbakirci:
Hi @raman-m . 
It should be `5001`

raman-m:
Duplicate of #1678 

@onurkanbakirci 
Thank you, Onurkan Bakırcı!

raman-m:
Duplicate of #1678

PR #1367, opened 2020-11-10
#893 Problems to cancel a request (CancellationToken)

jlukawska:
## Fixes #893 
- #893 

## Proposed Changes

  - One more problem related to `HttpContext`: cancellation token was not copied to the new `HttpContext`.

PR #1688, opened 2023-08-29
Copy RequestAborted

honestegg:
Copy RequestAborted when creating a new HttpContext in MultiplexingMiddleware

Fixes / New Feature #

## Proposed Changes

  - Fixes issue where request is not properly cancelled if cancelled by the client


dvandenhoek:
copy of #1367

raman-m:
Duplicate of #1367

Issue #1683, opened 2023-07-28
WebSocket exception


zhaoyongjie183:
Ocelot version: **19.0.2**

Ocelot configuration, a route：
```json
{
  "UpstreamPathTemplate": "/JavaMessageService/websocket/{url}",
  "UpstreamHttpMethod": [ "Get", "Post", "PUT", "DELETE", "OPTIONS" ],
  "DownstreamPathTemplate": "/api/websocket/{url}",
  "DownstreamScheme": "ws",
  "DownstreamHostAndPorts": [{
    "Host": "192.168.1.22",
    "Port": 9082
  }]
}
```

Direct Access to `ws://192.168.1.22:9082/api/websocket/xxl` was successful. 👇 
![image](https://github.com/ThreeMammals/Ocelot/assets/28800478/ec775226-606c-463a-8819-a9af83602271)

---

Ocelot UpstreamPathTemplate will report an error 👇 
![image](https://github.com/ThreeMammals/Ocelot/assets/28800478/a3c16774-51f3-4430-9e22-0539a8f5a9c4)

---

The error message of Ocelot pipeline is as follows:
```
System.ArgumentException: Only Uris starting with 'ws://' or 'wss://' are supported. (Parameter 'uri')
   at System.Net.WebSockets.ClientWebSocket.ConnectAsync(Uri uri, CancellationToken cancellationToken)
   at Ocelot.WebSockets.Middleware.WebSocketsProxyMiddleware.Proxy(HttpContext context, String serverEndpoint)
   at Ocelot.WebSockets.Middleware.WebSocketsProxyMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.DownstreamUrlCreator.Middleware.DownstreamUrlCreatorMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.LoadBalancer.Middleware.LoadBalancingMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.Request.Middleware.DownstreamRequestInitialiserMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.Multiplexer.MultiplexingMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.DownstreamRouteFinder.Middleware.DownstreamRouteFinderMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.Builder.Extensions.MapWhenMiddleware.Invoke(HttpContext context)
[...]

PR #1689, opened 2023-09-01
#1509 #1683 Replace non-WS protocols for the 'ClientWebSocket' in WebSocketsProxyMiddleware

### LINKED ISSUE #1509: Upstream Scheme is not applied to DownstreamScheme, if latter is not specified
If `DownstreamScheme` is not specified in a Route in configuration, the upstream scheme should be used to forward request to the downstream for the URLs other than http/https.

Basically, I want to support long polling as well as websockets on a SignalR hub/client in my application.
In both cases, I have to hit the same URL, except that my up and down stream schemes change.
I cannot create two entries with same url but different `DownstreamScheme` s at the same time. [I hope my understanding is right here. If not, please enlighten me. :) ]

When I follow the steps below, I get exception:

```
Exception caught in global error handler, exception message: Only Uris starting with 'ws://' or 'wss://' are supported. (Parameter 'uri'), exception stack:
   at System.Net.WebSockets.ClientWebSocket.ConnectAsync(Uri uri, CancellationToken cancellationToken)
   at Ocelot.WebSockets.Middleware.WebSocketsProxyMiddleware.Proxy(HttpContext context, String serverEndpoint)
   at Ocelot.WebSockets.Middleware.WebSocketsProxyMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.DownstreamUrlCreator.Middleware.DownstreamUrlCreatorMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.LoadBalancer.Middleware.LoadBalancingMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.Request.Middleware.DownstreamRequestInitialiserMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.Multiplexer.MultiplexingMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.DownstreamRouteFinder.Middleware.DownstreamRouteFinderMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.Builder.Extensions.MapWhenMiddleware.Invoke(HttpContext context)
   at Ocelot.Errors.Middleware.ExceptionHandlerMiddleware.Invoke(HttpContext httpContext) RequestId: 6fa0fd88-82ec-4ae4-95ed-29f1b1aefd14 {"SourceContext":"Ocelot.Errors.Middleware.ExceptionHandlerMiddleware","RequestId":"80000026-0006-ff00-b63f-84710c7967bb","RequestPath":"/tieto/collection/commsvc/connectcommunicationhub","CorrelationId":"6fa0fd88-82ec-4ae4-95ed-29f1b1aefd14"}
System.ArgumentException: Only Uris starting with 'ws://' or 'wss://' are supported. (Parameter 'uri')
   at System.Net.WebSockets.ClientWebSocket.ConnectAsync(Uri uri, CancellationToken cancellationToken)
   at Ocelot.WebSockets.Middleware.WebSocketsProxyMiddleware.Proxy(HttpContext context, String serverEndpoint)
   at Ocelot.WebSockets.Middleware.WebSocketsProxyMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.DownstreamUrlCreator.Middleware.DownstreamUrlCreatorMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.LoadBalancer.Middleware.LoadBalancingMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.Request.Middleware.DownstreamRequestInitialiserMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.Multiplexer.MultiplexingMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.DownstreamRouteFinder.Middleware.DownstreamRouteFinderMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.Builder.Extensions.MapWhenMiddleware.Invoke(HttpContext context)
   at Ocelot.Errors.Middleware.ExceptionHandlerMiddleware.Invoke(HttpContext httpContext)
```

1. Create a typical simple SignalR server/client application
2. Do not specify the `DownstreamScheme` in the `/myHub?id=<id>` url route configuration
3. Set the `HttpTransportType` to `LongPolling` on both client and server
4. Observe that it works
5. Set the `HttpTransportType` to `WebSockets` on both client and server
6. Observe that now it does not work, but gives above error

In each case, following URLs should be used on downstream without changing any configuration:

LongPolling: `https://localhost:1234/myHub?id=<id>`
WebSockets: `wss://localhost:1234/myHub?id=<id>`   [Logs show that Ocelot is still setting downstream to https]

If Ocelot behaves in the above (mentioned in Expected Behavior) manner, then my code will work without hassles.

ArtRoman:
## Fixes #1509 #1683 
- #1509 
- #1683 

Fix WebSocket for SignalR

## Proposed Changes
An ugly fix of `ArgumentException: Only Uris starting with 'ws://' or 'wss://' are supported. (Parameter 'uri')` exception from downstream host because 'uri' values starts from http, but this isn't allowed in `ClientWebSocket.ConnectAsync`.
This will return `Expected HTTP 101 response but was '500 Internal Server Error'` error to upstream host.

### DIFF
--- a/src/Ocelot/WebSockets/WebSocketsProxyMiddleware.cs
+++ b/src/Ocelot/WebSockets/WebSocketsProxyMiddleware.cs
@@ -128,7 +128,7 @@ namespace Ocelot.WebSockets
             // Only Uris starting with 'ws://' or 'wss://' are supported in System.Net.WebSockets.ClientWebSocket
             if (!serverEndpoint.StartsWith("ws"))
             {
-                Logger.LogError("Invalid scheme has detected!");
+                Logger.LogWarning("Invalid scheme has detected!");
                 serverEndpoint = serverEndpoint
                     .Replace("http://", "ws://")
                     .Replace("https://", "wss://");
--- a/src/Ocelot/WebSockets/WebSocketsProxyMiddleware.cs
+++ b/src/Ocelot/WebSockets/WebSocketsProxyMiddleware.cs
@@ -74,8 +74,11 @@ namespace Ocelot.WebSockets
         public async Task Invoke(HttpContext httpContext)
         {
             var uri = httpContext.Items.DownstreamRequest().ToUri();
+            uri = uri.Replace("https://", "wss://");
+            uri = uri.Replace("http://", "ws://");
             var downstreamRoute = httpContext.Items.DownstreamRoute();
             await Proxy(httpContext, uri, downstreamRoute);
+            await Proxy(httpContext, uri);
         }
 
         private async Task Proxy(HttpContext context, string serverEndpoint, DownstreamRoute downstreamRoute)
--- a/src/Ocelot/WebSockets/WebSocketsProxyMiddleware.cs
+++ b/src/Ocelot/WebSockets/WebSocketsProxyMiddleware.cs
@@ -126,10 +126,9 @@ namespace Ocelot.WebSockets
             }
 
             // Only Uris starting with 'ws://' or 'wss://' are supported in System.Net.WebSockets.ClientWebSocket
-            var wsServerEndpoint = serverEndpoint.Replace("https://", "wss://");
-            wsServerEndpoint = wsServerEndpoint.Replace("http://", "ws://");
-            
+            var wsServerEndpoint = serverEndpoint.Replace("https://", "wss://").Replace("http://", "ws://");
             var destinationUri = new Uri(wsServerEndpoint);
+
             await client.ConnectAsync(destinationUri, context.RequestAborted);
 
             using (var server = await context.WebSockets.AcceptWebSocketAsync(client.SubProtocol))
--- a/src/Ocelot/WebSockets/WebSocketsProxyMiddleware.cs
+++ b/src/Ocelot/WebSockets/WebSocketsProxyMiddleware.cs
@@ -18,9 +18,13 @@ namespace Ocelot.WebSockets
             "Connection", "Host", "Upgrade",
             "Sec-WebSocket-Accept", "Sec-WebSocket-Protocol", "Sec-WebSocket-Key", "Sec-WebSocket-Version", "Sec-WebSocket-Extensions",
         };
+
         private const int DefaultWebSocketBufferSize = 4096;
         private readonly RequestDelegate _next;
         private readonly IWebSocketsFactory _factory;
+
+        public const string IgnoredSslWarningFormat = $"You have ignored all SSL warnings by using {nameof(DownstreamRoute.DangerousAcceptAnyServerCertificateValidator)} for this downstream route! {nameof(DownstreamRoute.UpstreamPathTemplate)}: '{{0}}', {nameof(DownstreamRoute.DownstreamPathTemplate)}: '{{1}}'.";
+        public const string InvalidSchemeWarningFormat = "Invalid scheme has detected which will be replaced! Scheme '{0}' of the downstream '{1}'.";
 
         public WebSocketsProxyMiddleware(IOcelotLoggerFactory loggerFactory,
             RequestDelegate next,
@@ -106,7 +110,7 @@ namespace Ocelot.WebSockets
             if (route.DangerousAcceptAnyServerCertificateValidator)
             {
                 client.Options.RemoteCertificateValidationCallback = (request, certificate, chain, errors) => true;
-                Logger.LogWarning($"You have ignored all SSL warnings by using {nameof(DownstreamRoute.DangerousAcceptAnyServerCertificateValidator)} for this downstream route! {nameof(DownstreamRoute.UpstreamPathTemplate)}: '{route.UpstreamPathTemplate}', {nameof(DownstreamRoute.DownstreamPathTemplate)}: '{route.DownstreamPathTemplate}'.");
+                Logger.LogWarning(string.Format(IgnoredSslWarningFormat, route.UpstreamPathTemplate, route.DownstreamPathTemplate));
             }
 
             foreach (var protocol in context.WebSockets.WebSocketRequestedProtocols)
@@ -135,7 +139,7 @@ namespace Ocelot.WebSockets
             var scheme = request.Scheme;
             if (!scheme.StartsWith(Uri.UriSchemeWs))
             {
-                Logger.LogWarning($"Invalid scheme has detected which will be replaced! Scheme '{scheme}' of the downstream '{request.ToUri()}'.");
+                Logger.LogWarning(string.Format(InvalidSchemeWarningFormat, scheme, request.ToUri()));
                 request.Scheme = scheme == Uri.UriSchemeHttp ? Uri.UriSchemeWs
                     : scheme == Uri.UriSchemeHttps ? Uri.UriSchemeWss : scheme;
             }
--- a/src/Ocelot/WebSockets/WebSocketsProxyMiddleware.cs
+++ b/src/Ocelot/WebSockets/WebSocketsProxyMiddleware.cs
@@ -128,13 +128,14 @@ namespace Ocelot.WebSockets
             // Only Uris starting with 'ws://' or 'wss://' are supported in System.Net.WebSockets.ClientWebSocket
             if (!serverEndpoint.StartsWith("ws"))
             {
-                Logger.LogWarning("Invalid scheme has detected!");
+                var scheme = context.Items.DownstreamRequest().Scheme;
+                Logger.LogWarning($"Invalid scheme has detected which will be replaced! Scheme '{scheme}' of the downstream '{serverEndpoint}'.");
                 serverEndpoint = serverEndpoint
                     .Replace("http://", "ws://")
                     .Replace("https://", "wss://");
             }
-            var destinationUri = new Uri(serverEndpoint);
 
+            var destinationUri = new Uri(serverEndpoint);
             await client.ConnectAsync(destinationUri, context.RequestAborted);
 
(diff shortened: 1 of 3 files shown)


ArtRoman:
Any progress? 1 test fails but it's not connected with my changes:

> Application startup exception: System.AggregateException: One or more errors occurred. (Unable to start Ocelot, errors are: Unable to start Ocelot because either a Route or GlobalConfiguration are using QoSOptions but no QosDelegatingHandlerDelegate has been registered in dependency injection container. Are you missing a package like Ocelot.Provider.Polly and services.AddPolly()?)

raman-m:
Sorry, but the last build has failed in [one test only](https://app.circleci.com/pipelines/github/ThreeMammals/Ocelot/839/workflows/cb16adb8-9f1f-41df-b5b3-ae3463a3c87f/jobs/1556?invite=true#step-102-318415_123) which is [Ocelot.AcceptanceTests.ConfigurationReloadTests.should_reload_config_on_change](https://app.circleci.com/pipelines/github/ThreeMammals/Ocelot/839/workflows/cb16adb8-9f1f-41df-b5b3-ae3463a3c87f/jobs/1556?invite=true#step-102-291135_40)
This is unstable test. It fails in 1 of 2 cases or even more.

Which test did you mean in previous message? I have no idea.

raman-m:
@abhiphirke @zhaoyongjie183
Hello!
As an issue raisers,
Could you test this fix please, and confirm that your issue is gone with this fix?

src/Ocelot/WebSockets/Middleware/WebSocketsProxyMiddleware.cs
```diff
@@ -117,7 +117,10 @@ private static async Task Proxy(HttpContext context, string serverEndpoint)
                 }
             }
 
-            var destinationUri = new Uri(serverEndpoint);
+            // Only Uris starting with 'ws://' or 'wss://' are supported in System.Net.WebSockets.ClientWebSocket
+            var wsServerEndpoint = serverEndpoint.Replace("https://", "wss://").Replace("http://", "ws://");
+            var destinationUri = new Uri(wsServerEndpoint);
```
raman-m, line 122: I would suggest to test the URL string first to detect wrong scheme & protocol:
```suggestion
            if (!serverEndpoint.StartsWith("ws"))
            {
                Logger.LogError("Invalid scheme has detected!");
                serverEndpoint = serverEndpoint
                    .Replace("http://", "ws://")
                    .Replace("https://", "wss://");
            }
            var destinationUri = new Uri(serverEndpoint);
```
We could wrap the `client.ConnectAsync()` by **try-catch-finally** construction to process the exception:
`System.ArgumentException: Only Uris starting with 'ws://' or 'wss://' are supported. (Parameter 'uri')`
But simple **if**-testing is better.

raman-m (changes requested):
Probably we need to update configuration validators also to stop app forcibly if such WebSocket-routes with wrong scheme exist.

ArtRoman:
> Probably we need to update configuration validators also to stop app forcibly if such WebSocket-routes with wrong scheme exist.

It's not the route scheme issue. Websocket connection is usual HTTP connection with Upgrade header, it can be done with both "http://" or "ws://" endpoints in Url. For example, android okhttp library makes "http://" endpoint connection for websocket, even when "ws://" was in Uri, and that's fatal for Net.ClientWebSocket. So replacing protocol for real websocket connection resolves the issue.

src/Ocelot/WebSockets/Middleware/WebSocketsProxyMiddleware.cs
ArtRoman: This check looks goods, but Logger is not available in `private static async Task Proxy`, and LogError method requires message + exception parameters, maybe LogWarning will suit better?

RaynaldM: if protocol is invalid, you should raise an exception

RaynaldM (approved):

raman-m:
@ArtRoman 
Could you [Sync fork](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork) please?
We have the diff since yesterday: [Comparing ArtRoman:develop...ThreeMammals:develop](/ArtRoman/Ocelot/compare/develop...ThreeMammals:Ocelot:develop)

raman-m (approved):
Not elegant but it works! 😄 
Looks good! ✅ 

What about unit test for `Invoke` and `Proxy` methods?
Unit test should check exactly this user case and new added code.

ArtRoman:
> What about unit test for `Invoke` and `Proxy` methods?

Ok, I will try to add tests.

src/Ocelot/WebSockets/Middleware/WebSocketsProxyMiddleware.cs
```diff
@@ -117,6 +117,16 @@ private static async Task Proxy(HttpContext context, string serverEndpoint)
                 }
             }
 
+            // Only Uris starting with 'ws://' or 'wss://' are supported in System.Net.WebSockets.ClientWebSocket
+            if (!serverEndpoint.StartsWith("ws"))
+            {
+                var scheme = context.Items.DownstreamRequest().Scheme;
+                Logger.LogWarning($"Invalid scheme has detected which will be replaced! Scheme '{scheme}' of the downstream '{serverEndpoint}'.");
+                serverEndpoint = serverEndpoint
```
RaynaldM, line 125: question: is it really the role of this method to transform the scheme? 
I think that if the configuration isn't right, an exception should be raised.
There's nothing more confusing for a dev than to see in a config file a scheme defined as "http" and then find that it's executed as "wss", a magic trick!

RaynaldM (approved):

src/Ocelot/WebSockets/Middleware/WebSocketsProxyMiddleware.cs
raman-m: > @ArtRoman commented [on Sep 15](#issuecomment-1721577694)

Maybe because of that ☝️ WS protocol is based on http, and developer confuses too often.
@ArtRoman Explain please once again.

raman-m:
> @ArtRomancommented [on Sep 19](#issuecomment-1726544735)
Ok, I will try to add tests.

Perfect! After addition of unit tests this PR will be accepted.
Let me know if you need a help.

src/Ocelot/WebSockets/Middleware/WebSocketsProxyMiddleware.cs
ArtRoman: @RaynaldM If execution goes to WebSocketsProxyMiddleware class, that means user wants to proxy WebSocket connection, and configuration is OK, we shouldn't do anything here. 
But you can't open websocket directly, you should make a HTTP❗️ call with "Upgrade" header to webserver that handles websocket connections, then server responds with status "101 Switching Protocols" and then it [becomes a websocket](https://en.wikipedia.org/wiki/WebSocket#Protocol_handshake). 

"ws://" and "http://" in connection's URL are equal in this case (also "wss://" and "https://" accordingly), but passing user's "http" as protocol to [System.Net.WebSockets.ClientWebSocket.ConnectAsync()](https://learn.microsoft.com/en-us/dotnet/api/system.net.websockets.clientwebsocket.connectasync?view=net-7.0) will generate an exception "Only Uris starting with 'ws://' or 'wss://' are supported". So, we need to replace protocol in Url string before opening low-level WebSocket to handle proxy request. Replacing protocol [is common for constructing URL](https://stackoverflow.com/questions/10406930/how-to-construct-a-websocket-uri-relative-to-the-page-uri) from a given one, there is no magic here.

raman-m:
> @ArtRoman commented [on Sep 19](#issuecomment-1726544735)
Ok, I will try to add tests.

**Done!** See commit https://github.com/ThreeMammals/Ocelot/pull/1689/commits/5d9c214c67d51dbe9a1f598b63c50a76f2ef8cb0

raman-m:
@RaynaldM  Put your eye on this plz!
One more approval please! 😉

RaynaldM (approved):

raman-m (approved):

raman-m:
@ArtRoman Congrats! 🥳

raman-m:
>@JHennerley commented [on Sep 29](#issuecomment-1740862054)

Hi J!
I'm planning to release this Sunday, October 1. If something goes wrong, the late date is Monday, October 2.

---
> We're having some trouble handling websocket requests through Ocelot, it has to go through a few layers of **netscaler** -> **nginx** -> **ocelot**, but even when we try **nginx** -> **ocelot** we receive the above error.

Bad! What Ocelot version do you use currently? Based on .NET 7 ?
Inbound Ocelot traffic cannot be managed by the C# code, it is question of Ocelot app hosting. Do you see?
By the code and configs we can control & manage downstream connections and behavior. And a little bit, upstream connections. But entire upstream environment is Ocelot ASP.NET app hosting environment. 
Anyway, after release just try once again, and let me know testing results please.

ArtRoman:
> **Done!** See commit [5d9c214](https://github.com/ThreeMammals/Ocelot/commit/5d9c214c67d51dbe9a1f598b63c50a76f2ef8cb0)
> It was not easy to write them because there was no `WebSocketsProxyMiddlewareTests` class for unit tests

Yes, I'm stuck here. Thank you!

Issue #1700, opened 2023-09-23
Unstable test "should reload config on change"


raman-m:
## Expected Behavior
The test `should_reload_config_on_change` doesn't fail in 100% of running

## Actual Behavior / Motivation for New Feature
Currently the test [should_reload_config_on_change](/search?q=repo%3AThreeMammals%2FOcelot%20should_reload_config_on_change&type=code) fails sometimes: in 1/3 or even 1/2 cases.
Possible reason: multithreading for file access when test class tests being run in parallel.

## Steps to Reproduce the Problem
**develop** builds:
- Build [1666](https://circleci.com/gh/ThreeMammals/Ocelot/1666): [Scenario: should reload config on change](https://app.circleci.com/pipelines/github/ThreeMammals/Ocelot/895/workflows/02f04d5f-84ae-40a6-b9cb-bee3835bb65e/jobs/1666?invite=true#step-102-246180_40)
- Build [1592](https://circleci.com/gh/ThreeMammals/Ocelot/1592): [Scenario: should reload config on change](https://app.circleci.com/pipelines/github/ThreeMammals/Ocelot/857/workflows/cf2a0f20-4d12-4515-a251-c00a418f5259/jobs/1592?invite=true#step-102-261494_40)

PR builds:
- Build [1710](https://circleci.com/gh/ThreeMammals/Ocelot/1710): [Scenario: should reload config on change](https://app.circleci.com/pipelines/github/ThreeMammals/Ocelot/917/workflows/a8e23c40-a37a-40e4-b7c8-a071e6ce63f8/jobs/1710?invite=true#step-102-264073_40)
- [etc.](https://app.circleci.com/pipelines/github/ThreeMammals/Ocelot?status=failed)

## Specifications
  - Version: latest
  - Platform: Docker
  - Subsystem: CircleCI

raman-m:
@ggnaegi Hey Guillaume!
Could you look into this problem, please?

ggnaegi:
@raman-m Ok, we have a problem here. How could I possibly reproduce the same conditions as the CI Pipeline?

ggnaegi:
@raman-m What we could do, is provide a factory for the configuration files and change the configuration files id depending on the facts

raman-m:
@ggnaegi 
I see 2-3 possible solutions:

1. Make test class tests running sequentially via good known xUnit attribute. Most easy and fast fix delivery. But it seems you don't like this solution.

[...]

PR #1702, opened 2023-09-24
Adding configs factory to avoid side effects of changing the same con…

ggnaegi:
…fig.

Fixes / New Feature #1700 

## Proposed Changes
  - Adding configs factory to avoid side effects of changing the same config.
xUnit assumes that tests are isolated and side-effect-free, meaning they don't interfere with one another


ggnaegi:
@raman-m that would be my proposal, no need to mess with the parallelism.

PR #1703, opened 2023-09-24
#1700 Create custom Ocelot config file when instantiating steps during parallel execution

### LINKED ISSUE #1700: Unstable test "should reload config on change"

ggnaegi:
# Fixes #1700 
- #1700 

When instantiating steps, creating a custom ocelot config file name. Finally deleting the newly created config file when disposing the `Step` instance.

## Proposed Changes
  - Since, by design, a new instance of `Step` is instantiated for each fact/theory, we need to ensure that the corresponding fact/theory only uses its own ocelot configuration file, without interfering with the configuration files of other facts and theories.

### DIFF
--- a/test/Ocelot.AcceptanceTests/ConfigurationReloadTests.cs
+++ b/test/Ocelot.AcceptanceTests/ConfigurationReloadTests.cs
@@ -87,7 +87,6 @@ namespace Ocelot.AcceptanceTests
         public void Dispose()
         {
             _steps.Dispose();
-            GC.SuppressFinalize(this);
         }
     }
 }
--- a/test/Ocelot.AcceptanceTests/ConfigurationReloadTests.cs
+++ b/test/Ocelot.AcceptanceTests/ConfigurationReloadTests.cs
@@ -42,8 +42,7 @@ namespace Ocelot.AcceptanceTests
             this.Given(x => _steps.GivenThereIsAConfiguration(_initialConfig))
                 .And(x => _steps.GivenOcelotIsRunningReloadingConfig(true))
                 .And(x => _steps.GivenThereIsAConfiguration(_anotherConfig))
-                .And(x => _steps.GivenIWait(7500))
-                .And(x => _steps.ThenConfigShouldBe(_anotherConfig))
+                .And(x => _steps.ThenConfigShouldBeWithTimeout(_anotherConfig, 10000))
                 .BDDfy();
         }
 
--- a/test/Ocelot.AcceptanceTests/ConfigurationReloadTests.cs
+++ b/test/Ocelot.AcceptanceTests/ConfigurationReloadTests.cs
@@ -9,7 +9,7 @@ using Xunit;
 
 namespace Ocelot.AcceptanceTests
 {
-    public class ConfigurationReloadTests : IDisposable
+    public sealed class ConfigurationReloadTests : IDisposable
     {
         private readonly FileConfiguration _initialConfig;
         private readonly FileConfiguration _anotherConfig;
--- a/test/Ocelot.AcceptanceTests/Steps.cs
+++ b/test/Ocelot.AcceptanceTests/Steps.cs
@@ -54,13 +54,19 @@ namespace Ocelot.AcceptanceTests
         private HttpContent _postContent;
         private BearerToken _token;
         public string RequestIdKey = "OcRequestId";
-        private readonly Random _random = new();
-        private readonly string _ocelotConfigFileName = $"{Guid.NewGuid():N}-ocelot.json";
+        private readonly Random _random;
+        private readonly string _ocelotConfigFileName;
         private IWebHostBuilder _webHostBuilder;
         private WebHostBuilder _ocelotBuilder;
         private IWebHost _ocelotHost;
         private IOcelotConfigurationChangeTokenSource _changeToken;
         private bool _disposedValue;
+
+        public Steps()
+        {
+            _random = new();
+            _ocelotConfigFileName = $"{Guid.NewGuid():N}-ocelot.json";
+        }
 
         public async Task ThenConfigShouldBe(FileConfiguration fileConfig)
         {
--- a/test/Ocelot.AcceptanceTests/Steps.cs
+++ b/test/Ocelot.AcceptanceTests/Steps.cs
@@ -65,13 +65,35 @@ namespace Ocelot.AcceptanceTests
         public async Task ThenConfigShouldBe(FileConfiguration fileConfig)
         {
             var internalConfigCreator = _ocelotServer.Host.Services.GetService<IInternalConfigurationCreator>();
-
             var internalConfigRepo = _ocelotServer.Host.Services.GetService<IInternalConfigurationRepository>();
-            var internalConfig = internalConfigRepo.Get();
 
+            var internalConfig = internalConfigRepo.Get();
             var config = await internalConfigCreator.Create(fileConfig);
 
             internalConfig.Data.RequestId.ShouldBe(config.Data.RequestId);
+        }
+
+        public async Task ThenConfigShouldBeWithTimeout(FileConfiguration fileConfig, int timeoutMs)
+        {
+            var totalMs = 0;
+            var result = false;
+
+            while (!result && totalMs < timeoutMs)
+            {
+                result = await Wait.WaitFor(1000).Until(async () =>
+                {
+                    var internalConfigCreator = _ocelotServer.Host.Services.GetService<IInternalConfigurationCreator>();
+                    var internalConfigRepo = _ocelotServer.Host.Services.GetService<IInternalConfigurationRepository>();
+
+                    var internalConfig = internalConfigRepo.Get();
+                    var config = await internalConfigCreator.Create(fileConfig);
+
+                    return internalConfig.Data.RequestId == config.Data.RequestId;
+                });
+                totalMs += 1000;
+            }
+
+            result.ShouldBe(true);
         }
 
         public async Task StartFakeOcelotWithWebSockets()
--- a/test/Ocelot.AcceptanceTests/Steps.cs
+++ b/test/Ocelot.AcceptanceTests/Steps.cs
@@ -147,7 +147,6 @@ namespace Ocelot.AcceptanceTests
         public void GivenThereIsAConfiguration(FileConfiguration fileConfiguration)
         {
             var jsonConfiguration = JsonConvert.SerializeObject(fileConfiguration, Formatting.Indented);
-            DeleteOcelotConfig();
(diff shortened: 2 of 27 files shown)


raman-m:
Wow! Cool!
Let's make 10 PR builds of 10 commits being added to this PR, to get some statistics on stability of this fix.

Just commit a small change of Ocelot core, not tests, and the 2nd commit should be revert commit of previous one.
So, there will be 5 commits and 5 revert-commits.

Good plan?

ggnaegi:
@raman-m no way, that's really funny ;-)

ggnaegi:
@raman-m Ok, a bit better now, 9 out of 10...

ggnaegi:
@raman-m why did you remove xunit.runner.json?

raman-m:
I've reverted commit [9fac778](https://github.com/ThreeMammals/Ocelot/pull/1703/commits/9fac77895de19053092dc8cbe35e4bed34ab2dec)
Sorry!

It seems your solution doesn't work for some reason... That's strange... It should...

Have a rest today! 😉 
We will continue tomorrow, this week...

raman-m:
@ggnaegi [commented](#issuecomment-1732637953)
> why did you remove xunit.runner.json?

Because you've disabled parallel execution for all tests by default.

ggnaegi:
> @ggnaegi [commented](#issuecomment-1732637953)
> 
> > why did you remove xunit.runner.json?
> 
> Because you've disabled parallel execution for all tests by default.

Yeah, I tried...

ggnaegi:
@raman-m 11 out of 12. It's not that bad...

ggnaegi:
@raman-m we might have a solution...

raman-m:
It seems this fix works! 👍 
What was the root cause of failing? I don't think it was random port finder... 🤣

raman-m:
Do you confirm that we can still use parallelization of all tests?

ggnaegi:
> Do you confirm that we can still use parallelization of all tests?

@raman-m Yes, we still have an issue with Open_circuit_should_not_effect_different_route though... That's the test failing at the minute. I can reproduce that on my computer, it fails after 20 minutes.

raman-m:
> Open_circuit_should_not_effect_different_route 

Forget about this test now. It is out of the scope of linked bug-issue.
🆗 I'm going to review...

ggnaegi:
> > Open_circuit_should_not_effect_different_route
> 
> Forget about this test now. It is out of the scope of linked bug-issue. 🆗 I'm going to revi

> It seems this fix works! 👍 What was the root cause of failing? I don't think it was random port finder... 🤣

No it wasn't, but the random port finder code doesn't seem to be thread safe, since the concurrentbag would allow identical values to be stored and the method can be accessed by several threads concurrently

RaynaldM (dismissed):

test/Ocelot.AcceptanceTests/ConfigurationReloadTests.cs
```diff
@@ -88,7 +87,6 @@ public void should_not_trigger_change_token_with_no_change()
         public void Dispose()
         {
             _steps.Dispose();
-            GC.SuppressFinalize(this);
```
raman-m, line 91: Why did you remove the GC call? Because there is no descendants of the class, and there is no finalizer?
The `GC.SuppressFinalize` makes sense for subclasses and non-sealed class.
So, why not to mark the class as `sealed`? 😉

test/Ocelot.AcceptanceTests/Steps.cs
```diff
@@ -73,29 +54,41 @@ public class Steps : IDisposable
         private HttpContent _postContent;
         private BearerToken _token;
         public string RequestIdKey = "OcRequestId";
-        private readonly Random _random;
+        private readonly Random _random = new();
+        private readonly string _ocelotConfigFileName = $"{Guid.NewGuid():N}-ocelot.json";
```
raman-m, line 58: I can accept this: `private readonly Random _random = new();`
But I cannot accept on line 58. It contains business logic which is hidden in definition of the `_ocelotConfigFileName` member.
So, it is better to move all initializations to constructor.

test/Ocelot.AcceptanceTests/Steps.cs
```diff
@@ -73,29 +54,41 @@ public class Steps : IDisposable
...
+        private bool _disposedValue;
 
         public async Task ThenConfigShouldBe(FileConfiguration fileConfig)
         {
             var internalConfigCreator = _ocelotServer.Host.Services.GetService<IInternalConfigurationCreator>();
-
             var internalConfigRepo = _ocelotServer.Host.Services.GetService<IInternalConfigurationRepository>();
-            var internalConfig = internalConfigRepo.Get();
 
+            var internalConfig = internalConfigRepo.Get();
```
raman-m, line 70: Was it a root cause? I mean, absence of this call?

test/Ocelot.AcceptanceTests/Steps.cs
```diff
@@ -73,29 +54,41 @@ public class Steps : IDisposable
...
+        {
+            var result = await Wait.WaitFor(timeoutMs).Until(async () =>
+            {
+                var internalConfigCreator = _ocelotServer.Host.Services.GetService<IInternalConfigurationCreator>();
+                var internalConfigRepo = _ocelotServer.Host.Services.GetService<IInternalConfigurationRepository>();
+
+                var internalConfig = internalConfigRepo.Get();
+                var config = await internalConfigCreator.Create(fileConfig);
+
+                return internalConfig.Data.RequestId == config.Data.RequestId;
```
raman-m, line 86: Can we return `Tuple<T1, T2>` to perform `ShouldBe` checking?

test/Ocelot.AcceptanceTests/Steps.cs
```diff
@@ -73,29 +54,41 @@ public class Steps : IDisposable
...
         }
 
+        public async Task ThenConfigShouldBeWithTimeout(FileConfiguration fileConfig, int timeoutMs)
+        {
+            var result = await Wait.WaitFor(timeoutMs).Until(async () =>
+            {
+                var internalConfigCreator = _ocelotServer.Host.Services.GetService<IInternalConfigurationCreator>();
+                var internalConfigRepo = _ocelotServer.Host.Services.GetService<IInternalConfigurationRepository>();
+
+                var internalConfig = internalConfigRepo.Get();
```
raman-m, line 83: Can we extract it to upper context, because it is initial config, right? And initial config can be read outside of waiting task, right?

test/Ocelot.AcceptanceTests/Steps.cs
```diff
@@ -113,7 +106,7 @@ public async Task StartFakeOcelotWithWebSockets()
                     var env = hostingContext.HostingEnvironment;
                     config.AddJsonFile("appsettings.json", optional: true, reloadOnChange: false)
                         .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true, reloadOnChange: false);
-                    config.AddJsonFile("ocelot.json", false, false);
+                    config.AddJsonFile(_ocelotConfigFileName, false, false);
```
raman-m, line 109: Perfect! I love it! 😍 No hardcode anymore!

raman-m (changes requested):
Minor issues, but it is always good to improve the code.

test/Ocelot.AcceptanceTests/ConfigurationReloadTests.cs
ggnaegi: ok for sealed

test/Ocelot.AcceptanceTests/Steps.cs
ggnaegi: well...

test/Ocelot.AcceptanceTests/Steps.cs
ggnaegi: It was already in there.

test/Ocelot.AcceptanceTests/Steps.cs
ggnaegi: It needs a refactoring of Until. We could do it later

test/Ocelot.AcceptanceTests/Steps.cs
ggnaegi: Can we just keep it like that for the minute, We could create a new PR later, and implement a better design for Steps.

ggnaegi:
@raman-m imho ok to merge

raman-m:
> [Revert "commit #h5"](1703/commits/6e4b8a42c05207f73ed675db5b0c3468c20057cd) 
> [6e4b8a4](/1703/commits/6e4b8a42c05207f73ed675db5b0c3468c20057cd)

@ggnaegi #1706  FYI

test/Ocelot.AcceptanceTests/Steps.cs
raman-m: Please include this changes while working on #1706 if you wish

raman-m (dismissed):
```diff
+ Approved
```

raman-m (approved):

raman-m:
@ggnaegi Congrats!
We did that!
Thank you very mush for your efforts during this Sunday's evening!

PR #1705, opened 2023-09-25
#1700 Add sequential tests with disabled parallelization

raman-m:
- #1700 

## Proposed Changes
  - Define `SequentialTests` class with disabled parallelization


raman-m:
@ggnaegi  FYI
Let's add this too... It doesn't fix # 1700 actually, but it relates to it. And it will be good to have at least definition for future usage.

RaynaldM (approved):

Issue #1706, opened 2023-09-25
Unstable test "Open circuit should not effect different route"


raman-m:
## Expected Behavior / New Feature
The test `Ocelot.AcceptanceTests.PollyQoSTests.Open_circuit_should_not_effect_different_route` doesn't fail in 100% of running

## Actual Behavior / Motivation for New Feature
Currently the test [Open_circuit_should_not_effect_different_route](/search?q=repo%3AThreeMammals%2FOcelot%20Open_circuit_should_not_effect_different_route&type=code) fails sometimes, very rarely.
Possible reason: unknown.

## Steps to Reproduce the Problem
- PR builds: [1884](https://app.circleci.com/pipelines/github/ThreeMammals/Ocelot/1004/workflows/c97b3e46-4b2e-49a1-bc63-fdc8d0aebdc9/jobs/1884?invite=true#step-102-303655_56), [1732](https://app.circleci.com/pipelines/github/ThreeMammals/Ocelot/928/workflows/93b20a29-a8ed-4e24-862f-8dd383aad00f/jobs/1732?invite=true#step-102-309194_56), [1728](https://app.circleci.com/pipelines/github/ThreeMammals/Ocelot/926/workflows/e63bc07b-0abd-44d1-8037-da384cf6c509/jobs/1728?invite=true#step-102-256440_56)
- It is also reproducible in Visual Studio: [see screenshot](1700#issuecomment-1733220654)

## Specifications
  - Version: latest, 19.0.3
  - Platform: any
  - Subsystem: CircleCI/Docker, Visual Studio

ggnaegi:
@raman-m Well, it's very odd... At some point, we have a timing problem... It's very difficult to manage. First I thought, ok instead of using the local variable _requestCount, let's inject a singleton IRequestCountService like this:
``` c#
.ConfigureServices(services =>
                {
                    services.AddSingleton<IRequestCountService, RequestCountService>();
                })
```
and then refactor the method like that:
``` c#
private void GivenThereIsAPossiblyBrokenServiceRunningOn(string url, string responseBody)
        {
            _serviceHandler.GivenThereIsAServiceRunningOn(url, async context =>
            {
                var requestCountService = context.RequestServices.GetService<IRequestCountService>();
                context.Response.StatusCode = 200;

                if (requestCountService.GetRequestCount() == 1)
                {
                    await Task.Delay(1000);
                    await context.Response.WriteAsync("delayed service");
                    return;
                }

[...]

PR #1753, opened 2023-10-23
#1550 #1706 Addressing the QoS options ExceptionsAllowedBeforeBreaking issue

### LINKED ISSUE #1550: Circuit breaker not handling exception at end points even when ExceptionsAllowedBeforeBreaking is set.

As per the documentation of QoS functionality of Ocelot, it uses Polly to achieve Circuit Breaker among the other resilience features. From what I could understand, the following configuration should have triggered the circuit breaker after an endpoint raised exception more than 2 times.

```json
"QoSOptions": {
   "ExceptionsAllowedBeforeBreaking": 2,
   "DurationOfBreak": 10000,
   "TimeoutValue": 2000
}
```

However, it looked like the exception raised by the end points isn't handled by the circuit breaker. Please note that the Timeout works perfectly and triggers the circuit breaker as expected. However the case wasn't same with the _ExceptionsAllowedBeforeBreaking_.

  1. Create a route in the gateway with the above QoS configuration
  2. The specified Endpoint should raise an exception 
  3. Raise the request for end point via the gateway.

The code sample in the following link demonstrates the issue
[Demo Code](https://github.com/anuviswan/LearningPoint/tree/master/Microservices/AuthenticationDemo)

Similar problem is mentioned in the following Stackoverflow question too
- [Ocelot QoS doesn't break circuit when n number of exception arises](https://stackoverflow.com/questions/56851572/ocelot-qos-doesnt-break-circuit-when-n-number-of-exception-arises)

## Specifications
  - Ocelot: 17.0.1
  - Ocelot.Provider.Polly: 17.0.1

ggnaegi:
## Fixes #1550 #1706 
- #1550 
- #1706 

## Proposed Changes
- We shouldn't create a Circuit Breaker per route / request, only maintain one Circuit Breaker per route
- The Circuit Breaker is Expecting an Exception to kick in, a response with status 500 is not an Exception. If status 500, then throwing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -255,3 +255,4 @@ _templates/
 
 # Test Results
 *.trx
+/Ocelot.sln.DotSettings
--- a/Ocelot.sln.DotSettings
+++ b/Ocelot.sln.DotSettings
@@ -1,2 +1,0 @@
-﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
-	<s:Boolean x:Key="/Default/UserDictionary/Words/=appsettings/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
--- a/Ocelot.sln.DotSettings
+++ b/Ocelot.sln.DotSettings
@@ -1,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=appsettings/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
--- a/src/Ocelot.Provider.Polly/CircuitBreaker.cs
+++ b/src/Ocelot.Provider.Polly/CircuitBreaker.cs
@@ -7,7 +7,7 @@ public class CircuitBreaker<TResult>
     /// Initializes a new instance of the <see cref="CircuitBreaker{TResult}"/> class.
     /// We expect at least one policy to be passed in, default can't be null.
     /// </summary>
-    /// <param name="policies">the policies with at least a timeout policy</param>
+    /// <param name="policies">The policies with at least a <see cref="Policy.Timeout(int)"/> policy.</param>
     public CircuitBreaker(params IAsyncPolicy<TResult>[] policies)
     {
         var allPolicies = policies.Where(p => p != null).ToArray();
--- a/src/Ocelot.Provider.Polly/CircuitBreaker.cs
+++ b/src/Ocelot.Provider.Polly/CircuitBreaker.cs
@@ -1,12 +1,17 @@
-namespace Ocelot.Provider.Polly
-{
-    public class CircuitBreaker
-    {
-        public CircuitBreaker(params IAsyncPolicy[] policies)
-        {
-            Policies = policies.Where(p => p != null).ToArray();
-        }
+namespace Ocelot.Provider.Polly;
 
-        public IAsyncPolicy[] Policies { get; }
+public class CircuitBreaker<TResult>
+    where TResult : class
+{
+    public CircuitBreaker(params IAsyncPolicy<TResult>[] policies)
+    {
+        var allPolicies = policies.Where(p => p != null).ToArray();
+        CircuitBreakerAsyncPolicy = allPolicies.FirstOrDefault();
+
+        if (allPolicies.Length > 1)
+        {
+            CircuitBreakerAsyncPolicy = Policy.WrapAsync(allPolicies);
+        }
     }
+    public IAsyncPolicy<TResult> CircuitBreakerAsyncPolicy { get; }
 }
--- a/src/Ocelot.Provider.Polly/CircuitBreaker.cs
+++ b/src/Ocelot.Provider.Polly/CircuitBreaker.cs
@@ -3,15 +3,21 @@ namespace Ocelot.Provider.Polly;
 public class CircuitBreaker<TResult>
     where TResult : class
 {
+    /// <summary>
+    /// Initializes a new instance of the <see cref="CircuitBreaker{TResult}"/> class.
+    /// We expect at least one policy to be passed in, default can't be null.
+    /// </summary>
+    /// <param name="policies">the policies with at least a timeout policy</param>
     public CircuitBreaker(params IAsyncPolicy<TResult>[] policies)
     {
         var allPolicies = policies.Where(p => p != null).ToArray();
-        CircuitBreakerAsyncPolicy = allPolicies.FirstOrDefault();
+        CircuitBreakerAsyncPolicy = allPolicies.First();
 
         if (allPolicies.Length > 1)
         {
             CircuitBreakerAsyncPolicy = Policy.WrapAsync(allPolicies);
         }
     }
+
     public IAsyncPolicy<TResult> CircuitBreakerAsyncPolicy { get; }
 }
--- a/src/Ocelot.Provider.Polly/OcelotBuilderExtensions.cs
+++ b/src/Ocelot.Provider.Polly/OcelotBuilderExtensions.cs
@@ -1,8 +1,10 @@
-﻿using Microsoft.Extensions.DependencyInjection;
+﻿using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
 using Ocelot.Configuration;
 using Ocelot.DependencyInjection;
 using Ocelot.Errors;
 using Ocelot.Logging;
+using Ocelot.Provider.Polly.Interfaces;
 using Ocelot.Requester;
 using Polly.CircuitBreaker;
 using Polly.Timeout;
@@ -22,11 +24,12 @@ namespace Ocelot.Provider.Polly
 
             builder.Services
                 .AddSingleton(errorMapping)
+                .AddSingleton<IPollyQoSProvider, PollyQoSProvider>()
                 .AddSingleton<QosDelegatingHandlerDelegate>(GetDelegatingHandler);
             return builder;
         }
 
-        private static DelegatingHandler GetDelegatingHandler(DownstreamRoute route, IOcelotLoggerFactory logger)
-            => new PollyCircuitBreakingDelegatingHandler(new PollyQoSProvider(route, logger), logger);
+        private static DelegatingHandler GetDelegatingHandler(DownstreamRoute route, IHttpContextAccessor contextAccessor, IOcelotLoggerFactory loggerFactory)
+            => new PollyCircuitBreakingDelegatingHandler(route, contextAccessor, loggerFactory);
     }
 }
--- a/src/Ocelot.Provider.Polly/OcelotBuilderExtensions.cs
+++ b/src/Ocelot.Provider.Polly/OcelotBuilderExtensions.cs
@@ -31,5 +31,5 @@ public static class OcelotBuilderExtensions
     }
 
     private static DelegatingHandler GetDelegatingHandler(DownstreamRoute route, IHttpContextAccessor contextAccessor, IOcelotLoggerFactory loggerFactory) 
-        => new PollyCircuitBreakingDelegatingHandler(route, contextAccessor, loggerFactory);
+        => new PollyPoliciesDelegatingHandler(route, contextAccessor, loggerFactory);
 }
--- a/src/Ocelot.Provider.Polly/OcelotBuilderExtensions.cs
+++ b/src/Ocelot.Provider.Polly/OcelotBuilderExtensions.cs
@@ -30,9 +30,6 @@ public static class OcelotBuilderExtensions
         return builder;
     }
 
-    private static DelegatingHandler GetDelegatingHandler(DownstreamRoute route, IHttpContextAccessor contextAccessor,
-        IOcelotLoggerFactory loggerFactory)
-    {
-        return new PollyCircuitBreakingDelegatingHandler(route, contextAccessor, loggerFactory);
-    }
+    private static DelegatingHandler GetDelegatingHandler(DownstreamRoute route, IHttpContextAccessor contextAccessor, IOcelotLoggerFactory loggerFactory) 
+        => new PollyCircuitBreakingDelegatingHandler(route, contextAccessor, loggerFactory);
 }
--- a/src/Ocelot.Provider.Polly/OcelotBuilderExtensions.cs
+++ b/src/Ocelot.Provider.Polly/OcelotBuilderExtensions.cs
@@ -9,27 +9,30 @@ using Ocelot.Requester;
 using Polly.CircuitBreaker;
 using Polly.Timeout;
 
-namespace Ocelot.Provider.Polly
+namespace Ocelot.Provider.Polly;
+
+public static class OcelotBuilderExtensions
 {
-    public static class OcelotBuilderExtensions
+    public static IOcelotBuilder AddPolly(this IOcelotBuilder builder)
     {
-        public static IOcelotBuilder AddPolly(this IOcelotBuilder builder)
+        var errorMapping = new Dictionary<Type, Func<Exception, Error>>
         {
-            var errorMapping = new Dictionary<Type, Func<Exception, Error>>
-            {
-                {typeof(TaskCanceledException), e => new RequestTimedOutError(e)},
-                {typeof(TimeoutRejectedException), e => new RequestTimedOutError(e)},
-                {typeof(BrokenCircuitException), e => new RequestTimedOutError(e)},
-            };
+            { typeof(TaskCanceledException), e => new RequestTimedOutError(e) },
+            { typeof(TimeoutRejectedException), e => new RequestTimedOutError(e) },
+            { typeof(BrokenCircuitException), e => new RequestTimedOutError(e) },
+            { typeof(BrokenCircuitException<HttpResponseMessage>), e => new RequestTimedOutError(e) },
+        };
 
-            builder.Services
-                .AddSingleton(errorMapping)
-                .AddSingleton<IPollyQoSProvider, PollyQoSProvider>()
-                .AddSingleton<QosDelegatingHandlerDelegate>(GetDelegatingHandler);
-            return builder;
-        }
+        builder.Services
+            .AddSingleton(errorMapping)
+            .AddSingleton<IPollyQoSProvider<HttpResponseMessage>, PollyQoSProvider>()
+            .AddSingleton<QosDelegatingHandlerDelegate>(GetDelegatingHandler);
+        return builder;
+    }
 
-        private static DelegatingHandler GetDelegatingHandler(DownstreamRoute route, IHttpContextAccessor contextAccessor, IOcelotLoggerFactory loggerFactory)
-            => new PollyCircuitBreakingDelegatingHandler(route, contextAccessor, loggerFactory);
+    private static DelegatingHandler GetDelegatingHandler(DownstreamRoute route, IHttpContextAccessor contextAccessor,
+        IOcelotLoggerFactory loggerFactory)
+    {
+        return new PollyCircuitBreakingDelegatingHandler(route, contextAccessor, loggerFactory);
     }
 }
--- a/src/Ocelot.Provider.Polly/PollyCircuitBreakingDelegatingHandler.cs
+++ b/src/Ocelot.Provider.Polly/PollyCircuitBreakingDelegatingHandler.cs
@@ -1,48 +1,76 @@
+using System.Diagnostics;
+using System.Net;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Ocelot.Configuration;
 using Ocelot.Logging;
 using Ocelot.Provider.Polly.Interfaces;
 using Polly.CircuitBreaker;
 
-namespace Ocelot.Provider.Polly
+namespace Ocelot.Provider.Polly;
+
+public class PollyCircuitBreakingDelegatingHandler : DelegatingHandler
 {
-    public class PollyCircuitBreakingDelegatingHandler : DelegatingHandler
+    private readonly DownstreamRoute _route;
+    private readonly IHttpContextAccessor _contextAccessor;
+    private readonly IOcelotLogger _logger;
+
+    public PollyCircuitBreakingDelegatingHandler(
+        DownstreamRoute route,
+        IHttpContextAccessor contextAccessor, 
+        IOcelotLoggerFactory loggerFactory)
     {
-        private readonly IPollyQoSProvider _qoSProvider;
-        private readonly IOcelotLogger _logger;
+        _route = route;
+        _contextAccessor = contextAccessor;
+        _logger = loggerFactory.CreateLogger<PollyCircuitBreakingDelegatingHandler>();
+    }
 
-        public PollyCircuitBreakingDelegatingHandler(
-            IPollyQoSProvider qoSProvider,
-            IOcelotLoggerFactory loggerFactory)
-        {
-            _qoSProvider = qoSProvider;
-            _logger = loggerFactory.CreateLogger<PollyCircuitBreakingDelegatingHandler>();
-        }
+    private IPollyQoSProvider GetQoSProvider()
+    {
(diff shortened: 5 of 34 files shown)


ks1990cn (approved):

ks1990cn (approved):

ks1990cn (approved):

ks1990cn (approved):

ggnaegi:
@raman-m @RaynaldM This PR is ready for review. 

I have added several unit tests:
- Circuit breaker should build
- it contains only one policy if ExceptionsAllowedBeforeBreaking = 0
- it contains two policies of type ```AsyncCircuitBreakerPolicy<HttpResponseMessage>``` and ```AsyncTimeoutPolicy<HttpResponseMessage>``` if ExceptionsAllowedBeforeBreaking > 0
- should return same circuit breaker for given route
- should return different circuit breakers for two different routes
- should throw exception ```BrokenCircuitException<HttpResponseMessage>```` on third faulty attempt if ExceptionsAllowedBeforeBreaking = 2
- should no throw if status code is OK and not within 500...508 range
- should throw but after delay allow one new request
- should throw but atfer delay allow one faulty request and then throw again
- should throw but atfer delay allow one correct response and put counter back to zero
acceptance tests:
- should open circuit breaker after two exceptions

RaynaldM (approved):

raman-m:
Gui, would you like to include this fix in October release?

ggnaegi:
> Gui, would you like to include this fix in October release?

I think so yes.

raman-m:
Gui,
As you see, I've rebased feature branch... And I'll review tomorrow...
Happy presentation tomorrow! 😉

Ocelot.sln.DotSettings
```diff
@@ -0,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=appsettings/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
```
raman-m, line 2: New line before EOF is expected

src/Ocelot.Provider.Polly/CircuitBreaker.cs
```diff
@@ -1,12 +1,23 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="CircuitBreaker{TResult}"/> class.
+    /// We expect at least one policy to be passed in, default can't be null.
+    /// </summary>
+    /// <param name="policies">The policies with at least a <see cref="Policy.Timeout(int)"/> policy.</param>
+    public CircuitBreaker(params IAsyncPolicy<TResult>[] policies)
     {
-        public CircuitBreaker(params IAsyncPolicy[] policies)
+        var allPolicies = policies.Where(p => p != null).ToArray();
+        CircuitBreakerAsyncPolicy = allPolicies.First();
```
raman-m, line 14: Are you sure that 1st array item will be `CircuitBreakerAsyncPolicy` matching the type?
Maybe you need a predicate by type here...

src/Ocelot.Provider.Polly/OcelotBuilderExtensions.cs
```diff
@@ -1,32 +1,35 @@
...
+            { typeof(TaskCanceledException), e => new RequestTimedOutError(e) },
+            { typeof(TimeoutRejectedException), e => new RequestTimedOutError(e) },
+            { typeof(BrokenCircuitException), e => new RequestTimedOutError(e) },
+            { typeof(BrokenCircuitException<HttpResponseMessage>), e => new RequestTimedOutError(e) },
+        };
+
+        builder.Services
+            .AddSingleton(errorMapping)
+            .AddSingleton<IPollyQoSProvider<HttpResponseMessage>, PollyQoSProvider>()
+            .AddSingleton<QosDelegatingHandlerDelegate>(GetDelegatingHandler);
```
raman-m, line 29: Not an issue. But could we rename it removing suffix `Delegate` ?
According to Microsoft .NET conventions, [EventHandler](https://learn.microsoft.com/en-us/dotnet/api/system.eventhandler?view=net-7.0) or other `Handler` types are delegates.
So, there is no need to add this suffix, it is redundant.

src/Ocelot.Provider.Polly/PollyCircuitBreakingDelegatingHandler.cs
```diff
@@ -1,48 +1,56 @@
...
-                _logger.LogError("Reached to allowed number of exceptions. Circuit is open", ex);
-                throw;
-            }
-            catch (HttpRequestException ex)
-            {
-                _logger.LogError($"Error in {nameof(PollyCircuitBreakingDelegatingHandler)}.{nameof(SendAsync)}", ex);
-                throw;
-            }
+            _logger.LogError("Reached to allowed number of exceptions. Circuit is open", ex);
+            throw;
```
raman-m, line 48: For what purpose do you re-throw? To be caught by Exception-to-Error logic?

src/Ocelot.Provider.Polly/PollyCircuitBreakingDelegatingHandler.cs
```diff
@@ -1,48 +1,56 @@
...
+        }
+        catch (HttpRequestException ex)
```
raman-m, line 50: This is the last exception type we process. What about the rest of `Exception` objects?
Are they caught by Exception-to-Error logic?

src/Ocelot.Provider.Polly/PollyQoSProvider.cs
```diff
@@ -1,54 +1,81 @@
...
-
-namespace Ocelot.Provider.Polly
-{
-    public class PollyQoSProvider : IPollyQoSProvider
-    {
-        public PollyQoSProvider(DownstreamRoute route, IOcelotLoggerFactory loggerFactory)
-        {
-            AsyncCircuitBreakerPolicy circuitBreakerPolicy = null;
-            if (route.QosOptions.ExceptionsAllowedBeforeBreaking > 0)
+using System.Net;
```
raman-m, line 1: Please, show me the current Usings settings in Visual Studio which put system namespaces on top of all.
We need to use simple Remove and Sort refactoring...

src/Ocelot.Provider.Polly/PollyQoSProvider.cs
```diff
@@ -1,54 +1,81 @@
...
+
+    private CircuitBreaker<HttpResponseMessage> CircuitBreakerFactory(DownstreamRoute route)
+    {
+        AsyncCircuitBreakerPolicy<HttpResponseMessage> exceptionsAllowedBeforeBreakingPolicy = null;
+        if (route.QosOptions.ExceptionsAllowedBeforeBreaking > 0)
+        {
+            var info = $"Route: {GetRouteName(route)}; Breaker logging in {nameof(PollyQoSProvider)}: ";
+
+            exceptionsAllowedBeforeBreakingPolicy = Policy
+                .HandleResult<HttpResponseMessage>(r => _serverErrorCodes.Contains(r.StatusCode))
```
raman-m, line 63: Cannot understand that: `_serverErrorCodes` ...
Why do we handle all these HTTP statuses?
Is that some kind of best practice for Polly?

src/Ocelot/Requester/QosDelegatingHandlerDelegate.cs
```diff
@@ -1,7 +1,8 @@
-using Ocelot.Configuration;
-using Ocelot.Logging;
+using Microsoft.AspNetCore.Http;
+using Ocelot.Configuration;
+using Ocelot.Logging;
 
 namespace Ocelot.Requester
 {
-    public delegate DelegatingHandler QosDelegatingHandlerDelegate(DownstreamRoute route, IOcelotLoggerFactory logger);
+    public delegate DelegatingHandler QosDelegatingHandlerDelegate(DownstreamRoute route, IHttpContextAccessor contextAccessor, IOcelotLoggerFactory loggerFactory);
```
raman-m, line 7:
```suggestion
    public delegate DelegatingHandler QosDelegatingHandler(DownstreamRoute route, IHttpContextAccessor contextAccessor, IOcelotLoggerFactory loggerFactory);
```
So, remove "Delegate" suffix

test/Ocelot.AcceptanceTests/PollyQoSTests.cs
```diff
@@ -1,276 +1,311 @@
...
+        _serviceHandler = new ServiceHandler();
+        _steps = new Steps();
+    }
+
+    [Fact]
+    public void Should_not_timeout()
+    {
+        var port = PortFinder.GetRandomPort();
+
+        var configuration = new FileConfiguration
```
raman-m, line 22: Please, setup configuration being returned by private helper.
A lot of copy-pasted code!

test/Ocelot.UnitTests/Polly/PollyCircuitBreakingDelegatingHandlerTests.cs
```diff
@@ -131,23 +161,215 @@ public FakeAsyncPolicy(string name, object result)
...
+
+        public new IAsyncPolicy WithPolicyKey(string policyKey)
+        {
+            throw new NotImplementedException();
+        }
+
+        public Task ExecuteAsync(Func<Task> action)
+        {
+            throw new NotImplementedException();
+        }
```
raman-m, line 191: Strange redundant empty methods because of the `IAsyncPolicy` interface implementation.
Could we use `Mock<IAsyncPolicy>` or `Mock<AsyncPolicy<TResult>>` instead of `FakeAsyncPolicy<TResult>`?
Is it possible to use mocks to test?

test/Ocelot.AcceptanceTests/PollyQoSTests.cs
```diff
@@ -1,276 +1,311 @@
...
+            .Given(x => _steps.ThenTheStatusCodeShouldBe(HttpStatusCode.ServiceUnavailable))
+            .Given(x => GivenIWaitMilliseconds(3000))
+            .When(x => _steps.WhenIGetUrlOnTheApiGateway("/"))
+            .Then(x => _steps.ThenTheStatusCodeShouldBe(HttpStatusCode.OK))
+            .And(x => _steps.ThenTheResponseBodyShouldBe("Hello from Laura"))
+            .BDDfy();
+    }
+
+    [Fact]
+    public void Open_circuit_should_not_effect_different_route()
```
raman-m, line 194: Did you fix this #1706 unstable test? 😍 
Last fail which known for me is build [2454](https://circleci.com/gh/ThreeMammals/Ocelot/2454)

raman-m (changes requested):
Not bad! Good job! 👍 
Let's do some final improvements...

src/Ocelot.Provider.Polly/CircuitBreaker.cs
ggnaegi: We could just call it OcelotAsyncPolicy? The policies are wrapped together here.

src/Ocelot.Provider.Polly/PollyCircuitBreakingDelegatingHandler.cs
ggnaegi: That's the original logic, I didn't want to mess with it.

src/Ocelot.Provider.Polly/PollyCircuitBreakingDelegatingHandler.cs
ggnaegi: Same as above...

src/Ocelot.Provider.Polly/PollyQoSProvider.cs
ggnaegi: Sorry, I keep forgetting your settings

src/Ocelot.Provider.Polly/PollyQoSProvider.cs
ggnaegi: That was an assumption. Http Status is returned from SendAsync, polly will react on certain types. I thought 500...508 is great.

src/Ocelot/Requester/QosDelegatingHandlerDelegate.cs
ggnaegi: Ok

test/Ocelot.AcceptanceTests/PollyQoSTests.cs
ggnaegi: Ok, factories.

test/Ocelot.UnitTests/Polly/PollyCircuitBreakingDelegatingHandlerTests.cs
ggnaegi: Yeah, ok, that's the original code, let me check that

test/Ocelot.AcceptanceTests/PollyQoSTests.cs
ggnaegi: I hope so, I simplified the logic yes.

Ocelot.sln.DotSettings
ggnaegi: It shouldn't be part of code, it's resharper related, let's put that in the ignore list

test/Ocelot.AcceptanceTests/PollyQoSTests.cs
ggnaegi: @raman-m but that's not my code...

ggnaegi: Let's keep it like that for the minute?

src/Ocelot.Provider.Polly/CircuitBreaker.cs
ggnaegi: @raman-m Ok, since the timeout is part of the policies, it can't be named CircuitBreaker. I changed the names to be more precise: public PollyPolicyWrapper and then AsyncPollyPolicy for the wrapped policies.

Ocelot.sln.DotSettings
raman-m: 😮  Do you use ReSharper to write code for this project?
I **do not** recommend to use JetBrains tools for this project. I believe a developer should use Visual Studio IDE only. The project has standard settings styling, conventions, merge tools etc. for Visual Studio only.
That is why I see so strange diff highlighted by GitHub diff comparer in PRs... Keep in mind, that we have special line-ending settings!

If you use ReSharper IDE, please do not add any ReSharper files to the repo!

src/Ocelot.Provider.Polly/OcelotBuilderExtensions.cs
raman-m: Not changed! The [link](https://github.com/ThreeMammals/Ocelot/pull/1753/files#diff-586799ba26fd065bccd320cb27577637fa83b08a9ec758a28c792294115086b0R7)
It returns `DelegatingHandler`!

src/Ocelot.Provider.Polly/PollyCircuitBreakingDelegatingHandler.cs
raman-m: OK. I hate this Exception-to-Error mapping too. Bad design! We have a lot of problems because of this feature...
Global UnhandledException handler should catch exceptions for logging purposes. Only for rare cases we could catch exception to stabilize the behavior...

src/Ocelot.Provider.Polly/PollyCircuitBreakingDelegatingHandler.cs
raman-m: I mean, we need to ignore other exceptions here, right?

src/Ocelot.Provider.Polly/PollyQoSProvider.cs
raman-m: Not mine only 🤣  for whole community!
I don't see any benefits at all to shuffle usings lines. Keeping this simplest is better. Remove and Sort refactoring is ideal candidate to keep usings good.

src/Ocelot.Provider.Polly/PollyQoSProvider.cs
raman-m: If internal Polly implementations use these status codes (or overrides them) then it is OK for me to process them.

src/Ocelot/Requester/QosDelegatingHandlerDelegate.cs
raman-m: Not OK! 😛 Where is the change? The proof [here](https://github.com/ThreeMammals/Ocelot/pull/1753/files#diff-586799ba26fd065bccd320cb27577637fa83b08a9ec758a28c792294115086b0R7)

Why not to press the magic button "Commit suggestion"? 😉 
The rest of work is renaming all references.
OR
You can rename manually by pressing F2 button onto the class definition. This is the best option.

test/Ocelot.AcceptanceTests/PollyQoSTests.cs
raman-m: Let me help you...

src/Ocelot/Requester/QosDelegatingHandlerDelegate.cs
ggnaegi: I was wondering if we could just remove it. Why this abstraction?

src/Ocelot.Provider.Polly/PollyQoSProvider.cs
ggnaegi: Yes it reacts on the status codes you choose, or other specs from HttpResponseMessage class... With that we can avoid the overhead of throwing an exception.

Ocelot.sln.DotSettings
ggnaegi: @raman-m As I wrote once, I'm on several projects, sometimes I forget... That's why I'm happy to work with quality gates.

test/Ocelot.AcceptanceTests/PollyQoSTests.cs
ggnaegi: Mmmh, I have refactored it already. But it wasn't my code again..

src/Ocelot/Requester/QosDelegatingHandlerDelegate.cs
ggnaegi: Is it of any use?

ggnaegi: I mean, do we have other qos implementations? Why bother?

src/Ocelot.Provider.Polly/PollyCircuitBreakingDelegatingHandler.cs
ggnaegi: Imho, it should be all the exceptions should be catched or none, but now it's, sometimes yes, sometimes no, it's not consistent.

ggnaegi:
@raman-m are you sure that it's not a bit over-engineered?

test/Ocelot.AcceptanceTests/PollyQoSTests.cs
raman-m: Just commit this https://github.com/ThreeMammals/Ocelot/pull/1753/commits/50f9b7a928da0c84e19eacf77dd258c0cc42a33f
I'm sorry, if we did this in parallel.

test/Ocelot.AcceptanceTests/PollyQoSTests.cs
raman-m: Thanks!

raman-m:
> @ggnaegi commented [on Nov 3](#issuecomment-1792422894)
 a bit over-engineered?

Why not if I have some time for this...
I agree that refactoring of tests doesn't provide much benefit.
🆗 I won't you bother on tests... Let's focus on real code...

src/Ocelot/Requester/QosDelegatingHandlerDelegate.cs
raman-m: A developer's joke... `DelegatingHandlerDelegateHandlerDelegateFactory` 😄 
I saw such classes written by Tom.

src/Ocelot.Provider.Polly/PollyCircuitBreakingDelegatingHandler.cs
raman-m: OK Let's ignore for now...

test/Ocelot.UnitTests/Polly/PollyCircuitBreakingDelegatingHandlerTests.cs
raman-m: If refactoring requires a lot of effort, then it is better to ignore it.

.gitignore
```diff
@@ -255,3 +255,4 @@ _templates/
 
 # Test Results
 *.trx
+/Ocelot.sln.DotSettings
```
raman-m, line 258: Note 1

test/Ocelot.UnitTests/Configuration/Validation/FileConfigurationFluentValidatorTests.cs
```diff
@@ -11,7 +11,10 @@
 using Ocelot.UnitTests.Requester;
 using Ocelot.Values;
 using System.Security.Claims;
-using System.Text.Encodings.Web;
+using System.Text.Encodings.Web;
+using Microsoft.AspNetCore.Http;
+using Ocelot.Configuration;
+using Ocelot.Logging;
```
raman-m, line 17: Note 2 Order

test/Ocelot.UnitTests/Polly/PollyPoliciesDelegatingHandlerTests.cs
```diff
@@ -0,0 +1,373 @@
...
+                    ? (TResult)Result
+                    : Activator.CreateInstance<TResult>();
+                return r;
+            }
+
+            var result = await action(context, cancellationToken);
+            return result;
+        }
+
+        public new IAsyncPolicy WithPolicyKey(string policyKey)
```
raman-m, line 181: Note 3

test/Ocelot.UnitTests/Polly/PollyPoliciesDelegatingHandlerTests.cs
```diff
@@ -0,0 +1,373 @@
...
+            bool continueOnCapturedContext)
+        {
+            throw new NotImplementedException();
+        }
+
+        public Task<PolicyResult<TResult1>> ExecuteAndCaptureAsync<TResult1>(Func<Context, CancellationToken, Task<TResult1>> action, Context context, CancellationToken cancellationToken,
+            bool continueOnCapturedContext)
+        {
+            throw new NotImplementedException();
+        }
```
raman-m, line 371: Note 3 end

raman-m (approved):
Great job, Gui! 👍 
You've got approval! 😄 
If you don't mind, I will add a couple of commits...

raman-m:
@ggnaegi Let me know you're 🆗 to merge plz?
I'm going to add a couple of tiny fixes...
So, it seems, this PR will be merged first than #1745  ... And you will need to apply new logger factories for all new logging operators being added by this PR.

.gitignore
raman-m: @ggnaegi 
Could you send me a full list of ReSharper artifacts being added as standard settings by the tool to .NET solutions please?

I'll add them to ignore list...
In future, if someone will convert, or create std setup of ReSharper tools then we readd them and remove from **.gitignore** file.
But for now, I want to make sure that contributors don't change line-endings at all. This is critical settings for code review.

raman-m: What's your vision to fix the problem with line-endings in ReSharper?

raman-m:
Regarding the [Open_circuit_should_not_effect_different_route()](/ggnaegi/Ocelot/blob/bug/1550_circuit_breaker_not_handling_exception_at_end_points/test/Ocelot.AcceptanceTests/PollyQoSTests.cs#L109) test... Here is my testing results of 1000 runs for the test;

image

It is definitely stable now. But I don't know how to get CircleCI environment tests... Seems we have to watch for this test on CircleCI in the latest builds after merging PR...

Anyway, I would say, It should be stable now, and #1706 will **not be** reproducible after merging this PR.
So, we could attach #1706 to this PR.
Could we, Gui?

test/Ocelot.UnitTests/Configuration/Validation/FileConfigurationFluentValidatorTests.cs
raman-m: Fixed in https://github.com/ThreeMammals/Ocelot/pull/1753/commits/6504e3ef6d3c3a1d70a57bc9100f8f7397fa427a

test/Ocelot.UnitTests/Polly/PollyPoliciesDelegatingHandlerTests.cs
raman-m: Fixed in https://github.com/ThreeMammals/Ocelot/pull/1753/commits/4e5f1fa72338f77de206fad812b76a666b3bc2cd ✔️

ggnaegi:
Great!

PR #1786, opened 2023-11-17
#1783 Less logs for circuit breakers (Polly exceptions)

RaynaldM:
## Proposed Changes
Remove try/catch in PollyPoliciesDelegatingHandler
Add a more generic AddPolly<T> to be able to use a specific PollyQoSProvider

PR #1714, opened 2023-10-02
#1712 Bump to Polly 8.0

RaynaldM:
## Closes #1712 
- #1712 

Bump to Polly 8.0 and some code improvements

PR #1182, opened 2020-04-06
#952 #1174 Merge query strings without duplicate values

wast:
## Fixes #952 
- #952 

## Fixes #1174 
- #1174 

## Proposed Changes

  - added method to merge query strings without duplicate values
  - added tests
  - changed logic

PR #1180, opened 2020-04-04
#1179 Add missing documentation for Secured WebSocket

sam9291:
## Fixes #1179 
- #1179 

## Proposed Changes

  - Added missing documentation

PR #1805, opened 2023-11-24
Release 22.0

raman-m:
## October 2023 release

Milestone: **[October'23](https://github.com/ThreeMammals/Ocelot/milestone/1)**
Milestone Status:  **Past due by 24 days** because of #1790 and etc.
Version: **22.0**

## Features
- #1714 
- #1173 
- #1182 
- #1753 
- #1180 
- #1745 
- #1786 

## P.S.
Removed from the release
- #1172


RaynaldM (dismissed):

raman-m:
@ggnaegi @RaynaldM 
Any ideas on release notes?

raman-m:
We've found some low quality and bad design in caching feature by PR #1172 ...

src/Ocelot/Request/Middleware/DownstreamRequest.cs
```diff
@@ -36,7 +37,11 @@ public DownstreamRequest(HttpRequestMessage request)
 
         public string Query { get; set; }
 
-        public HttpContent Content { get; set; }
+        public virtual bool HasContent { get => _request?.Content != null; }
+
+        public virtual Task<string> ReadContentAsync() => HasContent
```
ggnaegi, line 42: I think we shouldn't allow that kind of virtual methods for the purpose of mocks in unit tests. It's virtual so it could be modified and it would hide the original implementation.

src/Ocelot/Cache/CacheKeyGenerator.cs
```diff
@@ -1,20 +1,43 @@
...
-            if (downstreamRequest.Content != null)
+            var builder = new StringBuilder()
+                .Append(downstreamRequest.Method)
+                .Append(Delimiter)
+                .Append(downstreamRequest.OriginalString);
+
+            var cacheOptionsHeader = downstreamRoute?.CacheOptions?.Header;
+            if (!string.IsNullOrEmpty(cacheOptionsHeader))
+            {
+                var header = downstreamRequest.Headers
```
ggnaegi, line 20: I'm not sure it's really what we want. Here we are using the header value to create the MD5 key. I'm not sure the design is safe. Besides, no acceptance tests were provided by the author, so I'm concerned about that change.

src/Ocelot/Request/Middleware/DownstreamRequest.cs
raman-m: Let's double check this after release!
Now I'm going to revert #1172 ...

src/Ocelot/Cache/CacheKeyGenerator.cs
raman-m: Design **wasn't** even not discussed by the team. Raynald had provided review and gave approval.
The author just wanted to merge it. I was not careful allowing to merge without acceptance tests...
We have an incident with dev process violation. This is my fault!

ggnaegi (approved):

PR #1853, opened 2023-12-15
#1724 Reverting back HttpClient full buffering

ggnaegi:
## Follows up #1724 
- #1724 

After testing the new http message invoker pool on a test environment, we noticed that some resources were not being released. After further checks, we realised that some streams were not being closed correctly. 

This doesn't appear in the application with the standard http client, but I made a mistake by adding the ```HttpCompletionOption.ResponseHeadersRead``` property when calling the SendAsync method, which could potentially cause the same problems as with the new message invoker pool. 

A revert should therefore be performed for the time being, and this problem will be addressed in the new message invoker pool.

🙈 🙉 🙊 

## Proposed Changes
- Reverting back `HttpClientWrapper` keeping the full buffering for now. Usage of the new option is out of scope and hazardous!!!

### DIFF
--- a/src/Ocelot/Requester/HttpClientWrapper.cs
+++ b/src/Ocelot/Requester/HttpClientWrapper.cs
@@ -12,13 +12,9 @@ namespace Ocelot.Requester
             Client = client;
         }
 
-        public Task<HttpResponseMessage> SendAsync(HttpRequestMessage request,
-            CancellationToken cancellationToken = default)
+        public Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken = default)
         {
-            // https://www.stevejgordon.co.uk/using-httpcompletionoption-responseheadersread-to-improve-httpclient-performance-dotnet
-            // When using this option, we avoid the intermediate MemoryStream buffer, instead of getting the content directly from the stream exposed on the Socket.
-            // This avoids unnecessary allocations which is a goal in highly optimised situations.
-            return Client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+            return Client.SendAsync(request, cancellationToken);
         }
     }
 }


ggnaegi:
@raman-m please, asap ;-)

raman-m:
> `HttpCompletionOption.ResponseHeadersRead` property 

I cannot get how does the property influence to disposing internal resources? 
If Yes then .NET framework has bugs... I cannot believe in that.
More realistic scenario is...Ocelot with DI has hidden services which consumes and produces and behaves badly.

raman-m:
I'm not going to approve till Raynald's report if he provides some tests in prod.

@RaynaldM **FYI**  We need your report and approve here.

src/Ocelot/Requester/HttpClientWrapper.cs
```diff
@@ -12,13 +12,9 @@ public HttpClientWrapper(HttpClient client)
...
 
-        public Task<HttpResponseMessage> SendAsync(HttpRequestMessage request,
-            CancellationToken cancellationToken = default)
+        public Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken = default)
         {
-            // https://www.stevejgordon.co.uk/using-httpcompletionoption-responseheadersread-to-improve-httpclient-performance-dotnet
-            // When using this option, we avoid the intermediate MemoryStream buffer, instead of getting the content directly from the stream exposed on the Socket.
-            // This avoids unnecessary allocations which is a goal in highly optimised situations.
-            return Client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+            return Client.SendAsync(request, cancellationToken);
```
raman-m, line 17: Can we separate small and large content buffering scenarios?

My idea is
We can separate full buffering and streaming scenarios, then apply
- Small content, small sizes with [0 up to DefaultSizeOfFullBuffering] should use full buffering into the buffer
- Large content, aka unknown Content-Length aka streaming scenario should use only streaming `HttpCompletionOption.ResponseHeadersRead` option

Sounds good?

ggnaegi: Yes, but that's out of scope, I shouldn't have put thoses changes in my last PR. That's something we should discuss later in the message invoker PR.

ggnaegi:
> > `HttpCompletionOption.ResponseHeadersRead` property
> 
> I cannot get how does the property influence to disposing internal resources? If Yes then .NET framework has bugs... I cannot believe in that. More realistic scenario is...Ocelot with DI has hidden services which consumes and produces and behaves badly.

As soon as you decide to handle the stream yourself, then you are responsible of releasing it. It's logical no?

ggnaegi:
> I'm not going to approve till Raynald's report if he provides some tests in prod.
> 
> @RaynaldM **FYI** We need your report and approve here.

@raman-m Raynald has tested a version without these specific changes

ggnaegi:
> > we noticed that some resources **were not being released**.
> 
> were not being **disposed**!

freed willy

raman-m:
I need Ray's input here... There is no problem to revert. The release is still active. This fix is a part of the Nov'23 milestone.

src/Ocelot/Requester/HttpClientWrapper.cs
raman-m: 🆗 let's discuss this design idea when we start Dec'23...

RaynaldM (approved):

RaynaldM:
> I need Ray's input here... There is no problem to revert. The release is still active. This fix is a part of the Nov'23 milestone.

I won't be able to validate these last changes in prod until January 3 or 4.

raman-m:
@RaynaldM  🆗 
In this case, it is better to wait new NuGet packs of Nov'23 release and deploy them and test...
So, Dec'23 version will not be ready on January 3...

PR #1908, opened 2024-01-08
#479 UpstreamPathTemplate hard-coded priority 0

### LINKED ISSUE #479: UpstreamPathTemplate hard-coded priority 0
`UpstreamTemplatePatternCreator.cs:33` creates an instance of `UpstreamPathTemplate` but sets priority to `0` instead of using `reRoute.Priority`. I haven't validated whether this affects behavior, but it seems like a potential bug.

## Specifications

  - Version: 8.0.0

AlyHKafoury:
## Fixes #479 
- #479 

## Proposed Changes
  - Fix UpstreamPathTemplate hard-coded priority 0 : Change the priority during a template path "{" to be route.Priority

### DIFF
--- a/src/Ocelot/Configuration/Creator/UpstreamTemplatePatternCreator.cs
+++ b/src/Ocelot/Configuration/Creator/UpstreamTemplatePatternCreator.cs
@@ -30,7 +30,7 @@ namespace Ocelot.Configuration.Creator
                     //hack to handle /{url} case
                     if (ForwardSlashAndOnePlaceHolder(upstreamTemplate, placeholders, postitionOfPlaceHolderClosingBracket))
                     {
-                        return new UpstreamPathTemplate(RegExForwardSlashAndOnePlaceHolder, 0, false, route.UpstreamPathTemplate);
+                        return new UpstreamPathTemplate(RegExForwardSlashAndOnePlaceHolder, route.Priority, false, route.UpstreamPathTemplate);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


AlyHKafoury:
Closes #479

AlyHKafoury:
[raman-m](https://github.com/raman-m), I have a test case where the Priority is ignored for the catchall in here 

```
        public void should_use_zero_priority()
        {
            var fileRoute = new FileRoute
            {
                UpstreamPathTemplate = "/{catchAll}",
                Priority = 1,
            };

            this.Given(x => x.GivenTheFollowingFileRoute(fileRoute))
                .When(x => x.WhenICreateTheTemplatePattern())
                .Then(x => x.ThenTheFollowingIsReturned("^/.*"))
                .And(x => ThenThePriorityIs(0))
                .BDDfy();
        }
```
should I change this test case because ignoring the priority is not the required behavior anymore ?

raman-m:
@AlyHKafoury commented [on Jan 8](https://github.com/ThreeMammals/Ocelot/pull/1908#issuecomment-1880632369):
> should I change `should_use_zero_priority` test case because ignoring the priority is not the required behavior anymore ?

**No**, you will break current [Catch All Priority](https://ocelot.readthedocs.io/en/latest/features/routing.html#priority:~:text=0%20is%20the%20lowest%20priority%2C%20Ocelot%20will%20always%20use%200%20for%20%2F%7BcatchAll%7D%20Routes%20and%20this%20is%20still%20hardcoded.) feature which should have lowest `0` priority!
I believe you need to make an exception for CatchAll case allowing `0` for CatchAll route. Then [this test](https://github.com/ThreeMammals/Ocelot/blob/develop/test/Ocelot.UnitTests/Configuration/UpstreamTemplatePatternCreatorTests.cs#L51) will turn green.

PR #1927, opened 2024-01-19
#683 Validate placeholder duplicates in path templates

### LINKED ISSUE #683: Ocelot crashes if there are same-named Catch-Alls e.g. `foo/{bar}/{bar}`
## Expected Behavior

An informative exception which tells you you can't reuse catch-all names within a path.

## Actual Behavior

This exception, which doesn't tell you what is wrong and looks like a bug in Ocelot:

```
[15:05:22 FTL] Application startup exception
System.AggregateException: One or more errors occurred. (Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: startIndex) ---> System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: startIndex
   at System.String.IndexOf(String value, Int32 startIndex, Int32 count, StringComparison comparisonType)
   at Ocelot.Configuration.Creator.UpstreamTemplatePatternCreator.Create(IReRoute reRoute)
   at Ocelot.Configuration.Creator.ReRoutesCreator.SetUpDownstreamReRoute(FileReRoute fileReRoute, FileGlobalConfiguration globalConfiguration)
   at Ocelot.Configuration.Creator.ReRoutesCreator.<>c__DisplayClass15_0.<Create>b__0(FileReRoute reRoute)
   at System.Linq.Enumerable.SelectListIterator`2.ToList()
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Ocelot.Configuration.Creator.FileInternalConfigurationCreator.Create(FileConfiguration fileConfiguration)
   at Ocelot.Middleware.OcelotMiddlewareExtensions.CreateConfiguration(IApplicationBuilder builder)
   at Ocelot.Middleware.OcelotMiddlewareExtensions.UseOcelot(IApplicationBuilder builder, OcelotPipelineConfiguration pipelineConfiguration)
   at Ocelot.Middleware.OcelotMiddlewareExtensions.UseOcelot(IApplicationBuilder builder)
```

  1. Create an ocelot.json which contains the following re-route:

```json
        {
            "DownstreamPathTemplate": "/bar/{everything}/{everything}",
            "DownstreamScheme": "http",
            "DownstreamHostAndPorts": [
                {
                    "Host": "foo",
                    "Port": 80
                }
            ],
            "UpstreamPathTemplate": "/foo/bar/{everything}/{everything}",
            "UpstreamHttpMethod": [
                "Get"
            ],
            "Priority": 0
        }
```

  1. Tell Ocelot to load it
  1. Get the above exception

## Specifications

  - Version: 12.0.1
  - Platform: ASP.Net Core 2.1
  - Subsystem:

AlyHKafoury:
## Fixes #683 
  - #683


AlyHKafoury:
@raman-m Draft for a better exception handling in upstreamcreator

AlyHKafoury:
@raman-m if we agree on the first solution which is failing gracefully, then this PR should do exactly that

raman-m:
@AlyHKafoury 
What was the line of code which generates original `ArgumentOutOfRangeException` error?
Have you reproduced the bug?

I'm not sure that simple exception throwing is right... But, yes, technically we provide information about invalid route.
Also, @philproctor recommended [on Jan 10, 2019](https://github.com/ThreeMammals/Ocelot/issues/683#issuecomment-452923573) to update validators. So, having validation error + exception throwing by `UpstreamTemplatePatternCreator` will be one solid solution.

Let's discuss more with the team...
@RaynaldM @ggnaegi  We need your attention here. My design approach is [here](https://github.com/ThreeMammals/Ocelot/issues/683#issuecomment-1900119267), please read it.

AlyHKafoury:
@raman-m I reproduced with the included unit test. the line that fails is this line 
```
var indexOfNextForwardSlash = upstreamTemplate.IndexOf("/", indexOfPlaceholder, StringComparison.Ordinal);
```

the reason is indexOfPlaceholder is -1 because instead of matching the first place holder in the line it matches the last 1 causing a whole set of unexpected behaviors to happen. Even if we like fix this part and allow duplicate placer-holders this will cause alot of issues down the line

raman-m:
Ok, got it! You've reproduced the bug by unit test. Please, postpone work on this. Or review related acceptance tests, if there is no something similar then write a new acceptance test to repro the bug too.

Need to discuss final design with the team.
So, this PR remains as draft.

AlyHKafoury:
@raman-m Other issues to work on ?

raman-m:
`AlyHKafoury:Fix-683`
😃 Instead of free naming convention for feature branch, better to use original issue label which is [![label: bug][~bug]](https://github.com/ThreeMammals/Ocelot/labels/bug)
So, feature name could be `bug-683` or `bug/683`.
Finally, it doesn't matter because it can have any name 🤣 even `683` and it is correct because I can find the issue in backlog by the number.

[~bug]: https://img.shields.io/badge/-bug-ee0701.svg

test/Ocelot.UnitTests/Configuration/UpstreamTemplatePatternCreatorTests.cs
```diff
@@ -233,7 +234,23 @@ public void should_create_template_pattern_that_matches_query_string_with_multip
                 .Then(x => x.ThenTheFollowingIsReturned($"^(?i)/api/subscriptions/[^/]+/updates(|\\?)unitId={MatchEverything}&productId={MatchEverything}$"))
                 .And(x => ThenThePriorityIs(1))
                 .BDDfy();
-        }
+        }
+
+        [Fact]
+        public void should_fail_gracefully_when_placeholder_name_is_duplicate()
```
raman-m, line 240: On what app stage we have the failure?
During route validation stage?
After validation, during `DownstreamRoute` object creation?
Another stage?

raman-m (changes requested):
Hi Aly!
I've marked the PR as ready for reviews, so we can continue development, because the author [replied](https://github.com/ThreeMammals/Ocelot/issues/683#issuecomment-1901115708) to us.
So, please read [new requirements](https://github.com/ThreeMammals/Ocelot/issues/683#issuecomment-1902045480) of linked issue #683 and let's get started active development once again.

test/Ocelot.UnitTests/Configuration/UpstreamTemplatePatternCreatorTests.cs
AlyHKafoury: @raman-m at replacing placeholders with the correct regex sequence

AlyHKafoury:
@raman-m since we are going to catch this early during validation should I remove exception handling at the upstreamceator stage ?

raman-m:
Aly, just comment old version by `//` to keep it in PR changes to compare to your new version.
Later you can make cleaning up.

test/Ocelot.UnitTests/Configuration/UpstreamTemplatePatternCreatorTests.cs
raman-m: This is quite too late stage ))
Please, keep this exception while replacing placeholders by their values. And write `Error` event into the log. I know this exception will be caught by Error Mapper. So, probably we will have 2 records in the log. Need testing this...

Write `Critical` event into the log during validation stage.

AlyHKafoury: @raman-m regarding this part, the route validater builds the route first before validating it, so even if I try to handle this in the route validation it will be even later, is there any earlier stages I should look into, If I try to handle this also at configuration builder stage I will have to duplicate the logic of parsing the placeholder from different places for example the upstreampatterncreator.cs

raman-m: Let's make final decision...
I believe `UpstreamTemplatePatternCreator` is not a right place to check the placeholders. Of course, you catch the problem even earlier than validators, but the issue is mixing responsibilities 👉 
- `UpstreamTemplatePatternCreator` creates a template via replacing template string with converted/parsed regular expression.
- Validators run checking functions via attached validation rules

```diff
! So, we have the violation of Single Responsibility principle here, aka SOLID
```
Please move checking of placeholders uniqueness to a validator please, and define a rule, and remove this exception for sure from `UpstreamTemplatePatternCreator` class. Thanks!

AlyHKafoury: @raman-m Ok I understand your point, but the upstreamtemplate is built before the validator rules are checked, so even if I add the rule in RouteFluentValidator.cs the exception will still happen. What should I do in that part ?

raman-m: Yes, the `RouteFluentValidator` class is the right place.
Please remove exception! Seems I've explained you why... **`UpstreamTemplatePatternCreator`  should have zero changes!**

I meant this live
https://github.com/ThreeMammals/Ocelot/pull/1927/files#diff-611108f6b3268be0a8735448eb7b2b23e994a1bbf18b4a99794a9f720cdd121aR40

AlyHKafoury:
@raman-m this is strange there is already a validator in here : https://github.com/ThreeMammals/Ocelot/blob/f4803c24bf9e9ca3929c78ca8eb23401e3c31c23/src/Ocelot/Configuration/Validator/FileConfigurationFluentValidator.cs#L36

and a test for it in here : 
https://github.com/ThreeMammals/Ocelot/blob/f4803c24bf9e9ca3929c78ca8eb23401e3c31c23/test/Ocelot.UnitTests/Configuration/Validation/FileConfigurationFluentValidatorTests.cs#L1476

The point is the current rule now checks for upstream only should I add also downstream check for this and a unit test for it or are we good with the only checking the upstream

AlyHKafoury:
> @raman-m this is strange there is already a validator in here :
> 
> https://github.com/ThreeMammals/Ocelot/blob/f4803c24bf9e9ca3929c78ca8eb23401e3c31c23/src/Ocelot/Configuration/Validator/FileConfigurationFluentValidator.cs#L36
> 
> and a test for it in here :
> 
> https://github.com/ThreeMammals/Ocelot/blob/f4803c24bf9e9ca3929c78ca8eb23401e3c31c23/test/Ocelot.UnitTests/Configuration/Validation/FileConfigurationFluentValidatorTests.cs#L1476
> 
> The point is the current rule now checks for upstream only should I add also downstream check for this and a unit test for it or are we good with the only checking the upstream

Yes the current problem is with downstream path only being duplicate, so I will add rule for that and add unit test for it

raman-m:
Please, add user scenario of bug #683 

Also, I'd ask you to check the following...
Are there some tests which check that each upstream placeholder is presented in downstream with the same name?
If Yes, it is fine.
If No, we have to add this validation rule too.

test/Ocelot.UnitTests/Configuration/Validation/FileConfigurationFluentValidatorTests.cs
```diff
@@ -1496,7 +1496,61 @@ public void configuration_is_invalid_when_placeholder_is_used_twice_in_upstream_
                 .Then(x => x.ThenTheResultIsNotValid())
                 .And(x => x.ThenTheErrorMessageAtPositionIs(0, "route /foo/bar/{everything}/{everything} has duplicated placeholder"))
                 .BDDfy();
-        }
+        }
+
+        [Fact]
+        [Trait("Bug", "683")]
```
AlyHKafoury, line 1502: @raman-m

test/Ocelot.UnitTests/Configuration/Validation/FileConfigurationFluentValidatorTests.cs
```diff
@@ -1496,7 +1496,61 @@ public void configuration_is_invalid_when_placeholder_is_used_twice_in_upstream_
...
+                },
+            }))
+                .When(x => x.WhenIValidateTheConfiguration())
+                .Then(x => x.ThenTheResultIsNotValid())
+                .And(x => x.ThenTheErrorMessageAtPositionIs(0, "route /bar/{everything}/{everything} has duplicated placeholder"))
+                .BDDfy();
+        }
+
+        [Fact]
+        [Trait("Bug", "683")]
```
AlyHKafoury, line 1529: @raman-m user usecase

src/Ocelot/Configuration/Validator/FileConfigurationFluentValidator.cs
```diff
@@ -101,6 +109,24 @@ private static bool IsPlaceholderNotDuplicatedIn(string upstreamPathTemplate)
             var upstreamPathPlaceholders = matches.Select(m => m.Value);
             return upstreamPathPlaceholders.Count() == upstreamPathPlaceholders.Distinct().Count();
         }
+
+        private static bool SameNamePlaceholdersInUpstreamandDownstream(FileRoute route)
```
AlyHKafoury, line 113: @raman-m this breaks multiple tests is this ok ?

raman-m: But this is your code! Why are you asking me? ))

test/Ocelot.UnitTests/Configuration/Validation/FileConfigurationFluentValidatorTests.cs
raman-m: Good! I 'll review...

src/Ocelot/Configuration/Validator/FileConfigurationFluentValidator.cs
raman-m, line 113:
```suggestion
        private static bool IsPlaceholderDefinedInBothTemplates(FileRoute route)
```

src/Ocelot/Configuration/Validator/FileConfigurationFluentValidator.cs
```diff
@@ -101,6 +109,24 @@ private static bool IsPlaceholderNotDuplicatedIn(string upstreamPathTemplate)
...
+        {
+            var regExPlaceholder = new Regex("{[^}]+}");
```
raman-m, line 115: I don't like this hacky expression!
I'm reading it as: "after opening curly bracket there is a set of chars which is not closing curly bracket, and finally ends with closing curly bracket". :rofl: 

Aly, have you ever heard about matching groups of `RegEx`?
- [Grouping constructs in regular expressions](https://learn.microsoft.com/en-us/dotnet/standard/base-types/grouping-constructs-in-regular-expressions)

I guess the expression like `({\w+})` should work. Then you have to read items from `match.Groups` property.

test/Ocelot.UnitTests/Configuration/Validation/FileConfigurationFluentValidatorTests.cs
```diff
@@ -1496,7 +1496,61 @@ public void configuration_is_invalid_when_placeholder_is_used_twice_in_upstream_
...
+                    },
+                },
+            }))
+                .When(x => x.WhenIValidateTheConfiguration())
+                .Then(x => x.ThenTheResultIsNotValid())
+                .And(x => x.ThenTheErrorMessageAtPositionIs(0, "route /bar/{everything}/{everything} has duplicated placeholder"))
+                .BDDfy();
+        }
+
+        [Fact]
```
raman-m, line 1528: Both tests are very similar.
```suggestion
        [Theory]
```
Why not to merge both cases into one theory?

raman-m (changes requested):
Please, work more and independently :wink: 
When you request code review, the build should be passed. Seems your code is not complete.

src/Ocelot/Configuration/Validator/FileConfigurationFluentValidator.cs
AlyHKafoury: @raman-m Sorry If I was not clear. It seems there are other unit tests that assume that the same Placeholder shouldn't exist in both upstream and downstream, should I change those unit tests because we are changing this feature ?

src/Ocelot/Configuration/Validator/FileConfigurationFluentValidator.cs
AlyHKafoury: I know regular expressions pretty well, I just copied the regex expression used in the function above thinking that since it exists in the code base it means it's ocelot's approved way of doing things, but I don't like it either, I will rewrite it in the correct way and I will fix the older one

src/Ocelot/Configuration/Validator/FileConfigurationFluentValidator.cs
raman-m: Yes, we should!
But what tests are you talking about?
Just convert old tests to a `Theory` and add new `InlineData` for the bug scenario. And, that's it!

src/Ocelot/Configuration/Validator/FileConfigurationFluentValidator.cs
raman-m: Good luck! :wink:

src/Ocelot/Configuration/Validator/FileConfigurationFluentValidator.cs
AlyHKafoury: @raman-m you can check the failing tests from Circle CI these are the tests that I left failing

raman-m: Are you stuck in fixing?
👌 I will check the failed build tomorrow...

raman-m: @AlyHKafoury  🆗 I've checked the 4 failed tests... They are related to service discovery for ServiceFabric etc...
So, DO you want to say me that our new validator breaks all these failed scenarios because the rule require the second placeholder od the pair?

AlyHKafoury: Correct, are we going to change the documentation and break compatibility and change the unit tests or we are gonna remove the code I added ?

raman-m: I don't know... 
You & I have to checkout feature branch and review the tests for carefully...
And we need to make final decision later... 
I guess, we can make exception for ServiceFabric routes, but I'm not sure they are real ServiceFabric ones...

AlyHKafoury: What I think about is that we allow the following : 
downstream: /bar/foo/
upstream: /foo/{placholder}/bar

but disallow the following : 
downstream: /foo/{placholder}/bar
upstream: /bar/foo/

RaynaldM (approved):

raman-m:
@AlyHKafoury Please resolve the merge conflict!

AlyHKafoury:
On it

raman-m:
And merge **develop** to feature branch plz!

raman-m:
@AlyHKafoury  Hi Aly!
How do you do?
What the rest of the work here?

test/Ocelot.UnitTests/Configuration/Validation/FileConfigurationFluentValidatorTests.cs
raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/1927/commits/6ef6b216f22ae335a0d2f5b17c80da456fcbf78b ✔️

src/Ocelot/Configuration/Validator/FileConfigurationFluentValidator.cs
raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/1927/commits/b27ea36ff193d2ba40c13943991fa2053b1aa6e5 ✔️

src/Ocelot/Configuration/Validator/FileConfigurationFluentValidator.cs
raman-m: @AlyHKafoury Your old regex worked, but I modified it to favor my suggestion of using `\{\w+\}`, no groups

raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/1927/commits/68ba69a222f9dfe4ef094803dd2348c9b19d97da ✔️

raman-m:
**TODO**: Tomorrow I will continue fixing of failed Acceptance tests... A typical failure is `Unable to start Ocelot, errors are: xxx`

src/Ocelot/Configuration/Validator/FileConfigurationFluentValidator.cs
raman-m: >@AlyHKafoury [on Jan 31](https://github.com/ThreeMammals/Ocelot/pull/1927#discussion_r1472732952)

Did you mean?
>upstream: /foo/{placeholder}/bar
>downstream: /bar/foo/{placeholder}

As far as I know, we don't allow empty placeholders in the middle of the path...
I haven't seen such logic... But we allow empty Catch All

----
>upstream: /foo/{placeholder}/bar

It doesn't make sense because when empty placeholder then this  is border case of `/foo/bar` URL path and better to define static route with this upstream path.

Allowing empty placeholders in the middle requires huge refactoring... I don't see much sense in this feature.

test/Ocelot.UnitTests/Configuration/UpstreamTemplatePatternCreatorTests.cs
raman-m: It seems this has already been fixed. We've changed design to validate route JSON during Ocelot starting.
So, no exceptions in Ocelot kernel now. Good! 👍

src/Ocelot/Configuration/Creator/UpstreamTemplatePatternCreator.cs
```diff
@@ -34,6 +34,11 @@ public UpstreamPathTemplate Create(IRoute route)
                     }
                 }
             }
+
+            if(placeholders.Count() > placeholders.Distinct().Count())
+            {
+                throw new Exception("Please use unique placeholder names in your templates, invalid template : " + upstreamTemplate);
+            }
```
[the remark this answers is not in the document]
raman-m: Outdated. **Not an issue anymore!** No changes in this class.

raman-m:
Aly, where are you? Seems you live in Egypt... So, your time zone is +2.
Ok, if you are not interested in this contribution then just ignore my comments.
I will deliver this bug fix by my own.

src/Ocelot/Configuration/Validator/FileConfigurationFluentValidator.cs
raman-m: @AlyHKafoury Today I understood finally your concern regarding Service Fabric tests failing! 🤩 
So, I've updated the validators and fixed tests in commit https://github.com/ThreeMammals/Ocelot/pull/1927/commits/643af3c09e1f9877a1cb4f483e8b18903b4d9b0b ✔️

raman-m:
### TODO
- [ ] Fix integration tests
- [ ] Release notes

Issue #1496, opened 2021-07-22
Route key is not unique when using Host header for service discovery


sliekens:
## Expected Behavior
The `RouteKey` format should include a route's `ServiceName` when you use service discovery: `"UpstreamPathTemplate|UpstreamHttpMethod|ServiceName"`.

## Actual Behavior

The default format is `"UpstreamPathTemplate|UpstreamHttpMethod|DownstreamHostAndPorts"`. The last section resolves to an empty string when you use service discovery: `"/some-path|GET|"`.

This leads to conflicts when multiple routes have the same upstream method and path template, but are discriminated by an `UpstreamHost`.

## Steps to Reproduce the Problem

1. Create two routes with identical `UpstreamPathTemplate` and `UpstreamHttpMethod`
2. Give each route a unique `UpstreamHost`
3. Do **not** configure the `DownstreamHostAndPorts`, instead configure a `ServiceName` and a `ServiceDiscoveryProvider`.

``` json
{
  "GlobalConfiguration": {
    "ServiceDiscoveryProvider": {
      "Scheme": "https",
      "Host": "localhost",
      "Port": 8500,
      "Type": "Consul"
    }
  },
  "Routes": [
    {
      "UpstreamHost": "api.contoso.de",
      "ServiceName": "BackofficeGermany",
      "UpstreamPathTemplate": "/some-path",
      "UpstreamHttpMethod": [ "Get" ],
      "DownstreamPathTemplate": "/some/path",
      "DownstreamScheme": "https"
    },
    {
      "UpstreamHost": "api.contoso.fr",
      "ServiceName": "BackofficeFrance",
      "UpstreamPathTemplate": "/some-path",
      "UpstreamHttpMethod": [ "Get" ],
      "DownstreamPathTemplate": "/some/path",
      "DownstreamScheme": "https"
    }
  ]
}
```

Now when you send two requests, once for each route:

```
GET /some-path HTTP/1.1
Host: api.contoso.de


GET /some-path HTTP/1.1
Host: api.contoso.fr


```

The first request initiates service discovery, which resolves the host and port for `BackofficeGermany`. The resolved configuration is stored in the `ILoadBalancerHouse` with the key `"/some-path|GET|"` .

The second request has the same route key so the `ILoadBalancerHouse` entirely skips the service discovery step. Instead it reuses the configuration for `BackofficeGermany`, which is wrong.

## Specifications

  - Version: 16.0.1
  - Platform: netcoreapp3.1
[...]

PR #1944, opened 2024-01-27
#849 #1496 Extend the route key format used for load balancing making it unique

### LINKED ISSUE #849: When I use Consul service discovery `UpstreamHost` property isn't working
When I use DownstreamHostAndPorts directly it works,The following two links point to different services.
article services: [http://gateway.afuyy.com/test](url)
live service: [http://test.gateway.afuyy.com/test](url)
image

when i use consul serviceDiscovery  UpstreamHost not working,Both of the following links point to the same service.
 [http://consulgateway.afuyy.com/test](url)
[http://test.consulgateway.afuyy.com/test](url)
image

image
image

## Specifications

  - Version: 
           ocelot version: 12.0.1
           .netcore version: 2.0
  - Platform:
  - Subsystem:

sliekens:
## Fixes #849 #1496 
- #849
- #1496 

## Proposed Changes
  - Make the `RouteKeyCreator` aware of `UpstreamHost` routing
  - Make the `RouteKeyCreator` aware of ServiceDiscovery with `ServiceName` and `ServiceNamespace`

Multiple routes with identical `UpstreamMethod` and `UpstreamPath`, but different `UpstreamHost` and `ServiceName` or `DownstreamHostAndPort` will now be load-balanced correctly.

### DIFF
--- a/src/Ocelot/Configuration/Creator/RouteKeyCreator.cs
+++ b/src/Ocelot/Configuration/Creator/RouteKeyCreator.cs
@@ -24,33 +24,18 @@ namespace Ocelot.Configuration.Creator
                 Append(upstreamHttpMethod, ',');
             }
 
-            Append(fileRoute.UpstreamHost);
             Append(fileRoute.UpstreamPathTemplate);
-
-            if (!string.IsNullOrEmpty(fileRoute.ServiceName))
-            {
-                Append(fileRoute.ServiceNamespace);
-                Append(fileRoute.ServiceName);
-            }
-            else
-            {
-                foreach (var downstreamHostAndPorts in fileRoute.DownstreamHostAndPorts)
-                {
-                    Append(downstreamHostAndPorts.Host);
-                    Append(downstreamHostAndPorts.Port.ToString(), ':');
-                }
-            }
+            Append(fileRoute.UpstreamHost);
+            Append(fileRoute.ServiceNamespace);
+            Append(fileRoute.ServiceName);
+            Append(fileRoute.LoadBalancerOptions.Type);
+            Append(fileRoute.LoadBalancerOptions.Key);
 
             return keyBuilder.ToString();
 
             // Helper function for appending a part to the key, automatically inserts a separator as needed
             void Append(string next, char separator = '|')
             {
-                if (string.IsNullOrEmpty(next))
-                {
-                    return;
-                }
-
                 if (keyBuilder.Length > 0)
                 {
                     keyBuilder.Append(separator);
--- a/src/Ocelot/Configuration/Creator/RouteKeyCreator.cs
+++ b/src/Ocelot/Configuration/Creator/RouteKeyCreator.cs
@@ -18,18 +18,24 @@ namespace Ocelot.Configuration.Creator
                 return $"{nameof(CookieStickySessions)}:{fileRoute.LoadBalancerOptions.Key}";
             }
 
+            // Build the key from the route's properties using the format:
+            // UpstreamHttpMethod|UpstreamPathTemplate|UpstreamHost|ServiceNamespace|ServiceName|LoadBalancerType|LoadBalancerKey
             var keyBuilder = new StringBuilder();
+
+            // UpstreamHttpMethod and UpstreamPathTemplate are required
             foreach (var upstreamHttpMethod in fileRoute.UpstreamHttpMethod)
             {
                 Append(upstreamHttpMethod, ',');
             }
 
             Append(fileRoute.UpstreamPathTemplate);
-            Append(fileRoute.UpstreamHost);
-            Append(fileRoute.ServiceNamespace);
-            Append(fileRoute.ServiceName);
-            Append(fileRoute.LoadBalancerOptions.Type);
-            Append(fileRoute.LoadBalancerOptions.Key);
+
+            // Other properties are optional, use constants for missing values to aid debugging
+            Append(CoalesceNullOrWhiteSpace(fileRoute.UpstreamHost, "no-host"));
+            Append(CoalesceNullOrWhiteSpace(fileRoute.ServiceNamespace, "no-svc-ns"));
+            Append(CoalesceNullOrWhiteSpace(fileRoute.ServiceName, "no-svc-name"));
+            Append(CoalesceNullOrWhiteSpace(fileRoute.LoadBalancerOptions.Type, "no-lb-type"));
+            Append(CoalesceNullOrWhiteSpace(fileRoute.LoadBalancerOptions.Key, "no-lb-key"));
 
             return keyBuilder.ToString();
 
@@ -43,6 +49,8 @@ namespace Ocelot.Configuration.Creator
 
                 keyBuilder.Append(next);
             }
+
+            string CoalesceNullOrWhiteSpace(string first, string second) => string.IsNullOrWhiteSpace(first) ? second : first;
         }
     }
 }
--- a/src/Ocelot/Configuration/Creator/RouteKeyCreator.cs
+++ b/src/Ocelot/Configuration/Creator/RouteKeyCreator.cs
@@ -43,13 +43,14 @@ namespace Ocelot.Configuration.Creator
 
             return keyBuilder.ToString();
 
-            // Helper method for appending a part to the key, automatically inserts a separator as needed
+            // Helper function for appending a part to the key, automatically inserts a separator as needed
             void Append(string next, char separator = '|')
             {
                 if (string.IsNullOrEmpty(next))
                 {
                     return;
                 }
+
                 if (keyBuilder.Length > 0)
                 {
                     keyBuilder.Append(separator);
--- a/src/Ocelot/Configuration/Creator/RouteKeyCreator.cs
+++ b/src/Ocelot/Configuration/Creator/RouteKeyCreator.cs
@@ -5,13 +5,54 @@ namespace Ocelot.Configuration.Creator
 {
     public class RouteKeyCreator : IRouteKeyCreator
     {
-        public string Create(FileRoute fileRoute) => IsStickySession(fileRoute)
-            ? $"{nameof(CookieStickySessions)}:{fileRoute.LoadBalancerOptions.Key}"
-            : $"{fileRoute.UpstreamPathTemplate}|{string.Join(',', fileRoute.UpstreamHttpMethod)}|{string.Join(',', fileRoute.DownstreamHostAndPorts.Select(x => $"{x.Host}:{x.Port}"))}";
+        public string Create(FileRoute fileRoute)
+        {
+            var isStickySession = fileRoute.LoadBalancerOptions is
+            {
+                Type: nameof(CookieStickySessions),
+                Key.Length: > 0
+            };
 
-        private static bool IsStickySession(FileRoute fileRoute) =>
-            !string.IsNullOrEmpty(fileRoute.LoadBalancerOptions.Type)
-            && !string.IsNullOrEmpty(fileRoute.LoadBalancerOptions.Key)
-            && fileRoute.LoadBalancerOptions.Type == nameof(CookieStickySessions);
+            if (isStickySession)
+            {
+                return $"{nameof(CookieStickySessions)}:{fileRoute.LoadBalancerOptions.Key}";
+            }
+
+            var keyBuilder = new StringBuilder();
+            foreach (var upstreamHttpMethod in fileRoute.UpstreamHttpMethod)
+            {
+                Append(upstreamHttpMethod, ',');
+            }
+
+            Append(fileRoute.UpstreamHost);
+            Append(fileRoute.UpstreamPathTemplate);
+
+            if (!string.IsNullOrEmpty(fileRoute.ServiceName))
+            {
+                Append(fileRoute.ServiceNamespace);
+                Append(fileRoute.ServiceName);
+            }
+            else
+            {
+                foreach (var downstreamHostAndPorts in fileRoute.DownstreamHostAndPorts)
+                {
+                    Append(downstreamHostAndPorts.Host);
+                    Append(downstreamHostAndPorts.Port.ToString(), ':');
+                }
+            }
+
+            return keyBuilder.ToString();
+
+            // Helper method for appending a part to the key, automatically inserts a separator as needed
+            void Append(string next, char separator = '|')
+            {
+                if (keyBuilder.Length > 0)
+                {
+                    keyBuilder.Append(separator);
+                }
+
+                keyBuilder.Append(next);
+            }
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


test/Ocelot.UnitTests/Configuration/RouteKeyCreatorTests.cs
```diff
@@ -57,7 +57,52 @@ public void should_return_re_route_key()
...
+                    new()
+                    {
+                        Host = "localhost",
+                        Port = 123,
+                    },
+                    new()
+                    {
+                        Host = "localhost",
+                        Port = 123,
+                    },
```
raman-m, line 83: Duplicate?

src/Ocelot/Configuration/Creator/RouteKeyCreator.cs
```diff
@@ -1,17 +1,58 @@
...
+
+            // Helper method for appending a part to the key, automatically inserts a separator as needed
+            void Append(string next, char separator = '|')
+            {
+                if (keyBuilder.Length > 0)
+                {
+                    keyBuilder.Append(separator);
+                }
+
+                keyBuilder.Append(next);
```
raman-m, line 54: You don't analyze empty value of `next` argument!
So, when empty value then we add separator only. Is it right?

src/Ocelot/Configuration/Creator/RouteKeyCreator.cs
```diff
@@ -1,17 +1,58 @@
...
+                    Append(downstreamHostAndPorts.Port.ToString(), ':');
+                }
+            }
+
+            return keyBuilder.ToString();
+
+            // Helper method for appending a part to the key, automatically inserts a separator as needed
+            void Append(string next, char separator = '|')
+            {
+                if (keyBuilder.Length > 0)
```
raman-m, line 49:
```suggestion
                if (string.IsNullOrEmpty(next))
                {
                    return;
                }
                if (keyBuilder.Length > 0)
```

raman-m (commented):
Steven,
Your proposal for the key format is
```
UpstreamHttpMethod,+|UpstreamHost|UpstreamPathTemplate|ServiceNamespace|ServiceName
OR
UpstreamHttpMethod,+|UpstreamHost|UpstreamPathTemplate|Host:Port,+
```
Maybe to include Type of Load Balancer for the second variant instead of `Host:Port`?
Because `UpstreamPathTemplate` cannot very, so it is unique, seems `Host:Port` is redundant... and type of load balancer cannot vary too...

raman-m:
@ggnaegi Welcome to code review!

ggnaegi:
> @ggnaegi Welcome to code review!

@raman-m yes but a I said i will have some time on Wednesday night.

test/Ocelot.UnitTests/Configuration/RouteKeyCreatorTests.cs
sliekens: I copied it from the existing test. I'll change the port of the second downstream host for clarity.

src/Ocelot/Configuration/Creator/RouteKeyCreator.cs
sliekens: Yea, I couldn't decide which is better, keeping or removing the empty parts. Keeping them felt like a safe choice, but I removed them now.

sliekens:
> Steven, Your proposal for the key format is
> 
> ```
> UpstreamHttpMethod,+|UpstreamHost|UpstreamPathTemplate|ServiceNamespace|ServiceName
> OR
> UpstreamHttpMethod,+|UpstreamHost|UpstreamPathTemplate|Host:Port,+
> ```
> 
> Maybe to include Type of Load Balancer for the second variant instead of `Host:Port`? Because `UpstreamPathTemplate` cannot vary, so it is unique, seems `Host:Port` is redundant... and type of load balancer cannot vary too...

---

I thought about it, what I want to do is make the route key format the same for all types of configs, because the differences in the format are causing me confusion.

Proposed format, which now includes the load balancer options instead of downstream host/port:

```
# sticky sessions
CookieStickySessions:LoadBalancerOptions.Key

# everything else
UpstreamHttpMethod|UpstreamPathTemplate|UpstreamHost|ServiceNamespace|ServiceName|LoadBalancerOptions.Type|LoadBalancerOptions.Key
```

However, for this to work correctly, empty parts must be kept, because removing empty parts will again lead to ambiguous route keys.

For example:

```
GET|/categories|my-api|LeastConnection
```

It's not clear if this means `my-api` is the `UpstreamHost` or `ServiceName`. The empty parts **must** be kept to solve this ambiguity.

src/Ocelot/LoadBalancer/LoadBalancers/LoadBalancerHouse.cs
```diff
@@ -22,8 +22,6 @@ public Response<ILoadBalancer> Get(DownstreamRoute route, ServiceProviderConfigu
 
                 if (_loadBalancers.TryGetValue(route.LoadBalancerKey, out var loadBalancer))
                 {
-                    loadBalancer = _loadBalancers[route.LoadBalancerKey];
```
sliekens, line 25: @raman-m I removed this line, the `out` var already contained the correct reference.

sliekens:
I pushed a few more changes. The route key is now calculated uniformly for all kinds of routing, except sticky session routing which uses a key from configuration, not a calculated key.

The `DownstreamHostAndPort` have been replaced by `LoadBalancerOptions.Type` and `LoadBalancerOptions.Key`.

The format is now:

`UpstreamHttpMethod|UpstreamPathTemplate|UpstreamHost|ServiceNamespace|ServiceName|LoadBalancerType|LoadBalancerKey`

For undefined route properties, the empty parts are replaced by default values:

``` diff
-GET,POST,PUT|/api/product|||||
+GET,POST,PUT|/api/product|no-host|no-svc-ns|no-svc-name|no-lb-type|no-lb-key
```

The default values don't serve any technical purpose, but it came up in the review that it empty parts look like a mistake when it's actually not.

src/Ocelot/Configuration/Creator/RouteKeyCreator.cs
```diff
@@ -1,17 +1,60 @@
...
+            Append(CoalesceNullOrWhiteSpace(downstreamHostAndPorts, "no-host-and-port"));
+            Append(CoalesceNullOrWhiteSpace(fileRoute.ServiceNamespace, "no-svc-ns"));
+            Append(CoalesceNullOrWhiteSpace(fileRoute.ServiceName, "no-svc-name"));
+            Append(CoalesceNullOrWhiteSpace(fileRoute.LoadBalancerOptions.Type, "no-lb-type"));
+            Append(CoalesceNullOrWhiteSpace(fileRoute.LoadBalancerOptions.Key, "no-lb-key"));
+
+            return keyBuilder.ToString();
+
+            // Helper function to append a string to the keyBuilder, separated by a pipe
+            void Append(string next)
```
RaynaldM, line 43: local methods could be static ?

RaynaldM (approved):

raman-m:
> @sliekens added 4 commits [on Jan 28](https://github.com/ThreeMammals/Ocelot/pull/1944#commits-pushed-2f030fa)

Thanks for issues resolving and working on code review more!
I'm busy these days... and sorry, I'll review after the release we have this week...

sliekens:
Let me know if you have any more suggestions. Otherwise, I'm happy with the current state of this branch.

ggnaegi (approved):
ok, thanks

RaynaldM (approved):

raman-m (changes requested):
> @raman-m do you have time this week to review?

@sliekens  Yes, I do..now... :wink: 
We have some unit tests. Nice!
It would be good if we add 1 acceptance test too.
I guess we have some acceptance tests already: so we could update old ones.
But better to add new test which covers `UpstreamHost` user scenario and other props of the route.

sliekens:
@raman-m I added a test with two services, 

I have verified that it fails as expected on the **develop** branch, before my changes:

image

When I run the test on this branch, it passes.

raman-m:
>@sliekens commented [on Feb 16](https://github.com/ThreeMammals/Ocelot/pull/1944#issuecomment-1947566536)

Thank you! Good job! 💪 
Having acceptance tests is a green light, and tomorrow I'll perform final code review with possible a few small commits checking styling & conventions, minor code improvements...
Also, I hope it will be merged tomorrow too 😉

raman-m:
Questionz
- Have you checked different types of load balancers? See [LoadBalancers](https://github.com/ThreeMammals/Ocelot/tree/develop/src/Ocelot/LoadBalancer/LoadBalancers) folder or [search](https://github.com/search?q=repo%3AThreeMammals%2FOcelot+%22%3A+ILoadBalancer%22+path%3A%2F%5Esrc%5C%2FOcelot%5C%2FLoadBalancer%5C%2FLoadBalancers%5C%2F%2F&type=code) for them
- Which type were you focused on?

sliekens:
I did all of my testing with the `LeastConnection` load balancer. However I want to emphasize that the bug was in the `LoadBalancerHouse`, so the issue should present itself with **any** load balancer, probably including the `NoLoadBalancer` implementation of `ILoadBalancer`.

raman-m:
>so the issue should present itself with **any** load balancer

I've converted the acceptance fact to theory:
https://github.com/sliekens/Ocelot/blob/bugfix/route-keys/test/Ocelot.AcceptanceTests/ServiceDiscoveryTests.cs#L461-L468
And now it also passes. So, the theory is 🟢 
I expected that at least one `"CookieStickySessions"` test fails, but it is green too.
Could you explain me why the theory works for all types of load balancers?

Also, in debug session I noticed that key is build always without reading from the cache...
Seems the acceptance test scenario should be longer! It must contain more steps!
Otherwise it is quite useless.

raman-m:
Based on docs in case of [Upstream Host](https://ocelot.readthedocs.io/en/latest/features/routing.html#upstream-host) in a route
The quote:
>This works by looking at the `Host` header the client has used and then using this as part of the information we use to identify a Route.

and
> The Route above will only be matched when the Host header value is `somedomain.com`.

and 
> If you do not set **UpstreamHost** on a Route then any `Host` header will match it.

Question: **How do you setup the header in acceptance tests?**
Is it added implicitly?

sliekens:
> Question: **How do you setup the header in acceptance tests?**
> Is it added implicitly?

Yep, it is added implicitly, when you do this:

``` csharp
_ocelotClient.GetAsync("http://us-shop/products");
```

It is translated to a request like:

``` http
GET /products HTTP/1.1
Host: us-shop
```

raman-m:
>That is intended. 

Intended not to load balancer? )) But this is the test about load balancing 🤣 

----
>The cache should not be used when calling two different services exactly once.

Clear!

----
> I could add a step to call the same service more than once, which then uses the cache instead of making another call to Consul.

Yes, please make the test complete.

-----
>However, I felt that it is already sufficiently covered by other existing tests.

Other acceptance tests even use counter approach to check times of calling!
Moreover, even your test regarding UpstreamHost should check the number of calls via counter.

test/Ocelot.AcceptanceTests/ServiceDiscoveryTests.cs
```diff
@@ -447,19 +448,119 @@ public void should_handle_request_to_poll_consul_for_downstream_service_and_make
...
+                    {
+                        Scheme = "http",
+                        Host = "localhost",
+                        Port = consulPort,
+                    },
+                },
+            };
+
+            this.Given(x => x.GivenThereIsAServiceRunningOn(downstreamServiceUrlUS, "/", 200, responseBodyUS))
+                .And(x => x.GivenThereIsAServiceRunningOn(downstreamServiceUrlEU, "/", 200, responseBodyEU))
```
raman-m, line 547: I think it is wrong!
Pay attention to other tests which use 2 services.
Valid setup should be:
```suggestion
            this.Given(x => x.GivenProductServiceOneIsRunning(downstreamServiceUrlUS, "/", 200, responseBodyUS))
                .And(x => x.GivenProductServiceTwoIsRunning(downstreamServiceUrlEU, "/", 200, responseBodyEU))
```

In current `GivenThereIsAServiceRunningOn` helper you use shared private variable `_downstreamPath` deciding on status OK. But I guess it is wrong!

raman-m (changes requested):
I found major issue with usage of `GivenThereIsAServiceRunningOn` which is based on the one private variables. But the status of both downstream services must use different private variables!

sliekens:
> My expectation that `ServiceName` should be the same as `products` or `product`, but they differ: `product-us` and `product-eu`...
> Is my assumption wrong? I see that you use two Consul service entries... Or in your scenario must Consul have 2 entries?
> So, I'm confused by different `ServiceName`s in both routes...

The setup is: two different services, with different names, and 2 Consul entries. Requests to Ocelot with `Host: us-shop` should re-route  routed to `product-us` while

sliekens:
> Intended not to load balancer? )) But this is the test about load balancing 🤣

@raman-m with the explanation above, do you still feel the test is incomplete? While it is true that the bug is in the `LoadBalancerHouse`, it actually has very little to do with load balancing, which is known and proven to work by other existing tests.

raman-m:
>@sliekens commented [on Feb 28](https://github.com/ThreeMammals/Ocelot/pull/1944#issuecomment-1969687245)

I love this explanation! ❤️ 😉 Thanks!

------
> PS: I originally said that calls to Consul are cached, but that is false. Only responses from `ILoadBalancerFactory` are cached. 

Correct!

---
>The created `ILoadBalancer` still calls Consul each time. I think this is the intended design and correct.

Correct! Default implicit type of poller is `Ocelot.Provider.Consul.`[Consul](https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot.Provider.Consul/Consul.cs#L8) which behaves as transient service calling Consul each request. There is another `PollConsul` type which has caching mechanism.

raman-m:
@sliekens [wrote](https://github.com/ThreeMammals/Ocelot/pull/1944#issuecomment-1969958568)

>@raman-m with the explanation above, do you still feel the test is incomplete? While it is true that the bug is in the `LoadBalancerHouse`, it actually has very little to do with load balancing, which is known and proven to work by other existing tests.

After your adding last 4 commits the test is complete now! It looks great! 😍 
Regarding `LoadBalancerHouse` class... OK, we 've fixed it. But yesterday I noticed one Clean Code issue aka code smell related to SOLID vs DRY princibles. Going to fix it as the last commit of the PR...

----
>/**edit**: I added a Consul request counter and repetitive Ocelot calls, to verify load balancing and service discovery are working correctly. I hope these changes are in line with your expectations, otherwise don't hesitate to comment.

Now the acceptance test is awesome thing! It shows most complete user scenario which can be described for UpstreamHost case.
Thanks for refactoring the test! 🤩 

Finally, I'd say: we have **Done** development!
Thank you for your great contribution! 😋

raman-m (approved):
Ready for delivery! ✅ 
- Code review ✔️ 
- Team approvals ✔️ 
- Unit tests ✔️ 
- Acceptance test(s) ✔️

Issue #928, opened 2019-06-14
Chunked encoding


szilardd:
## Expected Behavior / New Feature

Chunked encoding

## Actual Behavior / Motivation for New Feature

Would be great to have support for chunked encoding, to be used in case of endpoints that stream (download) large files in chunks (instead of storing the whole file in memory and then downloading it).

My team is interested in integrating this functionality if it is doable.

Based on this [comment](https://github.com/ThreeMammals/Ocelot/issues/403#issuecomment-400809167), there was an issue related to it and it was not a priority at that time to make it work.

Is this true, meaning that it can be done ? If so, would you be willing to provide some tips or a starting point on what and where needs to be changed and maybe about the complexity/effort required ?

Thanks

szilardd:
any tips would be appreciated. thanks

raman-m:
@ggnaegi  We must update docs after merging both #1724 and #1824 :warning:

ggnaegi:
@raman-m first check if the issue is solved then merge the docs.

raman-m:
@ggnaegi Welcome to check this, make sure, and finally deliver this feature!

raman-m:
@ggnaegi I believe so. I noticed some acceptance tests for chunked encoding, and if they cover real scenarios, the remaining work is just reviewing and updating the documentation:  
- Remove the Not Supported section for [Chunked Encoding](https://ocelot.readthedocs.io/en/develop/introduction/notsupported.html#chunked-encoding).  
- Consider adding a note to the [Errors and Gotchas](https://ocelot.readthedocs.io/en/develop/features/routing.html#errors-and-gotchas) section under Routing, or create a new Supported subsection. 
- Possibly explain how we process and generate [Content-Length](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Length) header

PR #1972, opened 2024-02-25
#1971 #928 Avoid content if original request has no content and avoid Transfer-Encoding: chunked if Content-Length is known

alexreinert:
## Follow-up to #1971 
- #1971 

With the changes in Ocelot [23.0](https://github.com/ThreeMammals/Ocelot/releases/tag/23.0.0) the `RequestMapper` sets the Property `Content` using the new `StreamHttpContent`. Because the **Content-Length** is not computed in `StreamHttpContent` the header **Transfer-Encoding**: chunked will be set on all upstream requests even if there is no content in the downstream request (e.g. on GET requests). This causes issues when the upstream is an IIS working as reverse proxy (using ARR), in that case all GET requests are failing with status code 502.7 inside ARR.

Additionally in some cases the **Content-Length** is required by the upstream server, it will send status code 411 if it is missing, behind Ocelot all request will fail with status code 411, as the **Content-Length** was not transferred to the upstream server and instead **Transfer-Encoding**: chunked is sent to the upstream server.

Both issue are addressed with this PR.

## Related issues
- #928


raman-m:
Thanks for PR recreation following our development process!

alexreinert:
> @alexreinert Ok, fine for the first part, we shouldn't have the headers if no content. And we should ensure that.

No, you have not and I have a wireshark dump, that proves that. If you set the Request.Content to a HttpContent, the framework will add headers when sending the request. Depending on TryComputeLength it will be Content-Length or Transfer-Encoding.
To avoid these headers you have to set Request.Content to null.

> About the 411 status, I need to investigate that a bit more. We need to support streaming. I'm a bit concerned about the changes you made, possibly breaking this feature.
Please go forward, I tested it, and it worked. But I have no good idea how to test that automatically. 

> As of GET and DELETE without body, it's a large debate, as I worked with some APIs expecting a body with the GET method (eg. Elasticsearch).
Sure, but I in my opinion it is not up to Ocelot to change the request and add content. If the client request with empty content, Ocelot should send the request with empty content and if the client requests with no content, Ocelot should send the request without content. And if you look in my acceptance tests you will see, that my code does exactly that.

ggnaegi:
@alexreinert Ok, fine for the first part, we shouldn't have the headers if no content. And we should ensure that.

About the 411 status, I need to investigate that a bit more. We need to support streaming. I'm a bit concerned about the changes you made, possibly breaking this feature.

As of GET and DELETE without body, it's a large debate - and not explicitely forbidden -, as I worked with some APIs expecting a body with the GET method (eg. Elasticsearch).

ggnaegi:
> > @alexreinert Ok, fine for the first part, we shouldn't have the headers if no content. And we should ensure that.
> 
> No, you have not and I have a wireshark dump, that proves that. If you set the Request.Content to a HttpContent, the framework will add headers when sending the request. Depending on TryComputeLength it will be Content-Length or Transfer-Encoding. To avoid these headers you have to set Request.Content to null.

Mmh, I was just writing that you were right on that part: We should ensure that, "should" meaning it's not the case yet.

> > About the 411 status, I need to investigate that a bit more. We need to support streaming. I'm a bit concerned about the changes you made, possibly breaking this feature.
> 
> Please go forward, I tested it, and it worked. But I have no good idea how to test that automatically.

Yes, it worked for you. You need to understand that we tested it too and we were very careful with the release since we refactored some important parts of the application. So now, I would like to make sure we are not introducing bugs with your PR.

src/Ocelot/Request/Mapper/StreamHttpContent.cs
```diff
@@ -8,25 +8,24 @@ public class StreamHttpContent : HttpContent
     private const int DefaultBufferSize = 65536;
     public const long UnknownLength = -1;
     private readonly HttpContext _context;
+    private readonly long _contentLength;
```
ggnaegi, line 11: Ok, it's fine like this, but I would prefer using Headers.ContentLength. This value is set when calling AddContentHeaders class.

src/Ocelot/Request/Mapper/StreamHttpContent.cs
```diff
@@ -8,25 +8,24 @@ public class StreamHttpContent : HttpContent
...
     protected override async Task SerializeToStreamAsync(Stream stream, TransportContext context)
-        => await CopyAsync(_context.Request.Body, stream, Headers.ContentLength ?? UnknownLength, false,
-            CancellationToken.None);
+        => await CopyAsync(_context.Request.Body, stream, _contentLength, false, CancellationToken.None);
 
     protected override bool TryComputeLength(out long length)
     {
-        length = -1;
-        return false;
+        length = _contentLength;
```
ggnaegi, line 27: I would prefer ```Headers.ContentLength ?? UnknownLength``` so the content length reflects the headers' value.

ggnaegi (changes requested):
Ok, you are right. We might need to add Content null for methods like HEAD or TRACE. But, I think you are going too far with a custom Http Content class for empty objects. I have written a lighter version:
```csharp
if (request.Body == null || (!request.ContentLength.HasValue && StringValues.IsNullOrEmpty(request.Headers.TransferEncoding)))
{
    return null;
}

HttpContent content = request.ContentLength is 0
    ? new ByteArrayContent([])
    : new StreamHttpContent(request.HttpContext);

AddContentHeaders(request, content);

return content;
```
When creating the HttpContent object, we are setting the headers, this include the Content-Length header. So I would prefer using Headers.ContentLength instead of introducing a variable in HttpContent.

ggnaegi:
@raman-m After a nightly review, I think we should push these changes in the January 24 Release.

alexreinert:
I used the EmptyHttpContent for two reasons: First is performance as both methods needs not to check the length of the content. Second, and important, the ByteArrayContent will add ghe Content-Length Header inside the constructor and later you add the header a second time.

src/Ocelot/Request/Mapper/StreamHttpContent.cs
alexreinert: If the headers do not contains the Content-Length header, when it will try to Calculate the content length using TryComputeLength. And on chunked content, there is no Content-Length header set, so it will result in a stack overflow.

ggnaegi:
> I used the EmptyHttpContent for two reasons: First is performance as both methods needs not to check the length of the content. Second, and important, the ByteArrayContent will add ghe Content-Length Header inside the constructor and later you add the header a second time.

No, the ByteArrayContent shouldn't add the Content-Length Header. You need to set the value explicitely.

Don't understand me wrong, your Empty Content class is probably better, but it's a custom class, whereas ByteArrayContent is maintained by Microsoft.

alexreinert:
I added tests for streaming data. The idea of the tests is just to use that much data (100GB), that it would fail because of missing memory if it is not streamed.
I also did all changes requested by the review, except the EmptyContent and the use of a custom _content_length property, where I added comments above where needs to be decided, how to proceed.

src/Ocelot/Request/Mapper/StreamHttpContent.cs
ggnaegi: No, just use ```length = Headers.ContentLength ?? UnknownLength``` Headers.ContentLength is a long?, so it is per default null, even if the header is not set.

ggnaegi: I prefer this approach, so we are relying again on Microsoft implementations.

alexreinert:
> > I used the EmptyHttpContent for two reasons: First is performance as both methods needs not to check the length of the content. Second, and important, the ByteArrayContent will add ghe Content-Length Header inside the constructor and later you add the header a second time.
> 
> No, the ByteArrayContent shouldn't add the Content-Length Header. You need to set the value explicitely.
> 
> Don't understand me wrong, your Empty Content class is probably better, but it's a custom class, whereas ByteArrayContent is maintained by Microsoft.

Ok, I changed it.

raman-m:
@ggnaegi [commented](https://github.com/ThreeMammals/Ocelot/pull/1972#issuecomment-1963135777):
>After a nightly review, I think we should push these changes in the January 24 Release. I have overseen scenarios that were covered by the ByteArrayContent implementation.

:ok: Added!

raman-m:
@alexreinert [commented](https://github.com/ThreeMammals/Ocelot/pull/1971#issuecomment-1962422167)
>It is not just the iis. A GET request should not have content, but the Transfer-Encoding header indicates, that there is content (even if it is just zero bytes). For a DELETE request there must be no content, but the issue will send the Transfer-Encoding header there, too. So this could be also an issue with some WAF, which enforce the RFC rules.

----
> GET request should not have content

We had the same debates 2 months ago. Our decision and compromise: don't change anything in request body. Ocelot should route the body as-is!

-----
> So this could be also an issue with some WAF, which enforce the RFC rules.

Let us know how does Ocelot changes these rules, or doesn't follow?
I believe you will see the same picture when connecting to your downstream service in the same IIS environment but without Ocelot. Or, am I wrong?

raman-m:
@alexreinert Do you have deployed Ocelot in any Production environments?

src/Ocelot/Request/Mapper/StreamHttpContent.cs
ggnaegi: If you sure about this, then you can open a ticket by microsoft.

raman-m: ## :rofl:

raman-m: @ggnaegi Gui, are you tired of this long discussions?
I care about your mood... :wink:

ggnaegi: my mood swings

alexreinert: If I look at the comment in the linked method, I would say "It's not a bug, it's a feature".

alexreinert:
> @alexreinert Do you have deployed Ocelot in any Production environments?

Yes

alexreinert:
> @alexreinert [commented](https://github.com/ThreeMammals/Ocelot/pull/1971#issuecomment-1962422167):
> 
> > It is not just the iis. A GET request should not have content, but the Transfer-Encoding header indicates, that there is content (even if it is just zero bytes). For a DELETE request there must be no content, but the issue will send the Transfer-Encoding header there, too. So this could be also an issue with some WAF, which enforce the RFC rules.
> 
> > GET request should not have content
> 
> We had the same debates 2 months ago. Our decision and compromise: don't change anything in request body. Ocelot should route the body as-is!

Same opinion on my side, it should be same. But with version 23 it is not the same. Neither if you send a request without content, nor when you send content with Content-Length Header, which will be "converted" to chunked encoding.

> > So this could be also an issue with some WAF, which enforce the RFC rules.
> 
> Let us know how does Ocelot changes these rules, or doesn't follow? I believe you will see the same picture when connecting to your downstream service in the same IIS environment but without Ocelot. Or, am I wrong?

Without Ocelot it works, with Ocelot 22 it works (but the request is changed and gets a Content-Lenght header), with Ocelot 23 it breaks because of the Transfer-Encoding Header Ocelot (or better say the non-null HttpContent in the HttpWebRequest) appends.

src/Ocelot/Request/Mapper/StreamHttpContent.cs
ggnaegi: Yeah well, Microsoft is using this approach with Headers.ContentLength in YARP, so that's why I'm a bit surprised.

ggnaegi: > I would say it is pretty uncommon, that the content has to "ask" the Header how long it is, the content should know that without the header. So even if MS will change that, it will be not in the near future.

No, that's your point of view, at the end, it's a content header, so it's per se part of the content

ggnaegi:
@alexreinert I will review the code in an hour or two. From your side could you deploy this version on an environment and give us a feedback? Thanks.

alexreinert:
> @alexreinert I will review the code in an hour or two. From your side could you deploy this version on an environment and give us a feedback? Thanks.

I deployed it in our staging environment and I see no issues.

ggnaegi (approved):
@alexreinert Maybe reduce the size to 1GB, I'm fine with the rest. We'll probably put the changes on some environments for a few days to validate them. Thanks, and you're welcome to contribute, we've got quite a few open issues. We're looking for someone who would like to propose a PR for SSE support. The new Release will be rolled out next week probably @raman-m  yes?

ggnaegi:
@RaynaldM Maybe you could put the changes from this PR on a staging environment? Thanks!

raman-m:
> [ggnaegi](https://github.com/ThreeMammals/Ocelot/pull/1972#pullrequestreview-1900898962): The new Release will be rolled out next week probably @raman-m yes?
> [ggnaegi](https://github.com/ThreeMammals/Ocelot/pull/1972#issuecomment-1964169068): @RaynaldM Maybe you could put the changes from this PR on a staging environment? Thanks!

Well... Guys, I'm not sure we are ready to deliver this fix right in the current release...
Time is marching! Jan'24 is small monthly release, and I cannot wait for a months! I want to release till Friday, March 1.

Please, keep me updated about the testing in your Production env please!

@ggnaegi Can I start final review? I guess PR is ready for final code review...

ggnaegi:
> > [ggnaegi](https://github.com/ThreeMammals/Ocelot/pull/1972#pullrequestreview-1900898962): The new Release will be rolled out next week probably @raman-m yes?
> > [ggnaegi](https://github.com/ThreeMammals/Ocelot/pull/1972#issuecomment-1964169068): @RaynaldM Maybe you could put the changes from this PR on a staging environment? Thanks!
> 
> Well... Guys, I'm not sure we are ready to deliver this fix right in the current release...
> Time is marching! Jan'24 is small monthly release, and I cannot wait for a months! I want to release till Friday, March 1.
> 
> **Please, keep me updated about the testing in your Production env please!**
> 
> @ggnaegi Can I start final review? I guess PR is ready for final code review...

@raman-m yes, code is ready for final review.

src/Ocelot/Request/Mapper/RequestMapper.cs
```diff
@@ -1,87 +1,91 @@
...
+            Content = MapContent(request),
+            Method = MapMethod(request, downstreamRoute),
+            RequestUri = MapUri(request),
+            Version = downstreamRoute.DownstreamHttpVersion,
+        };
+
+        MapHeaders(request, requestMessage);
+
+        return requestMessage;
+    }
```
raman-m, line 26: @alexreinert  Herr Reinert,
Why do I see this strange diff which is fake for sure?
Line endings? Which IDE do you use? Visual Studio, ReSharper, Visual Code or something else?
**You should not change line endings at all!**

src/Ocelot/Request/Mapper/RequestMapper.cs
```diff
@@ -1,87 +1,95 @@
...
+
+    private static HttpContent MapContent(HttpRequest request)
+    {
+        HttpContent content;
+
+        // no content if we have no body or if the request has no content according to RFC 2616 section 4.3
+        if (request.Body == null
```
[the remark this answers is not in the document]
raman-m: @ggnaegi Gui, is it resolved?

src/Ocelot/Request/Mapper/StreamHttpContent.cs
raman-m: > @ggnaegi: Yeah well, Microsoft is using this approach with Headers.ContentLength in YARP, so that's why I'm a bit surprised.

Alexander's concern: the code `Headers.ContentLength` leads to stack overflow. First of all, did someone caught this exception?
My recommendation: Moving `Headers.ContentLength` out of the method to the very beginning of the `Invoke`. So, can we move headers reading to upper contexts, 1 or even 2 levels up?
Will it help to avoid stack overflow error?

test/Ocelot.AcceptanceTests/MapRequestTests.cs
```diff
@@ -0,0 +1,245 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Ocelot.Configuration.File;
+
+namespace Ocelot.AcceptanceTests
```
raman-m, line 9: File-scoped declaration should be!

test/Ocelot.AcceptanceTests/MapRequestTests.cs
```diff
@@ -0,0 +1,245 @@
...
+                            {
+                                Host = "localhost",
+                                Port = port,
+                            },
+                        },
+                        UpstreamPathTemplate = "/",
+                        UpstreamHttpMethod = new List<string> { "Get" },
+                    },
+                },
+            };
```
raman-m, line 72: Why not to define setup helpers?
- `GivenRoute()`
- `GivenConfiguration()`

These helpers will allow you to reduce lines number in the file in times!

raman-m (commented):
It is hard to review the design here, as for me.
But my suggestions are below! 👇 

Question: Do we support [Chunked Encoding](https://ocelot.readthedocs.io/en/latest/introduction/notsupported.html#chunked-encoding) now? If Yes, we have to update docs in the "[Not Supported](https://ocelot.readthedocs.io/en/latest/introduction/notsupported.html)" chapter.
I see acceptance tests in `MapRequestTests` class which seems check exactly the case of Chunked Encoding...

alexreinert:
> Question: Do we support [Chunked Encoding](https://ocelot.readthedocs.io/en/latest/introduction/notsupported.html#chunked-encoding) now? If Yes, we have to update docs in the "[Not Supported](https://ocelot.readthedocs.io/en/latest/introduction/notsupported.html)" chapter. I see acceptance tests in `MapRequestTests` class which seems check exactly the case of Chunked Encoding...

Fixed length content and chunked encoding is supported, depending how the request by the client looks like. And for both ways streaming the content data works.

src/Ocelot/Request/Mapper/StreamHttpContent.cs
raman-m: >@alexreinert [2 days ago](https://github.com/ThreeMammals/Ocelot/pull/1972#discussion_r1502251125)

We are very sorry, but you've provided [the link](https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/System/net/System/Net/Http/Headers/HttpContentHeaders.cs#L76) to the repo with the code 5 years old!!! 🤣 
image
Why we should take it into account? 😉 
We need fresh .NET Core source!

alexreinert:
Feel free to change the PR, I will quit here. I have not the time to push it any further.

src/Ocelot/Request/Mapper/StreamHttpContent.cs
raman-m: Valid link is: https://github.com/dotnet/runtime/blob/81f0e3436f17dcd4efc4c2e04e501cf8dbcff650/src/libraries/System.Net.Http/src/System/Net/Http/Headers/HttpContentHeaders.cs#L42

But seems the code is similar.

raman-m:
@alexreinert 
> Feel free to change the PR, I will quit here.

Got it! Thanks for this PR! We arrange PR delivery ourselves.
We expect other interesting PRs from you! 😉

src/Ocelot/Request/Mapper/RequestMapper.cs
raman-m: No answer!... :angry:

src/Ocelot/Request/Mapper/RequestMapper.cs
```diff
@@ -1,87 +1,91 @@
...
+            || (!request.ContentLength.HasValue && StringValues.IsNullOrEmpty(request.Headers.TransferEncoding)))
```
raman-m, line 34: Old version
```diff
-        // TODO We should check if we really need to call HttpRequest.Body.Length
-        // But we assume that if CanSeek is true, the length is calculated without an important overhead
-        if (request.Body is null or { CanSeek: true, Length: <= 0 })
+        // no content if we have no body or if the request has no content according to RFC 2616 section 4.3
+        if (request.Body == null
+            || (!request.ContentLength.HasValue && StringValues.IsNullOrEmpty(request.Headers.TransferEncoding)))

```
New version

src/Ocelot/Request/Mapper/RequestMapper.cs
```diff
@@ -1,87 +1,91 @@
...
+        {
+            return null;
+        }
+
+        content = request.ContentLength is 0
+            ? new ByteArrayContent(Array.Empty<byte>()) 
+            : new StreamHttpContent(request.HttpContext);
```
raman-m, line 41: Old version
```diff
-        var content = new StreamHttpContent(request.HttpContext);
+        content = request.ContentLength is 0
+            ? new ByteArrayContent(Array.Empty<byte>()) 
+            : new StreamHttpContent(request.HttpContext);
```
New version

src/Ocelot/Request/Mapper/StreamHttpContent.cs
raman-m: And...

src/Ocelot/Request/Mapper/StreamHttpContent.cs
```diff
@@ -8,25 +8,24 @@ public class StreamHttpContent : HttpContent
...
         _context = context ?? throw new ArgumentNullException(nameof(context));
+        _contentLength = context.Request.ContentLength ?? UnknownLength;
     }
 
-    protected override async Task SerializeToStreamAsync(Stream stream, TransportContext context,
-        CancellationToken cancellationToken)
-        => await CopyAsync(_context.Request.Body, stream, Headers.ContentLength ?? UnknownLength, false,
-            cancellationToken);
+    protected override async Task SerializeToStreamAsync(Stream stream, TransportContext context, CancellationToken cancellationToken)
+        => await CopyAsync(_context.Request.Body, stream, _contentLength, false, cancellationToken);
```
raman-m, line 20: Note 3
No need to **await**!

src/Ocelot/Request/Mapper/StreamHttpContent.cs
```diff
@@ -8,25 +8,24 @@ public class StreamHttpContent : HttpContent
...
 
     protected override async Task SerializeToStreamAsync(Stream stream, TransportContext context)
-        => await CopyAsync(_context.Request.Body, stream, Headers.ContentLength ?? UnknownLength, false,
-            CancellationToken.None);
+        => await CopyAsync(_context.Request.Body, stream, _contentLength, false, CancellationToken.None);
```
raman-m, line 23: `await` can be removed!

src/Ocelot/Request/Mapper/StreamHttpContent.cs
raman-m: Did we finish debates here?

test/Ocelot.AcceptanceTests/MapRequestTests.cs
raman-m, line 1: Note 4
Move to a new `Request` folder?

test/Ocelot.AcceptanceTests/MapRequestTests.cs
```diff
@@ -0,0 +1,245 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Ocelot.Configuration.File;
+
+namespace Ocelot.AcceptanceTests
+{
+    public class MapRequestTests : IDisposable
```
raman-m, line 11: Cannot get it! What class do we test here? `RequestMapper`?
Better to rename to `ChunkedEncodingTests` ❗

test/Ocelot.AcceptanceTests/ServiceHandler.cs
```diff
@@ -46,6 +46,27 @@ public void GivenThereIsAServiceRunningOn(string baseUrl, string basePath, Reque
             _builder.Start();
         }
 
+        public void GivenThereIsAServiceRunningOn(string baseUrl, string basePath, long maxBodySize, RequestDelegate del)
+        {
+            _builder = new WebHostBuilder()
+                .UseUrls(baseUrl)
+                .UseKestrel()
+                .ConfigureKestrel(options =>
```
raman-m, line 54: Seems we have another service handlers which have Kestrel options.
So we need to merge such helpers, otherwise this copied-pasted code will never end.
So, better to have delegate argument than passing a set og them!

test/Ocelot.AcceptanceTests/StreamContentTest.cs
```diff
@@ -0,0 +1,149 @@
+﻿using System.Security.Cryptography;
+using Microsoft.AspNetCore.Http;
+using Ocelot.Configuration.File;
+
+namespace Ocelot.AcceptanceTests;
+
+public class StreamContentTest : IDisposable
```
raman-m, line 7:
```suggestion
public class StreamContentTests : IDisposable
```
and rename the file!

test/Ocelot.AcceptanceTests/StreamContentTest.cs
```diff
@@ -0,0 +1,149 @@
...
+{
+    private readonly ServiceHandler _serviceHandler;
+    private readonly Steps _steps;
```
raman-m, line 10: `Steps` should be inherited in favor of aggregation!

test/Ocelot.UnitTests/Request/Mapper/RequestMapperTests.cs
```diff
@@ -113,11 +114,51 @@ public void Should_handle_no_headers()
     public void Should_map_content()
     {
         this.Given(_ => GivenTheInputRequestHasContent("This is my content"))
-            .And(_ => GivenTheInputRequestHasMethod("GET"))
+            .And(_ => GivenTheInputRequestHasMethod("POST"))
```
raman-m, line 117: Attention!
`GET` was replaced with `POST`!!!
Better to convert to `Theory`

raman-m (changes requested):
Code review Round 2

test/Ocelot.AcceptanceTests/MapRequestTests.cs
raman-m: Fixed in https://github.com/alexreinert/Ocelot/commit/53e49d1ca701e8185e5f26a460f3892406470064

test/Ocelot.AcceptanceTests/MapRequestTests.cs
raman-m: Fixed in https://github.com/alexreinert/Ocelot/commit/d077a01cd71d151c9f6c79295e1c4a07479b1760

test/Ocelot.UnitTests/Request/Mapper/RequestMapperTests.cs
raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/1972/commits/dcbbab7119180345afaf001fbf3664f64b2fafd7 ✔️

test/Ocelot.AcceptanceTests/StreamContentTest.cs
raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/1972/commits/dcbbab7119180345afaf001fbf3664f64b2fafd7 ✔️

test/Ocelot.AcceptanceTests/ServiceHandler.cs
raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/1972/commits/dcbbab7119180345afaf001fbf3664f64b2fafd7 ✔️

test/Ocelot.AcceptanceTests/MapRequestTests.cs
raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/1972/commits/dcbbab7119180345afaf001fbf3664f64b2fafd7 ✔️

src/Ocelot/Request/Mapper/StreamHttpContent.cs
raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/1972/commits/dcbbab7119180345afaf001fbf3664f64b2fafd7 ✔️

src/Ocelot/Request/Mapper/StreamHttpContent.cs
raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/1972/commits/dcbbab7119180345afaf001fbf3664f64b2fafd7 ✔️

test/Ocelot.AcceptanceTests/MapRequestTests.cs
raman-m: Moved in commit https://github.com/ThreeMammals/Ocelot/pull/1972/commits/dcbbab7119180345afaf001fbf3664f64b2fafd7 ✔️

raman-m (approved):
Ready for delivery! ✔️ 

@ggnaegi  But the 2nd follow-up PR is required with your design review, please!
Need your approval!

raman-m:
@ggnaegi  I didn't link #928 in this PR because we have to check everything in the next your PR... So, we will link #928 to your PR only, to close officially.

ggnaegi (approved):
@raman-m Ok for me, as discussed, I will create a new PR after that.

ggnaegi:
@raman-m @alexreinert Here was the funny bit:
[GetComputedOrBufferLength](https://github.com/dotnet/runtime/blob/50d6e5d5ffd05dd4034cffd222ea610baedcc326/src/libraries/System.Net.Http/src/System/Net/Http/HttpContent.cs#L593)

PR #1655, opened 2023-05-21
#1351 #1510 New AddOcelot extensions to support custom ASP.NET MVC pipeline building via the IMvcCoreBuilder interface

guoyongchang:
## Closes #1351 #1510 
- #1351 
- #1510 

## Background
The default `AddOcelot` method adds `.AddNewtonsoftJson()`, which was necessary when Microsoft did not launch the `System.Text.Json` library, but now it affects normal use, so this PR is mainly to solve the problem problem.
> 默认的`AddOcelot`方法内部添加了`.AddNewtonsoftJson()`，在微软没有推出`System.Text.Json`库的时候是有必要的，但是现在反而影响了正常使用，所以本PR主要是解决该问题。

## New Feature
Added the following methods in `Ocelot.DependencyInjection.ServiceCollectionExtensions`
> 在`Ocelot.DependencyInjection.ServiceCollectionExtensions`添加了以下方法
- `AddOcelotWithCustomMvcCoreBuilder(this IServiceCollection services, Func<IMvcCoreBuilder, Assembly, IMvcCoreBuilder> customMvcCoreBuilder)`
- `AddOcelotWithCustomMvcCoreBuilder(this IServiceCollection services, IConfiguration configuration, Func<IMvcCoreBuilder, Assembly, IMvcCoreBuilder> customMvcCoreBuilder)`

## Proposed Changes
  - Support custom `MvcCoreBuilder` to adapt to more changes in the future, this change is mainly to support `System.Text.Json`
    > 支持自定义`MvcCoreBuilder`，以适应未来更多的变化，本次更改主要是为了支持`System.Text.Json`

PR #1986, opened 2024-03-03
#1518 Create building the `IServiceCollection`

### LINKED ISSUE #1518: Dependency Injection extensions build the service provider
When using the default DI extension to register the Ocelot service, the library builds the service provider.
This causes problems with the dependency injection library.

- [Dependency injection in ASP.NET Core | Microsoft Learn](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-5.0#recommendations)

Source : https://github.com/ThreeMammals/Ocelot/blob/bb795876c493af958f1d58c852ea56e94263906d/src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs#L20

ArwynFr:
## Fixes #1518 
- #1518 

## Proposed Changes
  - Replace the statement building the `ServiceCollection` by a new `ServiceProvider` by copying the `ServiceDescriptor`

Beware that this does not fix the problem of **Steeltoe** discovery client that might still build the service provider.
This is just a hotfix, there is still room for a larger refactor of removing the `IConfiguration` from the `OcelotBuilder`.


raman-m:
Hi Adrien!
Thank you for created PR!

I've added PR to [March'24](https://github.com/ThreeMammals/Ocelot/milestone/6) milestone. So, it will be released during first days of April. No objections?
Or should I prioritize it adding to the current [February'24](https://github.com/ThreeMammals/Ocelot/milestone/5) release?
But I guess we have to have some time for debates, discussions and writing tests.

ArwynFr:
Hey Raman,

As said on the issue, I'm not involved anymore in development of Ocelot-based software.
So I personally have no rush in the publication of this PR.

It is a mere hotfix regarding any aspnet 6+ compatibility, to prevent some apps to crash.
There is still a long way to go in adopting WebApplicationBuilder and options pattern.

raman-m:
@ArwynFr 
Could you explain me please?
What kind of app crashing do you fix by this tiny PR?

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -47,7 +50,9 @@ public static IOcelotBuilder AddOcelot(this IServiceCollection services, IConfig
     /// <returns>An <see cref="IOcelotBuilder"/> object.</returns>
     public static IOcelotBuilder AddOcelotUsingBuilder(this IServiceCollection services, Func<IMvcCoreBuilder, Assembly, IMvcCoreBuilder> customBuilder)
     {
-        var configuration = services.BuildServiceProvider().GetRequiredService<IConfiguration>();
+        var descriptor = services.Where(descriptor => descriptor.ServiceType == typeof(IConfiguration)).FirstOrDefault();
+        var provider = new ServiceCollection().Add(descriptor).BuildServiceProvider();
+        var configuration = provider.GetRequiredService<IConfiguration>();
```
raman-m, line 55: Why not to define a private helper to search for `IConfiguration` object?
Also, the XML description need to point out to the developer that the interface object is derived from DI and must be registered in DI container anyway!

raman-m: This is my fault when I reviewed #1655  and didn't pay my attention to the fact we add overloaded version without the interface object 👇 
https://github.com/ThreeMammals/Ocelot/blob/5a3c55e358afb7853819e3370daad9bb15452dd9/src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs#L48-L52
So, 😮  But Ocelot Builder has required 2nd argument:
https://github.com/ThreeMammals/Ocelot/blob/5a3c55e358afb7853819e3370daad9bb15452dd9/src/Ocelot/DependencyInjection/OcelotBuilder.cs#L51

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -17,15 +18,17 @@ public static class ServiceCollectionExtensions
     /// <returns>An <see cref="IOcelotBuilder"/> object.</returns>
     public static IOcelotBuilder AddOcelot(this IServiceCollection services)
     {
-        var configuration = services.BuildServiceProvider().GetRequiredService<IConfiguration>();
+        var descriptor = services.Where(descriptor => descriptor.ServiceType == typeof(IConfiguration)).FirstOrDefault();
```
raman-m, line 21: Not an issue, but my questions:
- How many `IConfiguration` objects can be registered in DI? If more than one instance, how to detect exactly `ocelot.json` one?
- Could not the `IConfiguration` be registered at all? Seems not, because of `appsettings.json`...

raman-m (changes requested):
Thanks for English text typo corrections in XML docs! 😄

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
ArwynFr: `IConfiguration` is a singleton. When adding multiple configuration sources, these are mixed together as a single configuration context, where the latest source overrides former ones in case of configuration key duplication.

Technically you could provoke an error if you provide a `new ServiceCollection()`, but these methods are designed to be used with `WebApplicationBuilder` which does formally creates an IConfiguration instance.

Perhpas this could cause problems with slimmer versions of the builder introduced in dotnet 8.

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
ArwynFr: The fact that `OcelotBuilder` requires an `IConfiguration` is actually the problem in the first place.
This builder should only focus on registering service descriptors with their lifetime managers.
Services would access configuration by injection of `IOptionsMonitor<T>` rather than an IConfiguration reference.
Overloads with IConfiguration parameters would then be the ones to deprecate.

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
ArwynFr: By flattening multiple sources in a single IConfiguration context, the framework allows to easily implement "configuration overriding" features. For instance the default WebApplicationBuilder registers appsettings.json, then user secret file, and then environment variables as configuration sources. This allows secret files to override appsettings.json and environment variables to override both.

ArwynFr:
> @ArwynFr Could you explain me please? What kind of app crashing do you fix by this tiny PR?

See discussion in #1518

Applications built using a dotnet 6+ with WebApplicationBuilder instead of Host/Startup, and using AddOcelot() override without the IConfiguration parameter will likely fail to start or crash. This is because the method builds the service provider which is expected not to happen with AddXXX methods. This in turns puts the provider in a weird state where resolving services could fail, throw an uncaught exception and provoke failure to start.

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
ArwynFr: Also using Builder without an IConfiguration allows to register configuration sources AFTER the registration of the services. This in turns allow to do some fancy stuff like configuration bootstraping (read configuration from file, create services to access a remote storage using this config, then provide the rest of the app with config values from that storage).

Which is exacly the kind of problems Ocelot had tried to solve with Consul/Eureka.

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
ArwynFr: I wasn't sure deduplication of 3 lines of code was worth writing a separate helper.
Also in my mind this code is expected to go away once the IConfiguration references are gone.

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
ArwynFr: If you want to isloate `ocelot.json`, you need to have a separate IConfiguration instance, using a separate ConfigurationBuilder for instance and not have it injected in the DI container. This is NOT what currently happens and NEITHER is what is described in the documentation.

Currently all configuration (ocelot routes and aspnet config) are mixed together in a single IConfiguration

ggnaegi (approved):
Ok, we should keep in mind it's a quick fix, but yes, it will work.  As of the issue with the Eureka Service Discovery, it's probably related to the Steeltoe extension methods: https://github.com/ggnaegi/Ocelot/blob/42ac3ca51829006bbe2e63bc684ab3a3caaccf1e/src/Ocelot.Provider.Eureka/OcelotBuilderExtensions.cs#L12
As we can see in their latest commits here:
https://github.com/SteeltoeOSS/Steeltoe/commit/e9343558c8d089ef794f456bdd7598e764495fc1
@ArwynFr Maybe it would make sense that you include the package update (3.6) in your PR, do you agree?

ArwynFr:
@ggnaegi Steeltoe is facing the same problem as Ocelot, but they have chosen to enforce the `IConfiguration`, pushing the problem towards the developer instead of solving it. Since we are already passing the configuration instance, upgrading the library won't change much. But it won't help with removing the OcelotBuilder.Configuration neither.

ggnaegi:
> @ggnaegi Steeltoe is facing the same problem as Ocelot, but they have chosen to enforce the `IConfiguration`, pushing the problem towards the developer instead of solving it. Since we are already passing the configuration instance, upgrading the library won't change much. But it won't help with removing the OcelotBuilder.Configuration neither.

@ArwynFr Yes, but they were, at least until version 3.5 using the same
```c#
var configuration = services.BuildServiceProvider().GetRequiredService<IConfiguration>();
```
as ocelot did. 
Since we know that, why would we keep something potentially hazardous in Ocelot? Yes, IConfiguration shouldn't be null, but we might have some edge cases. It was only some assumptions. We could create a new PR for the library update too.

ArwynFr:
> We could create a new PR for the library update too.

Since it's my first PR to Ocelot, I'd rather have the lib dependency update separated please.

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
ArwynFr: @raman-m I've added some default behavior in case the IConfiguration wasn't registered.
Note that Tom's code reads the configuration from current working path, which may be different from the executable's path.

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -70,4 +69,22 @@ public static IOcelotBuilder AddOcelotUsingBuilder(this IServiceCollection servi
...
     }
+
+    private static IConfiguration MakeDefaultConfiguration(IWebHostEnvironment env)
+    {
+        return ConfigurationBuilderExtensions.AddOcelot(new ConfigurationBuilder(), env).Build();
+    }
+
+    private static IConfiguration FindConfiguration(this IServiceCollection services, IWebHostEnvironment env)
+    {
+        var descriptor = services.Where(descriptor => descriptor.ServiceType == typeof(IConfiguration)).FirstOrDefault();
```
ggnaegi, line 80: maybe 
```c# 
services.FirstOrDefault(descriptor => descriptor.ServiceType == typeof(IConfiguration));
```

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -70,4 +69,22 @@ public static IOcelotBuilder AddOcelotUsingBuilder(this IServiceCollection servi
     {
         return new OcelotBuilder(services, configuration, customBuilder);
     }
+
+    private static IConfiguration MakeDefaultConfiguration(IWebHostEnvironment env)
+    {
+        return ConfigurationBuilderExtensions.AddOcelot(new ConfigurationBuilder(), env).Build();
```
ggnaegi, line 75: maybe
```c#
return new ConfigurationBuilder().AddOcelot(env).Build();
```

ggnaegi (changes requested):

RaynaldM (approved):

ggnaegi (approved):
Ok, thanks @ArwynFr !

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -17,15 +20,15 @@ public static class ServiceCollectionExtensions
     /// <returns>An <see cref="IOcelotBuilder"/> object.</returns>
     public static IOcelotBuilder AddOcelot(this IServiceCollection services)
     {
-        var configuration = services.BuildServiceProvider().GetRequiredService<IConfiguration>();
+        var configuration = services.FindConfiguration(null);
         return new OcelotBuilder(services, configuration);
     }
```
raman-m, line 25: `[Obsolete]` ?

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -47,7 +51,7 @@ public static IOcelotBuilder AddOcelot(this IServiceCollection services, IConfig
     /// <returns>An <see cref="IOcelotBuilder"/> object.</returns>
     public static IOcelotBuilder AddOcelotUsingBuilder(this IServiceCollection services, Func<IMvcCoreBuilder, Assembly, IMvcCoreBuilder> customBuilder)
     {
-        var configuration = services.BuildServiceProvider().GetRequiredService<IConfiguration>();
+        var configuration = services.FindConfiguration(null);
         return new OcelotBuilder(services, configuration, customBuilder);
     }
```
raman-m, line 56: My suggestion is removing this version at all because of absence of `IConfiguration configuration` parameter.
It will be breaking change for sure, but we will mention it in Release Notes.

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -65,4 +69,22 @@ public static IOcelotBuilder AddOcelotUsingBuilder(this IServiceCollection servi
     {
         return new OcelotBuilder(services, configuration, customBuilder);
     }
```
raman-m, line 71:
```c#
configuration ??= services.FindConfiguration(null);
return new OcelotBuilder(services, configuration, customBuilder);
```

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -65,4 +69,22 @@ public static IOcelotBuilder AddOcelotUsingBuilder(this IServiceCollection servi
...
+
+    private static IConfiguration MakeDefaultConfiguration(IWebHostEnvironment env)
+    {
+        return new ConfigurationBuilder().AddOcelot(env).Build();
+    }
```
raman-m, line 76:
```suggestion
    private static IConfiguration MakeDefaultConfiguration(IWebHostEnvironment env)
        => new ConfigurationBuilder().AddOcelot(env).Build();
```

raman-m (changes requested):
My suggestions below. And I worry about [this](https://github.com/ThreeMammals/Ocelot/issues/1518#issuecomment-2003874094)

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
ggnaegi: @ArwynFr @raman-m Mhh, well, the point of @ArwynFr was, using IConfiguration should be regarded as obsolete, not the opposite. We would then inject the settings using the options pattern, eg. 
```c#
builder.Services.Configure<PositionOptions>(
    builder.Configuration.GetSection(PositionOptions.Position));
```
... And then DI
```c#
public Test2Model(IOptions<PositionOptions> options)
```
https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-8.0

ggnaegi: Now, the problem is, the Ocelot Pipeline isn't supporting the options pattern yet. So, what we are doing here is a bit artificial, but it works.

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
ArwynFr: In my opinion, this method can be removed/obsoleted but it has more to do with the custom mvc builder.

ggnaegi: obsolete for me, not removed... In a further step remove

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
raman-m: 🆗 Guys, if you don't want to put `[Obsolete]` attribute, it is OK.  

> If we deprecate then undeprecate it, users will be confused.

OK! Got it! Not an issue!

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
raman-m: 🆗 Let's make it Obsolete.

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
raman-m: @ArwynFr  Do you have time today to fix these 3 minor issues? 😉 
Or, if no objections then I will make one commit today.

ArwynFr: Sorry didn't see your message, please help yourself I won't be available until tomorow

raman-m: Committed as https://github.com/ThreeMammals/Ocelot/pull/1986/commits/715a3641920d02332cc22628ce3d9a3040b645dc

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
raman-m: Fixed in https://github.com/ThreeMammals/Ocelot/pull/1986/commits/293c1d9d5b5bd30b7ebcfadfc50c71b69487118b ✔️

src/Ocelot/DependencyInjection/ServiceCollectionExtensions.cs
raman-m: Fixed in https://github.com/ThreeMammals/Ocelot/pull/1986/commits/293c1d9d5b5bd30b7ebcfadfc50c71b69487118b ✔️

ArwynFr:
@raman-m will look into it this weekend

raman-m:
### Coverage 100% &rarr;
- [Ocelot.DependencyInjection.ServiceCollectionExtensions - Coverage Report.zip](https://github.com/ThreeMammals/Ocelot/files/14671274/Ocelot.DependencyInjection.ServiceCollectionExtensions.-.Coverage.Report.zip)

### Report header
image

### Private methods: coverage is good
image

I like these pictures! And you, Adrien? 😉

raman-m (approved):
### Ready for delivery! ✅ 
- Code review ✔️
- Team approvals ✔️
- Unit testing ✔️ 
- Acceptance tests ❔ It would be crazy to try to emulate a basic application crash scenario by trying to juggle startup configurations 🤣 
- Updated docs ❔ _not required for system updates_ 🆗 

@ArwynFr 
Because this PR is like a hotfix: it contains pretty system updates then I believe we can go without docs updating.

### My input
- Defined new `ServiceCollectionExtensionsTests` class which contains exactly unit testing for private methods and public too
- Added unit tests to `OcelotBuilderTests` with new functionality coverage
- Switched off annoying reminder about `Obsolete` definitions aka **CS0618** Warn , but for test projects only
- The ultimate goal is 100% coverage of `ServiceCollectionExtensions` class achieved.
- Fixed ugly a little duck duck bug in the `GetMergedOcelotJson` method of `ConfigurationBuilderExtensions` class. So, `null` value of the `IWebHostEnvironment env` argument led to empty string passing to the cstor in line of this code: `new FileInfo(environmentFile)`. Well... It was my design gap. 😢 And especially these unit testing challenge helped me to uncover this Duck bug 🦆 🐛
  That's why I love unit testing 😍 

Yahoo! We can merge, @ArwynFr, right? 😋 
We haven't made any breaking changes, right? 😳😁

raman-m:
@ArwynFr  Wanna press the magic button? 🪄 
I mean **Squash and merge** ! 😉

raman-m:
@ArwynFr Where are you?

raman-m:
@ArwynFr  I have no time to wait for a simple action to do. I don't want to delay release too.
Merging...

raman-m:
@ArwynFr Thank you for the contribution! 
This hotfix will be released next week...

PR #2044, opened 2024-04-12
#2039 Buffer request body and copy request body to downstreams during multiplexing

### LINKED ISSUE #2039: Body cannot be forwarded twice on Aggregator
Aggregation cannot currently forward the body of the upstream to several downstreams. This is mentioned in :

https://github.com/ThreeMammals/Ocelot/blob/7c26c073dc02ec7e430d43428ad6906da484cd50/src/Ocelot/Multiplexer/MultiplexingMiddleware.cs#L218

The error is not really explicit right now, it's a hidden message. The first downstream correctly reads the body and returns the expected result, but the second request has the following error:
``` 
Error connecting to downstream service, exception: System.Net.Http.HttpRequestException: An error occurred while sending the request.
 ---> System.InvalidOperationException: Sent 0 request content bytes, but Content-Length promised xxx.
   at Ocelot.Request.Mapper.StreamHttpContent.CopyAsync(Stream input, Stream output, Int64 announcedContentLength, Boolean autoFlush, CancellationToken cancellation) 
```

This obviously comes from the fact that the body was already read and is not.

I feel this is a reasonable request to be able to forward a body to multiple downstreams.

  1. Define two routes with requests having a body (GET requests for me)
  2. Define an aggregation of these two
  3. Call the aggregation

## Specifications
Aggregators are said not to be able to manage POST requests, but they mainly can't manage bodies. The aggregation just forwards the stream of the body to the following requests.

Still, I got it to work but this leads to major discussions and impactful choices.

## Fix
I created a basic method (which should be rewritten properly) to clone the body
```csharp
    protected static async Task<MemoryStream> CloneBodyAsync(Stream body)
    {
        var memoryStream = new MemoryStream();
        await body.CopyToAsync(memoryStream);
        body.Position = 0;
        memoryStream.Position = 0;
        return memoryStream;
    }
```
Which can be used in the `CreateThreadContextAsync` (must be set Async) that way:
```csharp
    private static async Task<HttpContext> CreateThreadContextAsync(HttpContext source)
    {
        var from = source.Request;
        var bodyCopy = await CloneBodyAsync(from.Body);
        var target = new DefaultHttpContext
        {
            Request =
            {
                Body = bodyCopy,
…
```
The fix was pretty straightforward to integrate, even when switching to Async methods because we have an asynchronous context just before.

However, it is important to note that being able to set back the position of the stream body at the beginning (`body.Position = 0;`) is not supported by default.
I had to enable buffering to be able to do this.
Example code:
```C#
app.Use(
    async (context, next) =>
    {
        context.Request.EnableBuffering(); // necessary for aggregation body copy
        await next();
    }
);
```
ASP Net also doesn't allow synchronous operations for this, which is why it is heavily recommended to switch to async methods as I did.

## Discussion
The way I fixed it locally is definitely a choice and should be carefully considered. This requires to enable buffering of the body (pretty obvious if we want to use it multiple times). That means Ocelot has to buffer the requests, which could be potentially impactful on performances.

I fixed it in a very trivial way, but we could actually also buffer manually by loading the body into a first memory stream before doing the copy for each downstream. That way we are not reliant on enabling the buffering for all the requests, simply on complex aggregations that requires it.

I feel this is the main reason the POST method is not available for aggregations, and I think this could be a major improvement for Ocelot if done right. Supporting the forwarding of a body to multiple streams is really useful feature.
I know I stumped on this limitation by doing a borderline thing using body with GET (although I'll die on the hill that the spec does not forbid it), but I think it opens a really meaningful discussion on a key feature.

Another possibility would be to set back the Stream position at 0 after the copy in the mapper when available (and leave the user the responsibility to enable buffering) but I think it's a less elegant solution. (https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/Request/Mapper/StreamHttpContent.cs)

On a side-note, to fix it locally I had to copy a lot of Ocelot code — I think most private methods I encountered could just be protected / virtual. This would help customizing the behavior of some components.

I'm more than willing to work on it if necessary, but I feel this is a key discussion to validate with the maintainers before working on it as it has major implications.

Feel free to comment on thoughts, ideas or anything, I'm completely open to criticism and discussions.

  - Version: 23.2.2
  - Platform: .NET 8
  - Subsystem: Aggregation

PaulARoy:
## Closes #2039 
- #2039 

Buffers the body of the request in some cases and copy the buffered body to the downstreams.

## Proposed Changes

  - Creates a `MemoryStream` buffer for the body when `Content-Length` > 0
  - Copy the buffer to the downstream requests through a `MemoryStream`
  - Fixes broken tests due to `CreateThreadContext` name and parameter changes + acceptance test on multiple body forwarding
  - Updates the body stream in-place

## WIP
  - Decide wether we want to implement more complex buffer logic `BufferingMiddleware.BufferBodyAsync:L87`
  - Decide wether we want to use a `MemoryStream` for downstream body `MultiplexingMiddleware.CopyBufferToTargetRequestAsync:L253`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,13 +4,13 @@ orbs:
 jobs:
   build:
     docker:
-      - image: ocelot2/circleci-build:8.21.0
+      - image: ocelot2/circleci-build:latest
     steps:
       - checkout
       - run: dotnet tool restore && dotnet cake
   release:
     docker:
-      - image: ocelot2/circleci-build:8.21.0
+      - image: ocelot2/circleci-build:latest
     steps:
       - checkout
       - run: dotnet tool restore && dotnet cake --target=Release
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -1,4 +1,31 @@
-## Documentation patch (version {0}) for [{1}](https://github.com/ThreeMammals/Ocelot/releases/tag/{1}) release
-> Read the Docs: [Ocelot 23.2](https://ocelot.readthedocs.io/en/{0}/)
+## Hotfix release (version {0}) for #2031 issue
+> Route path template placeholders and their validation rules
 
-This is a technical release: no other information.
+Special thanks to **[Guillaume Gnaegi](https://github.com/ggnaegi)** and [Fabrizio Mancin](https://github.com/Fabman08)!
+
+### About
+The bug is related to the [Placeholders](https://ocelot.readthedocs.io/en/latest/features/routing.html#placeholders) feature in [Configuration](https://ocelot.readthedocs.io/en/latest/features/configuration.html) and [Routing](https://ocelot.readthedocs.io/en/latest/features/routing.html).
+The bug was introduced in version [23.2.0](https://github.com/ThreeMammals/Ocelot/releases/tag/23.2.0) as a part of PR #1927.
+
+### Breaking Change
+The new [validation rules](https://github.com/ThreeMammals/Ocelot/blob/23.2.0/src/Ocelot/Configuration/Validator/FileConfigurationFluentValidator.cs#L45-L50) of the `FileConfigurationFluentValidator` class do not allow the Ocelot app to start when implicit [placeholders](https://ocelot.readthedocs.io/en/latest/features/routing.html#placeholders) are defined in custom implementations, such as middlewares, delegating handlers, and replaced services in the dependency injection (DI) container.
+These new rules are capable of validating explicit [placeholders](https://ocelot.readthedocs.io/en/latest/features/routing.html#placeholders) only within the `UpstreamPathTemplate` and `DownstreamPathTemplate` properties. Unfortunately, they cannot oversee implicit placeholders in custom implementations, and they do not validate early during the Ocelot app startup process.
+
+Ensure that you avoid using version [23.2.0](https://github.com/ThreeMammals/Ocelot/releases/tag/23.2.0). If you are currently on that version, upgrade to version [{0}](https://github.com/ThreeMammals/Ocelot/releases/tag/{0}) by applying this hotfix patch.
+
+### Technical info
+With version [23.2.0](https://github.com/ThreeMammals/Ocelot/releases/tag/23.2.0), particularly if you have overridden certain service classes or implemented custom logic that manipulates placeholders, you may encounter Ocelot app crashes accompanied by the following errors in the log:
+```
+One or more errors occurred. (Unable to start Ocelot, errors are: XXX)
+```
+where `XXX` are the following validation error messages:
+- `UpstreamPathTemplate 'UUU' doesn't contain the same placeholders in DownstreamPathTemplate 'DDD'`
+- `DownstreamPathTemplate 'DDD' doesn't contain the same placeholders in UpstreamPathTemplate 'UUU'`
+
+**Finally**, the [validation rules](https://github.com/ThreeMammals/Ocelot/blob/23.2.0/src/Ocelot/Configuration/Validator/FileConfigurationFluentValidator.cs#L45-L50) resulted from the incorrect assumption that placeholders are always explicit and can be validated early. Therefore, custom implementations and feature services in the dependency injection (DI) container, which rely on or manipulate placeholders, should validate the configuration JSON and appropriate options later, directly within their service implementations.
+
+### Bug Artifacts
+- Released in version: [23.2.0](https://github.com/ThreeMammals/Ocelot/releases/tag/23.2.0)
+- Introduced in: PR #1927
+- Reported bug: #2031 by @ggnaegi and tested by @Fabman08
+- Hotfix PR: #2032 by @raman-m
--- a/build.cake
+++ b/build.cake
@@ -56,7 +56,7 @@ var nugetFeedStableSymbolsUploadUrl = "https://www.nuget.org/api/v2/package";
 string committedVersion = "0.0.0-dev";
 GitVersion versioning = null;
 int releaseId = 0;
-bool IsTechnicalRelease = true;
+bool IsTechnicalRelease = false;
 string gitHubUsername = "TomPallister";
 string gitHubPassword = Environment.GetEnvironmentVariable("OCELOT_GITHUB_API_KEY");
 
@@ -84,7 +84,7 @@ Task("RunTests")
 	.IsDependentOn("RunIntegrationTests");
 
 Task("Release")
-	//.IsDependentOn("Build")
+	.IsDependentOn("Build")
 	.IsDependentOn("CreateReleaseNotes")
 	.IsDependentOn("CreateArtifacts")
 	.IsDependentOn("PublishGitHubRelease")
@@ -306,11 +306,11 @@ Task("CreateReleaseNotes")
 			}
 		} // END of Top 3
 
-		releaseNotes.Add("### Honoring :medal_sports: aka Top Contributors :clap:");
-		releaseNotes.AddRange(topContributors);
-		releaseNotes.Add("");
-		releaseNotes.Add("### Starring :star: aka Release Influencers :bowtie:");
-		releaseNotes.AddRange(starring);
+		// releaseNotes.Add("### Honoring :medal_sports: aka Top Contributors :clap:");
+		// releaseNotes.AddRange(topContributors);
+		// releaseNotes.Add("");
+		// releaseNotes.Add("### Starring :star: aka Release Influencers :bowtie:");
+		// releaseNotes.AddRange(starring);
 		releaseNotes.Add("");
 		releaseNotes.Add($"### Features in Release {releaseVersion}");
 		var commitsHistory = GitHelper($"log --no-merges --date=format:\"%A, %B %d at %H:%M\" --pretty=format:\"<sub>%h by **%aN** on %ad &rarr;</sub>%n%s\" {lastRelease}..HEAD");
@@ -423,7 +423,7 @@ Task("RunIntegrationTests")
 
 Task("CreateArtifacts")
 	.IsDependentOn("CreateReleaseNotes")
-	//.IsDependentOn("Compile")
+	.IsDependentOn("Compile")
 	.Does(() =>
 	{
 		WriteReleaseNotes();
@@ -530,7 +530,6 @@ Task("PublishToNuget")
 
 		if (IsRunningOnCircleCI())
 		{
-			Information("Publish to NuGet...");
 			PublishPackages(packagesDir, artifactsFile, nugetFeedStableKey, nugetFeedStableUploadUrl, nugetFeedStableSymbolsUploadUrl);
 		}
 	});
@@ -588,7 +587,7 @@ private void PersistVersion(string committedVersion, string newVersion)
 /// Publishes code and symbols packages to nuget feed, based on contents of artifacts file
 private void PublishPackages(ConvertableDirectoryPath packagesDir, ConvertableFilePath artifactsFile, string feedApiKey, string codeFeedUrl, string symbolFeedUrl)
 {
-		Information("PublishPackages");
+		Information("Publishing to NuGet...");
         var artifacts = System.IO.File
             .ReadAllLines(artifactsFile)
 			.Distinct();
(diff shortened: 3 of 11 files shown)


src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
```diff
@@ -46,11 +46,14 @@ public async Task Invoke(HttpContext httpContext)
             return;
         }
 
+        // WIP: generalize buffer use?
+        var bufferedBody = await BufferBodyAsync(httpContext, route, downstreamRoutes);
```
ggnaegi, line 50: I don't like this idea, why haven't you put the body copy nearer to ```CreateThreadContextAsync```. Do you want to manipulate the request body?

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
```diff
@@ -97,11 +114,12 @@ protected virtual Task ProcessSingleRouteAsync(HttpContext context, DownstreamRo
     /// Processing the downstream routes (no route keys).
     /// </summary>
     /// <param name="context">The main http context.</param>
-    /// <param name="route">The route.</param>
-    private async Task ProcessRoutesAsync(HttpContext context, Route route)
+    /// <param name="route">The route.</param>
+    /// <param name="bufferedBody">The buffered body.</param>
+    private async Task ProcessRoutesAsync(HttpContext context, Route route, Stream bufferedBody)
```
ggnaegi, line 119: Stream bufferedBody everywhere...

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
```diff
@@ -256,4 +293,37 @@ protected virtual Task MapAsync(HttpContext httpContext, Route route, List<HttpC
...
+    private static async Task CopyBufferToTargetRequestAsync(
+        DefaultHttpContext target,
+        Stream bufferedBody,
+        long? fromContentLength,
+        CancellationToken cancellationToken
+    )
+    {
+        if (fromContentLength is not null)
+        {
+            if (fromContentLength > int.MaxValue)
```
ggnaegi, line 306: This as no impact, as the request will be chunked if Buffer size > 655536 (https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/Request/Mapper/StreamHttpContent.cs#L8)

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
```diff
@@ -206,16 +240,17 @@ private static void CopyItemsToNewContext(HttpContext target, HttpContext source
...
-    private static HttpContext CreateThreadContext(HttpContext source)
+    private static async Task<HttpContext> CreateThreadContextAsync(HttpContext source, Stream bufferedBody)
     {
         var from = source.Request;
         var target = new DefaultHttpContext
         {
             Request =
             {
-                Body = from.Body, // TODO Consider stream cloning for multiple reads
+                Body = new MemoryStream(), // WIP: by default it's a NullStream, do we really use MemoryStream?
```
ggnaegi, line 253: Why not Body = await CopyBufferToTargetRequestAsync?

ggnaegi (changes requested):
Hello, Thanks, please add some acceptance tests.

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
PaulARoy: Yeah that's the drawback for a more general use.

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
PaulARoy: 👍 
Would you rather me to just remove the check on int.MaxValue?

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
PaulARoy: that was my first implementation too (defined in the bug) but I hoped the DefaultHttpContext had a already declared buffer. I'd rather rollback.

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
ggnaegi: @PaulARoy @raman-m I think we should be clear about the purpose of the multiplexer: Create the http context copies (copying bodies etc.), send the requests and then aggregate them. If you want to deal with the buffer, modify the request body etc. then we should think about it somewhere else, not here.

PaulARoy: @ggnaegi I thought I could not update the body in-place, I 100% aggree with you. The next iteration I just pushed is more fluent and makes the update in-place. It's still in the multiplexer BUT it can be easily moved to a more general approach.

Maybe a new `BufferingMiddleware`? What do you think?

PaulARoy:
> Hello, Thanks, please add some acceptance tests.

Acceptance test added for multiplexer body copy (tested — wasn't passing before modifications and passes after).

I wonder if I could add as acceptance test for buffering, taking suggestions as it's more of an internal mechanism.

PaulARoy:
All tests on Visual Studio (+ `dotnet cake`) pass on my machine and the failing test from CI (`should_reload_config_on_change`) doesn't look like something I touched. Maybe the new CI PR I saw? I have the `dotnet cake` log file if needed.

raman-m:
> All tests on Visual Studio (+ `dotnet cake`) pass on my machine and the failing test from CI (`should_reload_config_on_change`) doesn't look like something I touched. Maybe the new CI PR I saw? I have the `dotnet cake` log file if needed.

Don't worry about `should_reload_config_on_change` test! This is unstable one and related to file system intensive file operations in Docker environment.
If tess are passed locally, but on CircleCI some of them failed then ask me to trigger new build.

The build is green now.

raman-m:
@PaulARoy 
It appears that you’ve used the [PaulARoy:develop](https://github.com/PaulARoy/Ocelot/tree/develop) branch as a feature branch which is wrong approach. Here’s the recommended approach:

1. Ensure that your forked repository has an **identical develop branch** to the original repository.
2. When working on multiple tickets or features, **create a new branch from the develop branch**. This will prevent any potential blocking issues.

While it may not be critical for single-ticket work, following this practice will help you avoid complications when dealing with multiple feature branches.

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
raman-m: @PaulARoy The decision is yours! Please create a backup branch in your local repository to save your work, or stash it. Could you also remove any additional logic that falls outside the scope of the bug? I'm confident that your innovative `BufferingMiddleware`  idea will come to fruition in the subsequent PR.

src/Ocelot/Buffering/BufferingMiddleware.cs
raman-m, line 1: I'm not sure why we need this additional middleware. Could we concentrate solely on the bug within `MultiplexingMiddleware` please?
I suggest backing up the branch and temporarily removing `BufferingMiddleware`, then revisiting your concept in a subsequent PR.

src/Ocelot/Buffering/BufferingMiddlewareExtensions.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Microsoft.AspNetCore.Builder;
+
+namespace Ocelot.Buffering;
+
+public static class BufferingMiddlewareExtensions
+{
+    public static IApplicationBuilder UseBufferingMiddleware(this IApplicationBuilder builder)
```
raman-m, line 7: Defining methods that are referenced only once throughout the entire solution is not advantageous. Moreover, since this middleware will be a system-level one, it can be incorporated into the pipeline using the `UseMiddleware<T>` method. Therefore, this definition does not address DRY issues.

src/Ocelot/Buffering/Constants.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Ocelot.Buffering;
+
+public class Constants
+{
+    public const string BufferedBodyItem = "BufferedBody";
```
raman-m, line 5: Indeed, it would be more effective to define it directly within the class as a public constant.

src/Ocelot/Middleware/OcelotPipelineExtensions.cs
```diff
@@ -55,6 +56,9 @@ public static class OcelotPipelineExtensions
             // Then we get the downstream route information
             app.UseDownstreamRouteFinderMiddleware();
 
+            // Then we buffer the body if required
+            app.UseBufferingMiddleware();
```
raman-m, line 60:
```suggestion
            app.UseMiddleware<BufferingMiddleware>();
```

test/Ocelot.AcceptanceTests/AggregateTests.cs
```diff
@@ -598,6 +598,81 @@ public void Should_return_response_200_with_user_forwarding()
             }
         }
 
+        [Fact]
+        [Trait("Bug", "2039")]
+        public void Should_return_response_200_with_body_sent_on_multiple_services()
```
raman-m, line 603: Excellent! However, unit tests are also necessary.

The testing workflow should be as follows:
- Conduct unit testing.
- Proceed to acceptance testing if unit testing is successful.
- Consider benchmark testing if acceptance testing is successful. While this is optional, it is highly recommended for system redesign.

raman-m (changes requested):
@PaulARoy Great work! However, I've noticed two significant issues:
1. The new `BufferingMiddleware` appears to be unnecessary. As a developer, I'm puzzled as to why we need separate middleware instead of simply fixing the bug in the `MultiplexingMiddleware`.
2. The testing workflow has been violated and not adhered to. It would be preferable to begin with unit testing first.

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
PaulARoy: Unfortunately buffering is needed for the functionnality to work.

Explanation: if we don't buffer the body, once the stream is read it cannot be repositioned to 0. The error is pretty non-explicit.

Still if we want to just isolate the basic copy, I can remove the buffering logic and entirely rely on the fact that buffering is manually activated (either with a future custom implementation of the buffering logic, or with the AspNet capability).

The PR becomes pretty small (and I have doubts regarding testability) but that's entirely up to you.

src/Ocelot/Buffering/BufferingMiddleware.cs
PaulARoy: As I said on a previous comment, Buffering is not absolutely necessary, but repositioning the Stream at 0 cannot be done without buffering (either ourselves or with AspNet implementation).

I thought you wanted this functionnality from your comments on #2039.

src/Ocelot/Buffering/BufferingMiddlewareExtensions.cs
PaulARoy: Just following the logic of `MultiplexingMiddlewareExtensions` (which in my opinion is not necessary too).

100% OK to remove this useless class.

src/Ocelot/Buffering/Constants.cs
PaulARoy: Wasn't a big fan to reference `BufferingMiddleware.BufferedBodyItem` inside another class (as it's not really related solely to the middleware but to the whole functionnality), but not entirely conviced by adding a class with a single item. OK to move this.

test/Ocelot.AcceptanceTests/AggregateTests.cs
PaulARoy: Duly noted. Let me rework that and get back to you with a more satisfying PR.

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
raman-m: > The PR becomes pretty small (and I have doubts regarding testability) but that's entirely up to you.

Well... testing your changes is crucial. However, I can assist you with delivering the bug fix.

raman-m:
Follow up PR is #2050

PR #2050, opened 2024-04-15
#2039 Buffer request body and copy the body to downstreams during multiplexing

### LINKED ISSUE #2039: Body cannot be forwarded twice on Aggregator

PaulARoy:
## Closes #2039 
- #2039 

Buffers the body of the request when multiplexing multiple routes and copy the buffered body to the downstreams.

I took consideration of all your feedbacks from the previous PR, don't hesitate to tell me if something's off, either in the code or in the process.

I looked at the benchmark tests but I'm not sure on how to integrate this correctly.

## Proposed Changes

  - Smartly enables ASP.NET buffering for body copy
  - Copy the buffer to the downstream requests through a `MemoryStream`
  - Fixes broken tests due to `CreateThreadContext` name and parameter changes
  - Unit tests on `CloneRequestBodyAsync` (called n times when n > 1, not called for 1)
  - Acceptance test of body copy for multiple services

### DIFF
--- a/src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
+++ b/src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
@@ -215,7 +215,7 @@ public class MultiplexingMiddleware : OcelotMiddleware
         {
             Request =
             {
-                Body = await CopyBufferToTargetRequestAsync(source),
+                Body = await CloneRequestBodyAsync(source),
                 ContentLength = from.ContentLength,
                 ContentType = from.ContentType,
                 Host = from.Host,
@@ -257,7 +257,7 @@ public class MultiplexingMiddleware : OcelotMiddleware
         return aggregator.Aggregate(route, httpContext, contexts);
     }
 
-    protected virtual async Task<Stream> CopyBufferToTargetRequestAsync(HttpContext source)
+    protected virtual async Task<Stream> CloneRequestBodyAsync(HttpContext source)
     {
         source.Request.EnableBuffering();
         if (source.Request.Body.Position == 0)
--- a/src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
+++ b/src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
@@ -215,7 +215,7 @@ public class MultiplexingMiddleware : OcelotMiddleware
         {
             Request =
             {
-                Body = await CloneRequestBodyAsync(source),
+                Body = await CloneRequestBodyAsync(source.Request, source.RequestAborted),
                 ContentLength = from.ContentLength,
                 ContentType = from.ContentType,
                 Host = from.Host,
@@ -257,17 +257,17 @@ public class MultiplexingMiddleware : OcelotMiddleware
         return aggregator.Aggregate(route, httpContext, contexts);
     }
 
-    protected virtual async Task<Stream> CloneRequestBodyAsync(HttpContext source)
+    protected virtual async Task<Stream> CloneRequestBodyAsync(HttpRequest request, CancellationToken aborted)
     {
-        source.Request.EnableBuffering();
-        if (source.Request.Body.Position == 0)
+        request.EnableBuffering();
+        if (request.Body.Position == 0)
         {
             var targetBuffer = new MemoryStream();
-            if (source.Request.ContentLength is not null)
+            if (request.ContentLength is not null)
             {
-                await source.Request.Body.CopyToAsync(targetBuffer, (int)source.Request.ContentLength, source.RequestAborted);
+                await request.Body.CopyToAsync(targetBuffer, (int)request.ContentLength, aborted);
                 targetBuffer.Position = 0;
-                source.Request.Body.Position = 0;
+                request.Body.Position = 0;
             }
 
             return targetBuffer;
@@ -275,7 +275,7 @@ public class MultiplexingMiddleware : OcelotMiddleware
         else
         {
             Logger.LogWarning("Ocelot does not support body copy without stream in initial position 0");
-            return source.Request.Body;
+            return request.Body;
         }
     }
 }
--- a/src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
+++ b/src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
@@ -184,7 +184,7 @@ public class MultiplexingMiddleware : OcelotMiddleware
     /// <returns>The cloned Http context.</returns>
     private async Task<HttpContext> ProcessRouteAsync(HttpContext sourceContext, DownstreamRoute route, List<PlaceholderNameAndValue> placeholders = null)
     {
-        var newHttpContext = CreateThreadContext(sourceContext);
+        var newHttpContext = await CreateThreadContextAsync(sourceContext);
         CopyItemsToNewContext(newHttpContext, sourceContext, placeholders);
         newHttpContext.Items.UpsertDownstreamRoute(route);
 
@@ -208,14 +208,14 @@ public class MultiplexingMiddleware : OcelotMiddleware
     /// </summary>
     /// <param name="source">The base http context.</param>
     /// <returns>The cloned context.</returns>
-    private static HttpContext CreateThreadContext(HttpContext source)
+    protected virtual async Task<HttpContext> CreateThreadContextAsync(HttpContext source)
     {
         var from = source.Request;
         var target = new DefaultHttpContext
         {
             Request =
             {
-                Body = from.Body, // TODO Consider stream cloning for multiple reads
+                Body = await CopyBufferToTargetRequestAsync(source),
                 ContentLength = from.ContentLength,
                 ContentType = from.ContentType,
                 Host = from.Host,
@@ -255,5 +255,27 @@ public class MultiplexingMiddleware : OcelotMiddleware
 
         var aggregator = _factory.Get(route);
         return aggregator.Aggregate(route, httpContext, contexts);
+    }
+
+    protected virtual async Task<Stream> CopyBufferToTargetRequestAsync(HttpContext source)
+    {
+        source.Request.EnableBuffering();
+        if (source.Request.Body.Position == 0)
+        {
+            var targetBuffer = new MemoryStream();
+            if (source.Request.ContentLength is not null)
+            {
+                await source.Request.Body.CopyToAsync(targetBuffer, (int)source.Request.ContentLength, source.RequestAborted);
+                targetBuffer.Position = 0;
+                source.Request.Body.Position = 0;
+            }
+
+            return targetBuffer;
+        }
+        else
+        {
+            Logger.LogWarning("Ocelot does not support body copy without stream in initial position 0");
+            return source.Request.Body;
+        }
     }
 }
--- a/src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
+++ b/src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
@@ -19,8 +19,7 @@ public class MultiplexingMiddleware : OcelotMiddleware
 
     public MultiplexingMiddleware(RequestDelegate next,
         IOcelotLoggerFactory loggerFactory,
-        IResponseAggregatorFactory factory
-    )
+        IResponseAggregatorFactory factory)
         : base(loggerFactory.CreateLogger<MultiplexingMiddleware>())
     {
         _factory = factory;
@@ -211,11 +210,12 @@ public class MultiplexingMiddleware : OcelotMiddleware
     protected virtual async Task<HttpContext> CreateThreadContextAsync(HttpContext source)
     {
         var from = source.Request;
+        var bodyStream = await CloneRequestBodyAsync(from, source.RequestAborted);
         var target = new DefaultHttpContext
         {
             Request =
             {
-                Body = await CloneRequestBodyAsync(source.Request, source.RequestAborted),
+                Body = bodyStream,
                 ContentLength = from.ContentLength,
                 ContentType = from.ContentType,
                 Host = from.Host,
@@ -237,12 +237,13 @@ public class MultiplexingMiddleware : OcelotMiddleware
             RequestAborted = source.RequestAborted,
             User = source.User,
         };
-
         foreach (var header in from.Headers)
         {
             target.Request.Headers[header.Key] = header.Value.ToArray();
         }
 
+        // Once the downstream request is completed and the downstream response has been read, the downstream response object can dispose of the body's Stream object
+        target.Response.RegisterForDisposeAsync(bodyStream); // manage Stream lifetime by HttpResponse object
         return target;
     }
 
--- a/src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
+++ b/src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
@@ -269,6 +269,10 @@ public class MultiplexingMiddleware : OcelotMiddleware
                 targetBuffer.Position = 0;
                 request.Body.Position = 0;
             }
+            else
+            {
+                Logger.LogWarning("Aggregation does not support body copy without Content-Length header!");
+            }
 
             return targetBuffer;
         }
--- a/src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
+++ b/src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
@@ -260,26 +260,24 @@ public class MultiplexingMiddleware : OcelotMiddleware
     protected virtual async Task<Stream> CloneRequestBodyAsync(HttpRequest request, CancellationToken aborted)
     {
         request.EnableBuffering();
-        if (request.Body.Position == 0)
-        {
-            var targetBuffer = new MemoryStream();
-            if (request.ContentLength is not null)
-            {
-                await request.Body.CopyToAsync(targetBuffer, (int)request.ContentLength, aborted);
-                targetBuffer.Position = 0;
-                request.Body.Position = 0;
-            }
-            else
-            {
-                Logger.LogWarning("Aggregation does not support body copy without Content-Length header!");
-            }
-
-            return targetBuffer;
-        }
-        else
+        if (request.Body.Position != 0)
         {
             Logger.LogWarning("Ocelot does not support body copy without stream in initial position 0");
             return request.Body;
         }
+
+        var targetBuffer = new MemoryStream();
+        if (request.ContentLength is not null)
+        {
+            await request.Body.CopyToAsync(targetBuffer, (int)request.ContentLength, aborted);
+            targetBuffer.Position = 0;
+            request.Body.Position = 0;
+        }
+        else
+        {
+            Logger.LogWarning("Aggregation does not support body copy without Content-Length header!");
+        }
+
+        return targetBuffer;
     }
 }
--- a/docs/features/requestaggregation.rst
+++ b/docs/features/requestaggregation.rst
@@ -217,9 +217,12 @@ Below is an example of an aggregator that you could implement for your solution:
 Gotchas
 -------
 
-You cannot use Routes with specific **RequestIdKeys** as this would be crazy complicated to track.
+* You cannot use Routes with specific **RequestIdKeys** as this would be crazy complicated to track.
+* Aggregation only supports the ``GET`` HTTP verb.
+* Aggregation allows for the forwarding of ``HttpRequest.Body`` to downstream services by duplicating the body data.
+  Form data and attached files should also be forwarded.
+  It is essential to always specify the ``Content-Length`` header in requests to upstream; otherwise, Ocelot will log warnings like *"Aggregation does not support body copy without Content-Length header!"*.
 
-Aggregation only supports the ``GET`` HTTP verb.
 
 """"
 
--- a/test/Ocelot.AcceptanceTests/AggregateTests.cs
+++ b/test/Ocelot.AcceptanceTests/AggregateTests.cs
@@ -639,8 +639,8 @@ namespace Ocelot.AcceptanceTests
                 new KeyValuePair<string, string>("param2", "from-form-REPLACESTRING"),
             };
 
-            var sub1ResponseContent = @"""[key:param1=value1&param2=from-form-s1]""";
-            var sub2ResponseContent = @"""[key:param1=value1&param2=from-form-s2]""";
+            var sub1ResponseContent = "\"[key:param1=value1&param2=from-form-s1]\"";
+            var sub2ResponseContent = "\"[key:param1=value1&param2=from-form-s2]\"";
             var expected = $"{{\"Service1\":{sub1ResponseContent},\"Service2\":{sub2ResponseContent}}}";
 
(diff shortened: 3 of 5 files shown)


src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
```diff
@@ -255,5 +255,27 @@ protected virtual Task MapAsync(HttpContext httpContext, Route route, List<HttpC
 
         var aggregator = _factory.Get(route);
         return aggregator.Aggregate(route, httpContext, contexts);
+    }
+
+    protected virtual async Task<Stream> CopyBufferToTargetRequestAsync(HttpContext source)
+    {
+        source.Request.EnableBuffering();
```
PaulARoy, line 262: This `EnableBuffering()` extension method comes from the AspNet framework: https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httprequestrewindextensions.enablebuffering?view=aspnetcore-8.0

This seems to be the recommended way to enable native capabilities: https://devblogs.microsoft.com/dotnet/re-reading-asp-net-core-request-bodies-with-enablebuffering/

Feel free to discuss this solution here, I think it's pretty elegant.

raman-m:
Paul, thanks for re-creation the PR!

>I looked at the benchmark tests but I'm not sure on how to integrate this correctly.

Benchmarks testing project is here: [Ocelot.Benchmarks](https://github.com/PaulARoy/Ocelot/tree/feat/multiplexing-body-copy/test/Ocelot.Benchmarks)
So, as I said it is not required, but let @ggnaegi to decide on updating benchmarks. But he is on sick leave now...

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
raman-m: We're aware of this fantastic ASP.NET feature 😄. However, as a team, we've opted not to use `HttpClient` and instead utilize the new `HttpMessageInvoker` pooling mechanism to enhance our heavy streaming capabilities.
For multiplexing, we should employ local buffering and replicate the body multiple times into downstreams.

raman-m: ### The Q
>Ensure the request [Body](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httprequest.body?view=aspnetcore-8.0#microsoft-aspnetcore-http-httprequest-body) **can be read multiple times**. Normally buffers request bodies in memory; writes requests larger than 30K bytes to disk.

>Ensure the request body can be read multiple times.

This is exactly what we need for multiplexing scenario!

raman-m:
> Would you like to move forward and add benchmark tests afterwards or do you prefer to wait for @ggnaegi? I wish him well!

Let's wait for @ggnaegi plz!

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
```diff
@@ -255,5 +255,27 @@ protected virtual Task MapAsync(HttpContext httpContext, Route route, List<HttpC
...
         return aggregator.Aggregate(route, httpContext, contexts);
+    }
+
+    protected virtual async Task<Stream> CloneRequestBodyAsync(HttpContext source)
+    {
+        source.Request.EnableBuffering();
+        if (source.Request.Body.Position == 0)
+        {
+            var targetBuffer = new MemoryStream();
+            if (source.Request.ContentLength is not null)
```
raman-m, line 266: What to do if the header is `null`?

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
PaulARoy: @raman-m thanks for the input, it is enabled on ASP.NET Request, not on HttpClient.

Where would you put the settings for this buffering? I thought it would be preferable to let it by default at first because this would not be enabled for requests that are not multiplexed, thus being a bit confusing.

Also buffering was not part of the story which is why I did the strict minimum buffering. Considering the remark on the previous PR on overstepping the boundaries of the bug, I think it would be best to postpone this (interesting) addition/customization to a more global buffering story.

PaulARoy:
> > Would you like to move forward and add benchmark tests afterwards or do you prefer to wait for @ggnaegi? I wish him well!
> 
> 1. Let's wait for @ggnaegi plz!
> 2. Paul, have you tested the fix in your staging environment or locally? You don't see original `HttpRequestException` from [Body cannot be forwarded twice on Aggregator #2039](https://github.com/ThreeMammals/Ocelot/issues/2039) in logs, right?

Yes, I tested locally first. Also the acceptance test can show quite efficiently the resolution (disabling the copy fails the test with the correct exception).

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
```diff
@@ -208,14 +208,14 @@ private static void CopyItemsToNewContext(HttpContext target, HttpContext source
...
     /// <returns>The cloned context.</returns>
-    private static HttpContext CreateThreadContext(HttpContext source)
+    protected virtual async Task<HttpContext> CreateThreadContextAsync(HttpContext source)
     {
         var from = source.Request;
         var target = new DefaultHttpContext
         {
             Request =
             {
-                Body = from.Body, // TODO Consider stream cloning for multiple reads
```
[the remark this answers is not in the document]
PaulARoy: Theoretically, I think so yes. But this is a way, way more complex mechanism and I'm not sure I know how to do this safely.

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
```diff
@@ -208,14 +208,14 @@ private static void CopyItemsToNewContext(HttpContext target, HttpContext source
...
+                Body = await CloneRequestBodyAsync(source),
```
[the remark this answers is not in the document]
PaulARoy: `Form` seems to rely on `Body` from what I can see in the aspnet github repository yeah

I'll test this to validate the hypothesis

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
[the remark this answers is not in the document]
raman-m: It's not an issue, merely a question. Yes, I understand... It's an enhancement that could be worked on in the future.

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
PaulARoy: I just forward an empty `MemoryStream`.

Default implementation uses a `NullStream`, an internal inaccessible subclass.

raman-m:
> Commit https://github.com/ThreeMammals/Ocelot/pull/2050/commits/8e399b06a463a6d3d417f89361583c1edb7a43b9

@PaulARoy  What is that? 
Don't edit C# code using GitHub editor please! Never! Use Visual Studio only.
Could you remove this commit please? `git reset --hard HEAD~1` should help.

PaulARoy:
> > Commit [8e399b0](https://github.com/ThreeMammals/Ocelot/commit/8e399b06a463a6d3d417f89361583c1edb7a43b9)
> 
> @PaulARoy What is that? Don't edit C# code using GitHub editor please! Never! Use Visual Studio only. Could you remove this commit please? `git reset --hard HEAD~1` should help.

I'm fixing it yeah. I hoped it wasn't that bad – I was wrong.

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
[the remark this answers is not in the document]
raman-m: Thank you!

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
```diff
@@ -215,7 +215,8 @@ protected virtual async Task<HttpContext> CreateThreadContextAsync(HttpContext s
         {
             Request =
             {
-                Body = await CloneRequestBodyAsync(source.Request, source.RequestAborted),
+                Body = await CloneRequestBodyAsync(source.Request, source.RequestAborted),
+                /* Form = from.Form, // TODO */
```
PaulARoy, line 219: No, Form is not required to be mapped here.

Form is just an interpretation of the body and copying the body is enough to make it work.

Acceptance test `Should_return_response_200_with_copied_form_sent_on_multiple_service` validates this point by sending a form, reading it and responding with data from it.

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
raman-m: This information should be included in the feature documentation. Developers need to understand that if a request contains a body, the `Content-Length` header must be present; otherwise, the body will be transmitted as `null` or empty.
Do you agree?

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
raman-m: I'm confused and surprised, but it's all good. If the Form property only reads the Body, there isn't much to discuss, except perhaps for Files. However, since Files is a sub-property of Form, I anticipate there will be no issues, and the files should be forwarded as well, correct?

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
PaulARoy: Hum that's actually an HTTP convention `https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Length`

I don't think it's necessary to document this: not sending a content-length header will prevent all http compliant servers not to read the body (in theory).

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
PaulARoy: I hesitated to put a file in an acceptance test, but yeah everything relies on the body and how it should be interpreted through the application content type headers. So yes, correct as far as I know.

It was a nice surprise for me too :)

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
```diff
@@ -255,5 +255,27 @@ protected virtual Task MapAsync(HttpContext httpContext, Route route, List<HttpC
...
+        request.EnableBuffering();
+        if (request.Body.Position == 0)
+        {
+            var targetBuffer = new MemoryStream();
+            if (request.ContentLength is not null)
+            {
+                await request.Body.CopyToAsync(targetBuffer, (int)request.ContentLength, aborted);
+                targetBuffer.Position = 0;
+                request.Body.Position = 0;
+            }
```
raman-m, line 271:
```suggestion
            }
            else
            {
                Logger.LogWarning("Aggregation does not support body copy without Content-Length header!");
            }
```

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
```diff
@@ -255,5 +255,27 @@ protected virtual Task MapAsync(HttpContext httpContext, Route route, List<HttpC
 
         var aggregator = _factory.Get(route);
         return aggregator.Aggregate(route, httpContext, contexts);
+    }
+
+    protected virtual async Task<Stream> CloneRequestBodyAsync(HttpRequest request, CancellationToken aborted)
+    {
+        request.EnableBuffering();
```
raman-m, line 262: ### Next goal
- [EnableBuffering(HttpRequest, Int32, Int64)](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httprequestrewindextensions.enablebuffering?view=aspnetcore-8.0#microsoft-aspnetcore-http-httprequestrewindextensions-enablebuffering(microsoft-aspnetcore-http-httprequest-system-int32-system-int64))

Let's add `MultiplexingMiddleware` options.

**Given** options for buffering from `FileAggregateRoute` class:
```csharp
public struct BufferingOptions
{
    public int BufferThreshold;
    public long BufferLimit;
}
```
**Then**, let's apply the options:
```suggestion
    protected virtual async Task<Stream> CloneRequestBodyAsync(HttpRequest request, CancellationToken aborted, BufferingOptions options)
    {
        request.EnableBuffering(options.BufferThreshold, options.BufferLimit);
```

raman-m (changes requested):
I think it's necessary to introduce additional buffering options in the middleware.
Please consider my suggestions listed below. 👇

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
raman-m: If you are inspired by the file forwarding feature, feel free to update the relevant acceptance test or write a new one.
Thank you!

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
raman-m: Your [concerns](https://github.com/ThreeMammals/Ocelot/pull/2050#discussion_r1567333939) about introducing this feature have been noted. It is designed to be non-global and solely for multiplexing, as buffering is permitted only for multiplexed requests.

@ggnaegi, could you share your thoughts on this?

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
PaulARoy: @raman-m it's done

I was on board for a test with a form file, but if it's enough for you I don't think it's necessary.

raman-m (approved):
LGFM

raman-m:
@ggnaegi @RaynaldM  Hi Team!
Could you take a look at this multiplexer improvement please?

@ggnaegi Gui, you are key reviewer for this PR!

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
```diff
@@ -255,5 +255,31 @@ protected virtual Task MapAsync(HttpContext httpContext, Route route, List<HttpC
 
         var aggregator = _factory.Get(route);
         return aggregator.Aggregate(route, httpContext, contexts);
+    }
+
+    protected virtual async Task<Stream> CloneRequestBodyAsync(HttpRequest request, CancellationToken aborted)
+    {
+        request.EnableBuffering();
+        if (request.Body.Position == 0)
```
ggnaegi, line 263: why not try to reduce nesting?
```c#
if (request.Body.Position != 0)
{
    return request.Body;
}
```

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
```diff
@@ -255,5 +255,31 @@ protected virtual Task MapAsync(HttpContext httpContext, Route route, List<HttpC
...
+            else
+            {
+                Logger.LogWarning("Aggregation does not support body copy without Content-Length header!");
+            }
+
+            return targetBuffer;
+        }
+        else
+        {
+            Logger.LogWarning("Ocelot does not support body copy without stream in initial position 0");
```
ggnaegi, line 281: By the way, should this possibly happen? I'm wondering if we should throw here...

ggnaegi (commented):
Just 1-2 suggestions but fine otherwise, thanks

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
raman-m: I'd keep this logged warning just for debugging and proving our solution with the fix is functioning.
So, we have improved version of multiplexer but it is draft from the point of view of Production environment.
Let's keep this `else` branch!

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
raman-m: Make sense! 😄

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
PaulARoy: @ggnaegi I don't think it should happen in a normal scenario. The only way I can think of is Ocelot reading partially the body elsewhere by mistake.

However any type of custom middleware could potentially be reading a partial body. I don't see any use case right now but I think it's preferable to log instead of throwing an exception.

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
```diff
@@ -260,26 +260,24 @@ protected virtual Task MapAsync(HttpContext httpContext, Route route, List<HttpC
...
+            request.Body.Position = 0;
         }
         else
         {
-            Logger.LogWarning("Ocelot does not support body copy without stream in initial position 0");
-            return request.Body;
+            Logger.LogWarning("Aggregation does not support body copy without Content-Length header!");
         }
+
+        return targetBuffer;
```
raman-m, line 281: [1]: https://stackoverflow.com/questions/8116541/what-exactly-is-streambuf-how-do-i-use-it ""
[2]: https://docs.unity3d.com/ScriptReference/AssetBundle.LoadFromStream.html ""
[3]: https://docs.simpleinjector.org/en/latest/lifetimes.html ""

Oops! I forgot to ask... How do we manage the `Stream` object's lifetime here?

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
raman-m: Fixed by https://github.com/ThreeMammals/Ocelot/pull/2050/commits/33d1743311d1e8c4bef27a656101ad5537fc11af

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
raman-m: Not an issue! Resolving...

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
PaulARoy: > @PaulARoy Regarding the management of the `Stream` object's lifetime, it appears we are relying solely on Garbage Collector to handle it.

Yes, I assume it's disposed once the request itself is disposed by aspnet

ggnaegi: I agree with @raman-m, it might be disposed... But since it's a clone of the original HttpContext, I'm not so sure anymore. Maybe we should try to dispose the cloned http contexts afterwards?

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
ggnaegi: I'm not a fan of deep nesting because the code can become very difficult to read. If we agree on a specific pattern we should follow it and we tend to use this pattern in later implementations in Ocelot.

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
raman-m: ### The issue has been resolved in the commit  https://github.com/ThreeMammals/Ocelot/pull/2050/commits/897c3488112d1f029639b27124831cf0556acf0f ✔️

I was wrong! The `RegisterForDispose(IDisposable)` method doesn't require implementation of `IDisposable` interface in the middleware class.

raman-m (approved):
### Ready for delivery❗

raman-m:
### TODO
- [ ] Next goal is https://github.com/ThreeMammals/Ocelot/pull/2050#discussion_r1568781288

PR #2058, opened 2024-04-30
#2054 #2059 Manage `EnableContentHashing` setting by global `CacheOptions`

### LINKED ISSUE #2059: `FileCacheOptions` not working after the header was introduced in FileCache settings in version 23.0.0
### Discussed in https://github.com/ThreeMammals/Ocelot/discussions/2054
Hi

We are facing issue regarding cache option, however in  earlier version till v22.0.1  FileCacheOptions was working fine but after the release of v23.0.0 we are facing issue mentioned below in two cases.

**Case 1: V.22.0.1
"FileCacheOptions": { "TtlSeconds": 15, "Region": "europe-central" }** 

It was working fine with the multiple requests. before the latest update 
case2:
After the addition of new parameter named Header in FileCacheOptions there, we are facing issue on multiple requests in case one request is already generated and in the same configured cache expiry time if another request come with different values, then we are getting same response as per the first request.

**Case 2: v23.2.2
"FileCacheOptions": {
  "TtlSeconds": 15,
  "Region": "europe-central"
}**

As per documentation header is not mandatory, hence the cache option should work without header also, which  is not the case. 

Observed Behaviour 
1. File cache returns same data for different request payload on same Route in new version 23.2.2 (we tested in this version directly)
2. File cache returns different data for different request payload on same Route in version 22.0.1.

Please can you help us fix this as this seems to be a breaking change since version 23.0.0 as mentioned by Raman in the discussion.

Thanks,
Sheru Kumar Gaur

thiagoloureiro:
## Fixes #2059 
- #2054 
- #2059 

## Proposed Changes
This PR is prepared to fix the `EnableContentHashing` not being considered on `ocelot.json`.

[FileCacheOptions](https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/Configuration/File/FileCacheOptions.cs) didn't have the option to fetch from the config and also the [RoutesCreator](https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/Configuration/Creator/RoutesCreator.cs) didn't call the overload with the settings.


raman-m:
Thank you, Thiago!
What issue is it related to?

src/Ocelot/Configuration/Creator/RoutesCreator.cs
```diff
@@ -122,7 +122,7 @@ private DownstreamRoute SetUpDownstreamRoute(FileRoute fileRoute, FileGlobalConf
                 .WithClaimsToDownstreamPath(claimsToDownstreamPath)
                 .WithRequestIdKey(requestIdKey)
                 .WithIsCached(fileRouteOptions.IsCached)
-                .WithCacheOptions(new CacheOptions(fileRoute.FileCacheOptions.TtlSeconds, region, fileRoute.FileCacheOptions.Header))
+                .WithCacheOptions(new CacheOptions(fileRoute.FileCacheOptions.TtlSeconds, region, fileRoute.FileCacheOptions.Header, fileRoute.FileCacheOptions.EnableContentHashing))
```
ggnaegi, line 125: @raman-m maybe we should foresee a CacheOptions creator? We are already retrieving the region ```var region = _regionCreator.Create(fileRoute);```, why couldn't we deal with the other properties in a CacheOptions creator?

ggnaegi (commented):
Hello thanks, it's fine, but I would like to check the CacheOptions creation. It looks a bit like a hack...

src/Ocelot/Configuration/Creator/RoutesCreator.cs
raman-m: Historically, Ocelot utilizes the design concept of `Creator` classes. Regarding copying constructors in C#, a constructor may accept parameters of any type. Employing a constructor can significantly reduce the amount of code written, as exemplified in line 125.

ggnaegi: Yes, I agree, but I was wondering if we could have a creator that would include the region creation too? So we wouldn't add too much lines of codes and we would group the cache options creation... The code is then easier to read.

raman-m: The decision should be left to the author. Personally, I favor constructors for simple initialization and member assignments. However, for complex logic, such as calling an interface object, it is preferable to define a "creator" class with dependencies.

ggnaegi: @raman-m Ok, so I decided to review this code since it seems to me that several users are now complaining about it. No feedback from the author until now... So I would recommend following the  creator strategy.  By the way, when the new cache has been implemented, at least one acceptance test verifying the EnableContentHashing option should have been foreseen, but that's not the case. I will add an acceptance test for that matter.

ggnaegi:
It is related to the following issue: https://github.com/ThreeMammals/Ocelot/issues/2059

ggnaegi:
@raman-m @thiagoloureiro I have added a global configuration for the cache options. Since we induced a breaking change by adding the property EnableContentHashing and setting it to false by default, I think the users shouldn't bear the burden. That's why I thought adding the global configuration point EnableContentHashing would help most of them. The documentation will be updated too.

ggnaegi:
@raman-m It's ready for review. I'm still using Bddfy, it was straightforward. And there is a silly problem the application is expecting TtlSeconds to be explicitly set for each route, since TtlSeconds > 0 -> IsCached.

src/Ocelot/Configuration/Creator/RoutesCreator.cs
raman-m: > No feedback from the author until now... 

According to my statistics, approximately 9 out of 10 pull requests in this project from open-source contributors consist of one or two commits that express an idea but lack a solid understanding of how to implement the feature or fix. As a team, we must adapt to this situation and ensure the delivery is taken care of.
Sad, but true...

----

> So I would recommend following the creator strategy.

### 👌

src/Ocelot/Configuration/Creator/RouteOptionsCreator.cs
```diff
@@ -17,6 +17,8 @@ public RouteOptions Create(FileRoute fileRoute)
                 && (!string.IsNullOrEmpty(authOpts.AuthenticationProviderKey)
                     || authOpts.AuthenticationProviderKeys?.Any(k => !string.IsNullOrWhiteSpace(k)) == true);
             var isAuthorized = fileRoute.RouteClaimsRequirement?.Any() == true;
+
+            // TODO: This sounds more like a hack, it might be better to refactor this at some point.
```
raman-m, line 21: Agreed, the property has one actual reference.
image
**The property is deemed absolutely useless**, indicative of the junior developer syndrome.

src/Ocelot/DependencyInjection/OcelotBuilder.cs
```diff
@@ -114,9 +112,11 @@ public OcelotBuilder(IServiceCollection services, IConfiguration configurationRo
             Services.TryAddSingleton<IHttpHandlerOptionsCreator, HttpHandlerOptionsCreator>();
             Services.TryAddSingleton<IDownstreamAddressesCreator, DownstreamAddressesCreator>();
             Services.TryAddSingleton<IDelegatingHandlerHandlerFactory, DelegatingHandlerHandlerFactory>();
-            Services.TryAddSingleton<ICacheKeyGenerator, DefaultCacheKeyGenerator>();
+
```
raman-m, line 115: Line 82 is also redundant for the `IRegionCreator` interface.

raman-m (changes requested):
Please note that we have introduced a new `CacheOptionsCreator` class, yet it lacks unit tests. Conversely, the addition of [new acceptance tests](https://github.com/ThreeMammals/Ocelot/pull/2058/files#diff-90ea0d8bf4ce64201097435dc591479ddc4533c3e5659ffa2a107af47f7cf474R119-R186) for the #2059 bug is commendable!

And another my suggestions are below 👇 

### P.S.
Concerning breaking changes, let's direct our focus to the JSON property named `FileCacheOptions`, which represents an outdated naming error. Consequently, let's consider adding [another breaking change](https://github.com/ThreeMammals/Ocelot/issues/2059#issuecomment-2105650203) that I've suggested. 😄

src/Ocelot/DependencyInjection/OcelotBuilder.cs
ggnaegi: ok

src/Ocelot/Cache/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Ocelot.Cache;
+
+public static class ServiceCollectionExtensions
+{
+    /// <summary>
+    /// Extension method used to add Ocelot cache services to the DI container.
+    /// </summary>
+    /// <param name="services">The services collection.</param>
+    public static void AddOcelotCache(this IServiceCollection services)
```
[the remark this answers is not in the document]
ggnaegi: ok!

src/Ocelot/Cache/IRegionCreator.cs
```diff
@@ -1,9 +1,6 @@
...
+namespace Ocelot.Cache;
 
-namespace Ocelot.Cache
+public interface IRegionCreator
 {
-    public interface IRegionCreator
-    {
-        string Create(FileRoute route);
-    }
+    string Create(string region, string upstreamPathTemplate, IList<string> upstreamHttpMethod);
```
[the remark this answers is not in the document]
ggnaegi: Ok, let's remove it.

src/Ocelot/Cache/RegionCreator.cs
```diff
@@ -1,21 +1,19 @@
-using Ocelot.Configuration.File;
+using Ocelot.Configuration.File;
 
-namespace Ocelot.Cache
+namespace Ocelot.Cache;
+
+public class RegionCreator : IRegionCreator
 {
-    public class RegionCreator : IRegionCreator
+    public string Create(string region, string upstreamPathTemplate, IList<string> upstreamHttpMethod)
```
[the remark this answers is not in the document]
ggnaegi: ok

src/Ocelot/Configuration/Creator/RouteOptionsCreator.cs
ggnaegi: @raman-m The property is not useless, that's very unfortunate, it is used to check if the cache should be used or not...
https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/Cache/Middleware/OutputCacheMiddleware.cs#L28-L32 I think, it must be refactored, but it's out of scope and could have side effects. So I would put that in a follow-up PR.

ggnaegi:
@raman-m added some unit tests for the CacheOptionsCreator and further suggestions

ggnaegi (approved):

src/Ocelot/Configuration/Creator/RouteOptionsCreator.cs
raman-m: Indeed, it is utilized in the middleware. My point is that propagating a simple Boolean value across two classes constitutes an overly complex design. This issue pertains to the entire `RouteOptions` class.

However, this is not a concern for the current PR.

test/Ocelot.UnitTests/Cache/RegionCreatorTests.cs
raman-m, line 1: It should be renamed to `CacheOptionsCreatorTests`!

raman-m (approved):
Technically we could merge, but I'd like to propose more...
Wait for my next commits plz!

ggnaegi:
@raman-m it's sound for me, should I squash and merge?

src/Ocelot/Cache/DefaultMemoryCache.cs
```diff
@@ -2,12 +2,12 @@
 
 namespace Ocelot.Cache
 {
-    public class AspMemoryCache<T> : IOcelotCache<T>
+    public class DefaultMemoryCache<T> : IOcelotCache<T>
     {
         private readonly IMemoryCache _memoryCache;
         private readonly Dictionary<string, List<string>> _regions;
```
RaynaldM, line 8:
```suggestion
        private readonly IDictionary<string, List<string>> _regions;
```

RaynaldM (approved):

raman-m:
@ggnaegi Please don't merge yet! Hold on for a 10-20 minutes... I want to present a commit related to the property name.

raman-m:
@ggnaegi The commit found at https://github.com/ThreeMammals/Ocelot/pull/2058/commits/3290aa64f0b2f4b16d929d5acee4ddb10ae1dae5 reflects my earlier [suggestion](https://github.com/ThreeMammals/Ocelot/pull/2058#issuecomment-2107869883) regarding [the proposed renaming](https://github.com/ThreeMammals/Ocelot/issues/2059#issuecomment-2105650203). I'm confident that this commit does not introduce any breaking changes. Please share your thoughts.

src/Ocelot/Cache/DefaultMemoryCache.cs
raman-m: @RaynaldM Hey Ray!
The `_regions` is an internal private class member and is not visible to class clients. Additionally, this member is not initialized by the constructor's argument. I also don't see a benefit in changing to a more abstract type since the member isn't exposed for testing. It would be more logical to change the storage type for rapid concurrent operations (for thread safety), but as far as I know, `Dictionary<string, List<string>>` is [thread-safe](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2?view=net-8.0#thread-safety), correct?

raman-m:
@ggnaegi [commented](https://github.com/ThreeMammals/Ocelot/pull/2058#issuecomment-2107801717):
> it's sound for me, should I squash and merge?

Yes, please do, provided you have no objections to my last commit https://github.com/ThreeMammals/Ocelot/pull/2058/commits/3290aa64f0b2f4b16d929d5acee4ddb10ae1dae5.

raman-m:
### Ready for delivery ✅ 
- Code review ✔️✔️✔️
- Team approvals ✔️✔️✔️
- Unit testing ✔️ &rarr; `CacheOptionsCreatorTests` &rarr; [link](https://github.com/ThreeMammals/Ocelot/pull/2058/files#diff-ee75b7fec590548b243e0be694415879b2360931ae8524cff8d022399b91ae10)
- Acceptance testing ✔️ &rarr; `CachingTests`  &rarr; [link to 2 tests](https://github.com/ThreeMammals/Ocelot/pull/2058/files#diff-90ea0d8bf4ce64201097435dc591479ddc4533c3e5659ffa2a107af47f7cf474R119-R186)
- Updated feature docs ✔️ &rarr; `docs/features/caching.rst` &rarr; [link](https://github.com/ThreeMammals/Ocelot/pull/2058/files#diff-f538914b011cae5b5b13cb60f35d484a8fceef6569520783161d100d25d0b8f1)

ggnaegi:
@raman-m I'm not sure about FileCacheOptions being renamed to CacheOptions. I would prefer keeping FileCacheOptions for now. For me it was all sound and we are going now a step too far. I would prefer having those changes in another PR.

raman-m:
@ggnaegi Understood! Reverting...

raman-m:
@ggnaegi 
The recent changes have been reverted. I think we're ready to merge now. 😃

ggnaegi:
@raman-m Ok, I got it, I just realized I named the Global FileCacheOptions CacheOptions... Well... Are you ok with that, and we announce the community that in a next version we will rename it?

raman-m:
### TODO
- [ ] Renaming from `FileCacheOptions` to `CacheOptions`, after release

thiagoloureiro:
thanks for adding all the comments and commits guys!

Issue #1869, opened 2023-12-26
Add to configuration hardcoded timeout of HttpClient


RightDecision2:
## Expected Behavior / New Feature
// HttpClientBuilder.cs
`_defaultTimeout = TimeSpan.FromSeconds(_configuration.TimeOut);`

## Actual Behavior / Motivation for New Feature
// HttpClientBuilder.cs
`_defaultTimeout = TimeSpan.FromSeconds(90);`

## Specifications

  - Version: All
  - Platform: All
  - Subsystem: Any

raman-m:
Are you going to create a PR these days?

RightDecision2:
Yes

вт, 26 дек. 2023 г., 22:40 Raman Maksimchuk ***@***.***>:

raman-m:
What is the problem with default 90 seconds timeout?

As a quick fix use QoS [TimeoutValue](https://ocelot.readthedocs.io/en/latest/features/qualityofservice.html#quality-of-service:~:text=You%20can%20set%20the%20TimeoutValue%20in%20isolation%20of%20the%20ExceptionsAllowedBeforeBreaking%20and%20DurationOfBreak%20options%3A) property to have custom timeout value. 
I don't think that performance will decrease dramatically... A few microseconds will be added 😄

raman-m:
>@RightDecision2 commented [on Dec 26](https://github.com/ThreeMammals/Ocelot/issues/1869#issuecomment-1869713955)

👌 Great! Waiting for your PR creation... Will review with pleasure!

RightDecision2:
I have OCR service, which process pdf file. Every request average 2mins.
When I use QuOS its random cancelled after 90 seconds (few times might be
worked).
So fo me, is only one possible solution - increase default timeout. I think
for someone it could be usefull to.

вт, 26 дек. 2023 г. в 22:52, Raman Maksimchuk ***@***.***>:

raman-m:
Dear author, What's your name?
What's your LinkedIn?

raman-m:
> I have OCR service, which process pdf file. Every request average 2mins.
> **When I use QoS its random cancelled after 90 seconds (few times might be worked).**
> 

Strange cancellation of your requests by QoS logic... You need to show this behavior by showing us Ocelot logs.
It should not be cancelled randomly... All requests should be done without problems if `TimeoutValue` in QoS is greater than 2 mins (average processing time by your downstream service). If some requests failed that means root cause is not on Ocelot's side...

[...]

Issue #1314, opened 2020-08-17
Make default timeout for HttpClientBuilder configurable


danielcandea:
## New Feature
Make default timeout for [HttpClientBuilder](https://github.com/ThreeMammals/Ocelot/blob/master/src/Ocelot/Requester/HttpClientBuilder.cs) configurable.

## Motivation for New Feature
It is easier to set a custom default global timeout than adding QoS settings for each handler, especially if there are more than one handlers defined.

raman-m:
@RaynaldM Useful?
Could your team develop this feature please?

The issue has been shifted to the current monthly [May-June'24](https://github.com/ThreeMammals/Ocelot/milestone/7) release to provide your team with ample time for development.

raman-m:
@danielcandea Do you have intention to contribute?

danielcandea:
@raman-m I'm not in the team using Ocelot anymore, for a long time now. The file I mentioned doesn't exist anymore, I would need to get familiar with the code again, after almost four years.
I can't commit time at the moment but would love to give this a go at some point in in the future. If this isn't ok for you, I'm happy to close this.

raman-m:
@danielcandea Thanks for the feedback!
This issue will not be closed. We will develop it within the current [May-June'24](https://github.com/ThreeMammals/Ocelot/milestone/7) milestone.

RaynaldM:
> @RaynaldM Useful? Could your team develop this feature please?
> 
> The issue has been shifted to the current monthly [May-June'24](https://github.com/ThreeMammals/Ocelot/milestone/7) release to provide your team with ample time for development.

It is

PR #2073, opened 2024-05-25
RequestTimeoutSeconds FileGlobalConfiguration

hogwartsdeveloper:
Fixes / New Feature #

## Proposed Changes

  - FileGlobalConfiguration add RequestTimeoutSeconds
  - MessageInvokerPool use RequestTimeoutSeconds

issues #1869


src/Ocelot/Configuration/File/FileGlobalConfiguration.cs
```diff
@@ -17,6 +17,8 @@ public FileGlobalConfiguration()
 
         public string RequestIdKey { get; set; }
 
+        public int RequestTimeoutSeconds { get; set; } = 90;
```
raman-m, line 20: Wrong intention to encapsulate some value which is not shared.
We usually define public constant! See examples: [search for "90"](https://github.com/search?q=repo%3AThreeMammals%2FOcelot+90+language%3AC%23&type=code&l=C%23)

### Returning type must be `Nullable`!

src/Ocelot/Requester/MessageInvokerPool.cs
```diff
@@ -10,13 +12,18 @@ public class MessageInvokerPool : IMessageInvokerPool
     private readonly IDelegatingHandlerHandlerFactory _handlerFactory;
     private readonly IOcelotLogger _logger;
 
-    public MessageInvokerPool(IDelegatingHandlerHandlerFactory handlerFactory, IOcelotLoggerFactory loggerFactory)
+    public MessageInvokerPool(
+        IDelegatingHandlerHandlerFactory handlerFactory,
+        IOcelotLoggerFactory loggerFactory,
+        IOptions<FileConfiguration> options)
```
raman-m, line 18: `IOptions<FileConfiguration> options`
Good approach to inject configuration model using Options pattern.
But why do you inject whole configuration if you need global value only?
Seems the code doesn't reuse other config options, so entire configuration is not required.

src/Ocelot/Requester/MessageInvokerPool.cs
```diff
@@ -31,17 +38,15 @@ public HttpMessageInvoker Get(DownstreamRoute downstreamRoute)
     }
 
     public void Clear() => _handlersPool.Clear();
-
-    /// <summary>
-    /// TODO This should be configurable and available as global config parameter in ocelot.json.
-    /// </summary>
-    public const int DefaultRequestTimeoutSeconds = 90;
+    
+    private readonly int _defaultRequestTimeoutSeconds;
```
raman-m, line 42: Seems the member should be converted to public property

src/Ocelot/Requester/MessageInvokerPool.cs
```diff
@@ -31,17 +38,15 @@ public HttpMessageInvoker Get(DownstreamRoute downstreamRoute)
...
+    
     private int _requestTimeoutSeconds;
 
     public int RequestTimeoutSeconds
     {
-        get => _requestTimeoutSeconds > 0 ? _requestTimeoutSeconds : DefaultRequestTimeoutSeconds;
-        set => _requestTimeoutSeconds = value > 0 ? value : DefaultRequestTimeoutSeconds;
+        get => _requestTimeoutSeconds > 0 ? _requestTimeoutSeconds : _defaultRequestTimeoutSeconds;
+        set => _requestTimeoutSeconds = value > 0 ? value : _defaultRequestTimeoutSeconds;
```
raman-m, line 49: These changes are not required if `DefaultRequestTimeoutSeconds` will be converted to a property.

test/Ocelot.UnitTests/Requester/MessageInvokerPoolTests.cs
```diff
@@ -27,12 +28,15 @@ public class MessageInvokerPoolTests : UnitTest
...
     private IWebHost _host;
+    private IOptions<FileConfiguration> _fileConfiguration;
 
     public MessageInvokerPoolTests()
     {
         _ocelotLoggerFactory = new Mock<IOcelotLoggerFactory>();
         _ocelotLogger = new Mock<IOcelotLogger>();
         _ocelotLoggerFactory.Setup(x => x.CreateLogger<MessageInvokerPool>()).Returns(_ocelotLogger.Object);
+        _fileConfiguration = new OptionsWrapper<FileConfiguration>(
+            new FileConfiguration { GlobalConfiguration = new FileGlobalConfiguration() });
```
raman-m, line 39: :ok: We will discuss tests later, because the code is not complete.
Not an issue for now!

raman-m (changes requested):
### Good start! :rocket: 
But could you work more plz?
Please pay attention also that you need to implement requirements of both issues: #1314 and #1869

src/Ocelot/Configuration/File/FileGlobalConfiguration.cs
hogwartsdeveloper: fixed

src/Ocelot/Requester/MessageInvokerPool.cs
hogwartsdeveloper: I did received the only FileGlobalConfiguration, but registered in DI?
How can I better get File GlobalConfiguration?

src/Ocelot/Requester/MessageInvokerPool.cs
hogwartsdeveloper: fixed

hogwartsdeveloper:
> Thank you for PR creation!
> 
> ### Good start! 🚀
> But could you work more plz? Please pay attention also that you need to implement requirements of both issues: #1314 and #1869

Thanks, I'll do this for both tasks

raman-m:
Please find some time to rebase branch onto develop and I'll review.
Do not forget to **Sync fork** to have [actual develop](https://github.com/hogwartsdeveloper/Ocelot/tree/develop) before rebasing.

src/Ocelot/Requester/MessageInvokerPool.cs
raman-m: After conducting some code research, I realized my mistake; the issue was not as significant as I thought. It's possible to inject the entire configuration object. The issue with `IOptions<T>` is that it's not utilized across all service classes currently, which indicates a design and development oversight.

`IOptions<FileConfiguration>` can be injected because we bind this `IConfiguration` section in `OcelotBuilder` at [line 55](https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/DependencyInjection/OcelotBuilder.cs#L55).

Alternatively, binding a new global configuration section would be preferable.

In conclusion, it's feasible to inject both `IConfiguration` and `IOptions<FileConfiguration>` at this point.
But having injected `IOptions<FileGlobalConfiguration>` is better.

raman-m (changes requested):
2nd round has finished! My suggestions are below.

raman-m:
Also we have to take into consideration this https://github.com/ThreeMammals/Ocelot/discussions/2072#discussioncomment-9469641 !
So, we have old `Timeout` property 👇 
https://github.com/ThreeMammals/Ocelot/blob/cc8f5c5dfc7c56e4446a2e80d97391429c8b364d/src/Ocelot/Configuration/File/FileRoute.cs#L75
We have to reuse it somehow. It's more relevant for #1869 ...

hogwartsdeveloper:
> Additionally, we must consider this [#2072 (comment)](https://github.com/ThreeMammals/Ocelot/discussions/2072#discussioncomment-9469641). We have an existing `Timeout` property as shown here:
> 
> https://github.com/ThreeMammals/Ocelot/blob/cc8f5c5dfc7c56e4446a2e80d97391429c8b364d/src/Ocelot/Configuration/File/FileRoute.cs#L75
> 
> 
> We need to find a way to repurpose it, as it is particularly pertinent to issue #1869

Ok, I found a solution, can you take a look?

src/Ocelot/Configuration/File/FileGlobalConfiguration.cs
```diff
@@ -17,6 +17,8 @@ public FileGlobalConfiguration()
 
         public string RequestIdKey { get; set; }
 
+        public int? RequestTimeoutSeconds { get; set; }
```
ggnaegi, line 20: I would recommend using `int` instead of `int?` and set the default value (90 seconds) in the constructor. So, if not explicitely modified in the global configuration or for a given route, then 90 seconds will be used as default value. It is quite handy to do so, we avoid references to the default timeout value anywhere else.

src/Ocelot/Requester/MessageInvokerPool.cs
```diff
@@ -31,11 +33,9 @@ public HttpMessageInvoker Get(DownstreamRoute downstreamRoute)
     }
 
     public void Clear() => _handlersPool.Clear();
-
-    /// <summary>
-    /// TODO This should be configurable and available as global config parameter in ocelot.json.
-    /// </summary>
+    
     public const int DefaultRequestTimeoutSeconds = 90;
```
ggnaegi, line 37: We could probably remove the `DefaultRequestTimeoutSeconds` from `MessageInvokerPool`

src/Ocelot/Requester/MessageInvokerPool.cs
```diff
@@ -58,9 +58,11 @@ private HttpMessageInvoker CreateMessageInvoker(DownstreamRoute downstreamRoute)
 
         // Adding timeout handler to the top of the chain.
         // It's standard behavior to throw TimeoutException after the defined timeout (90 seconds by default)
-        var timeoutHandler = new TimeoutDelegatingHandler(downstreamRoute.QosOptions.TimeoutValue == 0
-            ? TimeSpan.FromSeconds(RequestTimeoutSeconds)
-            : TimeSpan.FromMilliseconds(downstreamRoute.QosOptions.TimeoutValue))
+        var timeoutHandler = new TimeoutDelegatingHandler(downstreamRoute.QosOptions.TimeoutValue > 0
```
ggnaegi, line 61: I'm not sure about the logic here. The `ITimeoutCreator` will return you a timeout value, it includes the default timeout value too.

src/Ocelot/Configuration/Creator/TimeoutCreator.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Ocelot.Configuration.Creator
+{
+    public class TimeoutCreator : ITimeoutCreator
+    {
+        public int Create(FileRoute fileRoute, FileGlobalConfiguration globalConfiguration)
+        {
+            return fileRoute.Timeout > 0 
+                ? fileRoute.Timeout 
+                : (globalConfiguration.RequestTimeoutSeconds ?? 0) * 1000;
```
ggnaegi, line 11: After the changes, the `globalConfiguration.RequestTimeoutSeconds` will never be null. Maybe it would be wise to use a unique naming convention everywhere.

ggnaegi (changes requested):
Hello, thanks a lot, we should clarify the logic though. In my opinion, we should set a default global timeout value in the FileGlobalConfiguration and handle everything in the `ITimeoutCreator` implementation.

src/Ocelot/Configuration/File/FileGlobalConfiguration.cs
raman-m: @ggnaegi Gui, do you realize the File model is solely a class for JSON de(serialization)? All File models must be devoid of business logic; otherwise, testing them would seem odd to me. However, I concur that a centralized location for maintaining a public default value is necessary. This could be a separate class or shared logic, which would then be utilized in `MessageInvokerPool` and Polly QoS classes. But assigning the value directly in the File model is not something I agree with.

> using `int` instead of `int?`

What is the issue with this alteration? Is it because it necessitates a review of current implementations? I have requested the author to adopt `Nullable<T>`. My rationale is to initiate a new design convention that mirrors the behavior of reference-type properties for value types, aiming to reduce redundant and confusing logic when a value is missing. Moreover, JSON (de)serialization relies on nulls! `Nullable<T>` is an excellent enhancement for reducing code, although it initially appears that we are writing more code for null checks. However, that's simply the nature of it.

> set the default value (90 seconds) in the constructor.

I must express my disagreement here. Constructors are invoked by JSON (de)serialization managers, and to my knowledge, the default constructor is what's called. We must exercise caution when defining an explicit default constructor. My recommendation is to avoid defining explicit default constructors whenever possible.

src/Ocelot/Requester/MessageInvokerPool.cs
raman-m: I agree with this movement. We need to reference to the value of new shared class.
`GlobalPolicy`?

src/Ocelot.Provider.Polly/PollyQoSResiliencePipelineProvider.cs
```diff
@@ -48,11 +50,9 @@ private void PollyResiliencePipelineWrapperFactory(ResiliencePipelineBuilder<Htt
     {
         var options = route.QosOptions;
 
-        // Add TimeoutStrategy if TimeoutValue is not int.MaxValue and greater than 0
-        if (options.TimeoutValue != int.MaxValue && options.TimeoutValue > 0)
-        {
-            builder.AddTimeout(TimeSpan.FromMilliseconds(options.TimeoutValue));
-        }
+        builder.AddTimeout(options.TimeoutValue > 0
```
RaynaldM, line 53: not agree with that, if an user (like me) does not want to add a TimeOutStrategy, he can do it

src/Ocelot.Provider.Polly/v7/PollyQoSProvider.cs
```diff
@@ -60,12 +57,10 @@ private PollyPolicyWrapper<HttpResponseMessage> PollyPolicyWrapperFactory(Downst
 
         // No default set for polly timeout at the minute.
         // Since a user could potentially set timeout value = 0, we need to handle this case.
-        // TODO throw an exception if the user sets timeout value = 0 or at least return a warning
-        // TODO the design in DelegatingHandlerHandlerFactory should be reviewed
         var timeoutPolicy = Policy
-            .TimeoutAsync<HttpResponseMessage>(
-                TimeSpan.FromMilliseconds(route.QosOptions.TimeoutValue),
-                TimeoutStrategy.Pessimistic);
+            .TimeoutAsync<HttpResponseMessage>(route.QosOptions.TimeoutValue > 0
```
RaynaldM, line 61: useless, V7 will be remove in the next release 24, so, It will be better to change nothing to the behavior

src/Ocelot/Configuration/File/FileGlobalConfiguration.cs
ggnaegi: @raman, ok, you can keep `int?` but the default value must be then set in the `ITimeoutCreator` implementation.

RaynaldM (changes requested):

src/Ocelot/Requester/MessageInvokerPool.cs
raman-m: I apologize, but which `ITimeoutCreator` are you referring to? We don't have such an interface in our solution.
**Could you provide a code snippet for the author?**
Perhaps it would be more sensible to integrate the business logic into the existing `TimeoutDelegatingHandler` class, wouldn't it? If we modify this logic, wouldn't it necessitate a review of numerous tests?

In my view, the [TimeoutDelegatingHandler](https://github.com/search?q=repo%3AThreeMammals%2FOcelot%20TimeoutDelegatingHandler&type=code) is the ideal place to set the default 90-second timeout value and the new property. We also need to access the global shared value in this class.

My concern is that the `TimeoutDelegatingHandler` isn't a standard service class in DI, and since it's part of the private messaging system logic, it cannot be overridden. However, this is beyond the scope of the current linked issues.

src/Ocelot/Configuration/Creator/TimeoutCreator.cs
raman-m: We haven't decided on nullable types yet!
Seems we have the conflict of two design visions... Could we discuss this privately in Slack please?
We need a consensus here definitely! 

**Second**, Regarding creator class... It will be added to DI. Do you believe this micro service class should be in DI and overridden in theory? What's your vision of allowing this for end developers?
I'm just curious about possible user scenarios of overriding this creator?
**Pay attention we have end consumer class which is `TimeoutDelegatingHandler`❗** 

**Third**, both of you guys forgot about Polly QoS value 🤣 
I think we have 3 or even 4 values to merge somewhere:
- current `Timeout` at Route-level
- global timeout seconds for all routes
- Polly QoS `TimeoutValue` of QoS options
- default 90 seconds value

💥 Bingo!

src/Ocelot/Configuration/DownstreamRoute.cs
```diff
@@ -77,10 +78,11 @@ public class DownstreamRoute
...
             SecurityOptions = securityOptions;
             DownstreamHttpMethod = downstreamHttpMethod;
-            DownstreamHttpVersion = downstreamHttpVersion;
-            DownstreamHttpVersionPolicy = downstreamHttpVersionPolicy;
+            DownstreamHttpVersion = downstreamHttpVersion;
+            DownstreamHttpVersionPolicy = downstreamHttpVersionPolicy;
             UpstreamHeaders = upstreamHeaders ?? new();
-            MetadataOptions = metadataOptions;
+            MetadataOptions = metadataOptions;
+            Timeout = timeout;
```
raman-m, line 85: Yet another value propagation 🤕

src/Ocelot/DependencyInjection/OcelotBuilder.cs
```diff
@@ -74,6 +72,7 @@ public OcelotBuilder(IServiceCollection services, IConfiguration configurationRo
             Services.TryAddSingleton<IAuthenticationOptionsCreator, AuthenticationOptionsCreator>();
             Services.TryAddSingleton<IUpstreamTemplatePatternCreator, UpstreamTemplatePatternCreator>();
             Services.TryAddSingleton<IRequestIdKeyCreator, RequestIdKeyCreator>();
+            Services.TryAddSingleton<ITimeoutCreator, TimeoutCreator>();
```
raman-m, line 75: Who knows, maybe we will introduce new feature: Default timeout overriding 😉

raman-m (commented):
I will continue to make suggestions without the expectation of immediate implementation.
We are currently facing design conflicts... 😢 

@hogwartsdeveloper, please exercise patience! Our discussions may be lengthy, and delivering both linked features will be challenging. I am optimistic that we will reach a consensus on the current design and necessary refactoring.

At present, it is imperative to include this PR in the [current release](https://github.com/ThreeMammals/Ocelot/milestone/6). The community has been eagerly awaiting the "default timeout feature", which is a high-priority item. Therefore, I will assign this PR to the [Spring'24](https://github.com/ThreeMammals/Ocelot/milestone/6) milestone.

src/Ocelot/DependencyInjection/OcelotBuilder.cs
hogwartsdeveloper: > Who knows, maybe we will introduce new feature: Default timeout overriding 😉

I added)

src/Ocelot/Configuration/File/FileGlobalConfiguration.cs
raman-m: Good!
Default 90 secs constant must be defined in `TimeoutCreator` class or in end consumers (old logic).

Do you agree to use `Nullable<T>` type more for data models? If Yes, it will be very good approved standard/convention in our project.

src/Ocelot/Requester/MessageInvokerPool.cs
raman-m: #### Update 2
I overlooked the QoS user scenarios... Our existing logic uses two different constants. On the other hand, should the behavior of a standard route request (with a 90-second timeout) differ from the QoS default value? I realized today that the 90-second timeout for both traditional forwarding (without QoS) and QoS-enabled routes might need different values. A normal route without QoS is stable, so 90 seconds is acceptable. However, a QoS-enabled route is less stable due to high load, suggesting the timeout should be shorter. Since we define `TimeoutValue` in most cases, it could be reduced by two to three times.
@RaynaldM, what do you think? Should we maintain the 90-second default value for QoS when timeouts aren't specified? Although, this seems to be a very rare case.

src/Ocelot/DependencyInjection/OcelotBuilder.cs
raman-m: @RaynaldM @ggnaegi This is major issue in my opinion!
We need to have consensus here. Do we add it to DI or not to add?
Will we allow to developers to override default timeout logic or not to allow?

RaynaldM: I agree, but we have to be careful not to change expected behavior (Open-closed principle), unless it's really necessary.
The idea is that, if I update the package, it continues to work without any change on my side (unless there's a breaking change).

raman:
Would you stop tagging me?

src/Ocelot/Configuration/File/FileGlobalConfiguration.cs
raman: > @raman, ok, you can keep `int?` but the default value must be then set in the `ITimeoutCreator` implementation.

No.

raman-m: @ggnaegi  Why are you calling phone numbers in India? 
### 🤣

raman-m: @raman Hello! It's a pleasure to meet you! Your username is outstanding. 😉

src/Ocelot/DependencyInjection/OcelotBuilder.cs
raman-m: @hogwartsdeveloper @RaynaldM @ggnaegi 🆗 guys
Current votes are
- @raman-m ➖ disagree
- @ggnaegi ➕ agree, proposed `ITimeoutCreator` interface
- @RaynaldM ➕ agree

src/Ocelot/Configuration/Creator/ITimeoutCreator.cs
```diff
@@ -0,0 +1,9 @@
+using Ocelot.Configuration.File;
+
+namespace Ocelot.Configuration.Creator
+{
+    public interface ITimeoutCreator
+    {
+        int Create(FileRoute fileRoute, FileGlobalConfiguration globalConfiguration);
```
raman-m, line 7: @ggnaegi @hogwartsdeveloper 
Why not to consider to move this tiny business logic method to [IRoutesCreator](https://github.com/search?q=repo%3AThreeMammals%2FOcelot%20IRoutesCreator&type=code) interface like this:
```C#
int CreateTimeout(FileRoute fileRoute, FileGlobalConfiguration globalConfiguration);
```

src/Ocelot/Configuration/Creator/RoutesCreator.cs
```diff
@@ -21,8 +21,9 @@ public class RoutesCreator : IRoutesCreator
         private readonly IRouteKeyCreator _routeKeyCreator;
         private readonly ISecurityOptionsCreator _securityOptionsCreator;
         private readonly IVersionCreator _versionCreator;
-        private readonly IVersionPolicyCreator _versionPolicyCreator;
-        private readonly IMetadataCreator _metadataCreator;
+        private readonly IVersionPolicyCreator _versionPolicyCreator;
+        private readonly IMetadataCreator _metadataCreator;
+        private readonly ITimeoutCreator _timeoutCreator;
```
raman-m, line 26: We can consider new design approach
- don't extend params of the constructor, not injecting micro services 
- extend `IRoutesCreator` by new method
- implement new method in this class

Instead of creation of tons of micro-interfaces we can implement creator-logic in this class.
💥 Bingo!

src/Ocelot/Configuration/Creator/RoutesCreator.cs
```diff
@@ -153,9 +158,10 @@ private DownstreamRoute SetUpDownstreamRoute(FileRoute fileRoute, FileGlobalConf
                 .WithDangerousAcceptAnyServerCertificateValidator(fileRoute.DangerousAcceptAnyServerCertificateValidator)
                 .WithSecurityOptions(securityOptions)
                 .WithDownstreamHttpVersion(downstreamHttpVersion)
-                .WithDownstreamHttpVersionPolicy(downstreamHttpVersionPolicy)
+                .WithDownstreamHttpVersionPolicy(downstreamHttpVersionPolicy)
                 .WithDownStreamHttpMethod(fileRoute.DownstreamHttpMethod)
-                .WithMetadata(metadata)
+                .WithMetadata(metadata)
+                .WithTimeout(timeout)
```
raman-m, line 164:
```suggestion
                .WithTimeout(CreateTimeout(fileRoute, globalConfiguration))
```

src/Ocelot/Configuration/Creator/TimeoutCreator.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace Ocelot.Configuration.Creator
+{
+    public class TimeoutCreator : ITimeoutCreator
+    {
+        public int Create(FileRoute fileRoute, FileGlobalConfiguration globalConfiguration)
+        {
+            return fileRoute.Timeout > 0 
+                ? fileRoute.Timeout 
+                : globalConfiguration.RequestTimeoutSeconds * 1000;
+        }
```
raman-m, line 12: Move this method to `RoutesCreator` class!

src/Ocelot/Configuration/File/FileGlobalConfiguration.cs
```diff
@@ -13,10 +13,13 @@ public FileGlobalConfiguration()
             HttpHandlerOptions = new FileHttpHandlerOptions();
             CacheOptions = new FileCacheOptions();
             MetadataOptions = new FileMetadataOptions();
+            RequestTimeoutSeconds = 90;
         }
 
         public string RequestIdKey { get; set; }
 
+        public int RequestTimeoutSeconds { get; set; }
```
raman-m, line 21:
```suggestion
        public int TimeoutSeconds { get; set; }
```
and add XML-docs

src/Ocelot/DependencyInjection/OcelotBuilder.cs
raman-m: @hogwartsdeveloper
Please delete this line as we are phasing out `ITimeoutCreator` in favor of `IRoutesCreator`. 
The testing of logic will be conducted within the `RoutesCreator` class.

P.S. `IRoutesCreator` has already been incorporated into DI.

raman-m (changes requested):
@hogwartsdeveloper
Here's a summary of today's team discussions:
- Change all `int` return types to `int?` for the properties: `FileRoute.Timeout`, `FileGlobalConfiguration.TimeoutSeconds`, and `FileQoSOptions.TimeoutValue`. We've reached a consensus to make them nullable, which will enable us to write cleaner code without gaps in business logic.
- Relocate the `ITimeoutCreator.Create` method to the `IRoutesCreator` interface as the `CreateTimeout` method. The `ITimeoutCreator` micro-interface seems unnecessary with only one reference and one injection. Moreover, the implementation is one line of code. It will be tested as part of the `RoutesCreator` class.

### These are the essential changes required! ❗

src/Ocelot/Configuration/File/FileGlobalConfiguration.cs
```diff
@@ -17,6 +17,14 @@ public FileGlobalConfiguration()
 
         public string RequestIdKey { get; set; }
 
+        /// <summary>
+        /// The timeout in seconds for requests.
+        /// </summary>
+        /// <value>
+        /// The timeout value in seconds.
+        /// </value>
+        public int? TimeoutSeconds { get; set; }
```
ggnaegi, line 26: I would suggest using the same property names for the file global configuration and file route (eg. Timeout or TimeoutMs), otherwise it's misleading. Yes, I agree, I have used seconds in the MessageInvoker pool implementation, but it wasn't exposed to the user.

src/Ocelot/Configuration/File/FileQoSOptions.cs
```diff
@@ -32,6 +32,6 @@ public FileQoSOptions(QoSOptions from)
 
         public int DurationOfBreak { get; set; }
         public int ExceptionsAllowedBeforeBreaking { get; set; }
-        public int TimeoutValue { get; set; }
+        public int? TimeoutValue { get; set; }
```
ggnaegi, line 35: We might see some side effects in my opinion. Since, i'm not sure, but we are comparing TimeoutValue with zero somewhere.

src/Ocelot/Configuration/Creator/RoutesCreator.cs
```diff
@@ -76,6 +78,13 @@ public List<Route> Create(FileConfiguration fileConfiguration)
                 .ToList();
         }
 
+        public int CreateTimeout(FileRoute fileRoute, FileGlobalConfiguration globalConfiguration)
+        {
+            return fileRoute.Timeout > 0 
+                ? fileRoute.Timeout.Value 
+                : (globalConfiguration.TimeoutSeconds ?? DefaultRequestTimeoutSeconds) * 1000;
```
ggnaegi, line 85: Using ms everywhere is probably a better choice.

ggnaegi (changes requested):
@raman-m Thanks a a lot @hogwartsdeveloper, I think the QoS issues are not part of this current PR scope, we should address in a later PR. We are almost there, but it would be wise to use the same property names for the Timeout on the global configuration and on the route configuration, and with milliseconds precision. Some Acceptance tests are missing (Testing if the Global configuration Timeout is working as expected, that the QoS Timeout value as priority over the global / route Timeout value).

src/Ocelot.Provider.Polly/v7/PollyQoSProvider.cs
raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/2073/commits/e87010a9d36462e65dc06d358781f26545bc6898

src/Ocelot/Configuration/File/FileGlobalConfiguration.cs
raman-m: `Timeout` is better

src/Ocelot.Provider.Polly/PollyQoSResiliencePipelineProvider.cs
```diff
@@ -47,12 +49,15 @@ public ResiliencePipeline<HttpResponseMessage> GetResiliencePipeline(DownstreamR
...
-
-        // Add TimeoutStrategy if TimeoutValue is not int.MaxValue and greater than 0
+        
         if (options.TimeoutValue != int.MaxValue && options.TimeoutValue > 0)
         {
             builder.AddTimeout(TimeSpan.FromMilliseconds(options.TimeoutValue));
         }
+        else
+        {
+            builder.AddTimeout(TimeSpan.FromMilliseconds(route.Timeout));
```
RaynaldM, line 59: I still don't agree: route.timeOut is basically used to indicate the HttpClient timeout, which is different from Polly's own timeOut.
This means we can combine the 2.
As it is, I couldn't use this version.

src/Ocelot/Configuration/Creator/RoutesCreator.cs
```diff
@@ -21,8 +21,10 @@ public class RoutesCreator : IRoutesCreator
         private readonly IRouteKeyCreator _routeKeyCreator;
         private readonly ISecurityOptionsCreator _securityOptionsCreator;
         private readonly IVersionCreator _versionCreator;
-        private readonly IVersionPolicyCreator _versionPolicyCreator;
-        private readonly IMetadataCreator _metadataCreator;
+        private readonly IVersionPolicyCreator _versionPolicyCreator;
+        private readonly IMetadataCreator _metadataCreator;
+        
+        public const int DefaultRequestTimeoutSeconds = 90;
```
RaynaldM, line 27: to go all the way, this parameter should be a setting in globalConfiguration

src/Ocelot/Configuration/File/FileRoute.cs
```diff
@@ -55,77 +55,77 @@ public FileRoute(FileRoute from)
...
+        public FileQoSOptions QoSOptions { get; set; }
+        public FileRateLimitRule RateLimitOptions { get; set; }
+        public string RequestIdKey { get; set; }
+        public Dictionary<string, string> RouteClaimsRequirement { get; set; }
+        public bool RouteIsCaseSensitive { get; set; }
+        public FileSecurityOptions SecurityOptions { get; set; }
+        public string ServiceName { get; set; }
+        public string ServiceNamespace { get; set; }
+        public int? Timeout { get; set; }
+        public Dictionary<string, string> UpstreamHeaderTransform { get; set; }
```
RaynaldM, line 76:
```suggestion
        public IDictionary<string, string> UpstreamHeaderTransform { get; set; }

```

src/Ocelot/Configuration/File/FileRoute.cs
```diff
@@ -55,77 +55,77 @@ public FileRoute(FileRoute from)
...
-            to.DelegatingHandlers = new(from.DelegatingHandlers);
-            to.DownstreamHeaderTransform = new(from.DownstreamHeaderTransform);
-            to.DownstreamHostAndPorts = from.DownstreamHostAndPorts.Select(x => new FileHostAndPort(x)).ToList();
-            to.DownstreamHttpMethod = from.DownstreamHttpMethod;
-            to.DownstreamHttpVersion = from.DownstreamHttpVersion;
+        public int Priority { get; set; }
+        public FileQoSOptions QoSOptions { get; set; }
+        public FileRateLimitRule RateLimitOptions { get; set; }
+        public string RequestIdKey { get; set; }
+        public Dictionary<string, string> RouteClaimsRequirement { get; set; }
```
RaynaldM, line 70:
```suggestion
        public IDictionary<string, string> RouteClaimsRequirement { get; set; }

```

src/Ocelot/Configuration/File/FileRoute.cs
```diff
@@ -55,77 +55,77 @@ public FileRoute(FileRoute from)
...
+        public bool RouteIsCaseSensitive { get; set; }
+        public FileSecurityOptions SecurityOptions { get; set; }
+        public string ServiceName { get; set; }
+        public string ServiceNamespace { get; set; }
+        public int? Timeout { get; set; }
+        public Dictionary<string, string> UpstreamHeaderTransform { get; set; }
+        public string UpstreamHost { get; set; }
+        public List<string> UpstreamHttpMethod { get; set; }
```
RaynaldM, line 78:
```suggestion
        public IList<string> UpstreamHttpMethod { get; set; }

```

RaynaldM (changes requested):

src/Ocelot/Configuration/Creator/RoutesCreator.cs
raman-m: @ggnaegi Gui, having `ms` dimension for timeout logic will 
- confuse developers 
- introduce breaking change for current QoS 

2nd argument, the [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header spec has seconds as measurement unit, not milliseconds!
Also [Keep-Alive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive) header has seconds as a measurement unit.

So, I don't recommend to use milliseconds for timeouts.

P.S. But for QoS we use millisecs: [QoSOptions](https://ocelot.readthedocs.io/en/latest/search.html?q=QoSOptions). It's OK because QoS should have millisecs precision for settings.

src/Ocelot/Configuration/File/FileGlobalConfiguration.cs
raman-m: Disagree to use milliseconds.
Agree to use seconds! Read my [comment](https://github.com/ThreeMammals/Ocelot/pull/2073#discussion_r1622477480)

src/Ocelot.Provider.Polly/PollyQoSResiliencePipelineProvider.cs
raman-m: Rest assured, it won't be merged without your approval. 😄 
You're correct, @RaynaldM! We should not conflate the non-QoS `Timeout` with the QoS `TimeoutValue`.
Thus, if the `TimeoutValue` is defined, we should use it. If not, we should default to 90 seconds or a global value exclusively for QoS.

src/Ocelot/Configuration/Creator/RoutesCreator.cs
raman-m: Good proposal!
Good news that we already have global value in current settings:
https://github.com/ThreeMammals/Ocelot/blob/cc8f5c5dfc7c56e4446a2e80d97391429c8b364d/src/Ocelot/Configuration/File/FileGlobalConfiguration.cs#L24
But we don't use it! 🤯

src/Ocelot/Configuration/File/FileRoute.cs
raman-m: Not an issue.
It is beyond the scope of Timeout changes!

src/Ocelot/Configuration/File/FileRoute.cs
raman-m: ## 🤕

src/Ocelot/Configuration/File/FileRoute.cs
raman-m: ## 🙈

src/Ocelot.Provider.Polly/PollyQoSResiliencePipelineProvider.cs
raman-m: @hogwartsdeveloper this `else`-block must read the global value from
https://github.com/ThreeMammals/Ocelot/blob/cc8f5c5dfc7c56e4446a2e80d97391429c8b364d/src/Ocelot/Configuration/File/FileGlobalConfiguration.cs#L24
Don't forget about absolute default value which must be redefined in this provider.

src/Ocelot/Configuration/File/FileQoSOptions.cs
raman-m: @hogwartsdeveloper Propagate nullable everywhere up to end consumer, our provider.
Inside of provider add null check and substitute default value for QoS provider only.

src/Ocelot/Configuration/Creator/RoutesCreator.cs
raman-m: @RaynaldM We have [this](https://github.com/ThreeMammals/Ocelot/pull/2073/files#diff-95ae956905b7ba73232d0c91aaccc4f723aaa9eeb879210e718b62e43b847ca6R26) as global timeout value.
This 90sec constant is used when both route and global values are not defined!

hogwartsdeveloper:
@raman-m Write tests?)

raman-m:
Why did you push the code which cannot be compiled.
[Build error](https://app.circleci.com/pipelines/github/ThreeMammals/Ocelot/2221/workflows/c30c943c-f097-44f4-868e-e820bba30775/jobs/4137?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary&invite=true#step-102-42191_48)

> Write tests?)

Certainly! But I'm going to review finally...

src/Ocelot.Provider.Polly/PollyQoSResiliencePipelineProvider.cs
```diff
@@ -10,16 +12,23 @@ namespace Ocelot.Provider.Polly;
...
+    private readonly FileGlobalConfiguration _global;
+    
+    public const int DefaultRequestTimeoutSeconds = 40;
 
-    public PollyQoSResiliencePipelineProvider(IOcelotLoggerFactory loggerFactory, 
-        ResiliencePipelineRegistry<OcelotResiliencePipelineKey> resiliencePipelineRegistry)
+    public PollyQoSResiliencePipelineProvider(
+        IOcelotLoggerFactory loggerFactory, 
+        ResiliencePipelineRegistry<OcelotResiliencePipelineKey> resiliencePipelineRegistry,
+        IOptions<FileGlobalConfiguration> global)
```
raman-m, line 27: Good!
Added object binding to DI?

src/Ocelot.Provider.Polly/PollyQoSResiliencePipelineProvider.cs
```diff
@@ -10,16 +12,23 @@ namespace Ocelot.Provider.Polly;
...
 /// </summary>
-public class PollyQoSResiliencePipelineProvider : PollyQoSProviderBase, IPollyQoSResiliencePipelineProvider<HttpResponseMessage>
+public class PollyQoSResiliencePipelineProvider 
+    : PollyQoSProviderBase, IPollyQoSResiliencePipelineProvider<HttpResponseMessage>
 {
     private readonly ResiliencePipelineRegistry<OcelotResiliencePipelineKey> _resiliencePipelineRegistry;
     private readonly IOcelotLogger _logger;
+    private readonly FileGlobalConfiguration _global;
+    
+    public const int DefaultRequestTimeoutSeconds = 40;
```
raman-m, line 22:
```suggestion
    public const int DefaultRequestTimeoutMilliseconds = 40_000; // 40 seconds
```
See line 66!

src/Ocelot/Configuration/Creator/RoutesCreator.cs
```diff
@@ -76,6 +78,13 @@ public List<Route> Create(FileConfiguration fileConfiguration)
                 .ToList();
         }
 
+        public int CreateTimeout(FileRoute fileRoute, FileGlobalConfiguration globalConfiguration)
+        {
+            return fileRoute.Timeout > 0 
+                ? fileRoute.Timeout.Value 
+                : (globalConfiguration.Timeout ?? DefaultRequestTimeoutSeconds) * 1000;
```
raman-m, line 85:
```suggestion
                : globalConfiguration.Timeout ?? DefaultRequestTimeoutSeconds;
```
Both values ​​are in seconds!

src/Ocelot/Configuration/Creator/IRoutesCreator.cs
```diff
@@ -5,4 +5,6 @@ namespace Ocelot.Configuration.Creator;
 public interface IRoutesCreator
 {
     List<Route> Create(FileConfiguration fileConfiguration);
+    
+    int CreateTimeout(FileRoute fileRoute, FileGlobalConfiguration globalConfiguration);
```
raman-m, line 9:
```suggestion
    // Returns value in seconds!
    int CreateTimeout(FileRoute fileRoute, FileGlobalConfiguration globalConfiguration);
```

src/Ocelot/Configuration/DownstreamRoute.cs
```diff
@@ -128,8 +130,10 @@ public class DownstreamRoute
         ///   <item><see href="https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httprequestmessage.versionpolicy">HttpRequestMessage.VersionPolicy Property</see></item>
         /// </list>
         /// </remarks>
-        public HttpVersionPolicy DownstreamHttpVersionPolicy { get; }
+        public HttpVersionPolicy DownstreamHttpVersionPolicy { get; }
         public Dictionary<string, UpstreamHeaderTemplate> UpstreamHeaders { get; }
         public MetadataOptions MetadataOptions { get; }
+        
+        public int Timeout { get; }
```
raman-m, line 137:
```suggestion
        // TODO Add XML docs -> Returns timeout value in seconds.
        public int Timeout { get; }
```

src/Ocelot/Configuration/File/FileGlobalConfiguration.cs
```diff
@@ -17,6 +17,14 @@ public FileGlobalConfiguration()
 
         public string RequestIdKey { get; set; }
 
+        /// <summary>
+        /// The timeout in seconds for requests.
+        /// </summary>
+        /// <value>
+        /// The timeout value in seconds.
```
raman-m, line 24: Very well!

src/Ocelot/Configuration/File/FileRoute.cs
```diff
@@ -55,77 +55,78 @@ public FileRoute(FileRoute from)
...
+        public int Priority { get; set; }
+        public FileQoSOptions QoSOptions { get; set; }
+        public FileRateLimitRule RateLimitOptions { get; set; }
+        public string RequestIdKey { get; set; }
+        public Dictionary<string, string> RouteClaimsRequirement { get; set; }
+        public bool RouteIsCaseSensitive { get; set; }
+        public FileSecurityOptions SecurityOptions { get; set; }
+        public string ServiceName { get; set; }
+        public string ServiceNamespace { get; set; }
+        public int? Timeout { get; set; }
```
raman-m, line 75:
```suggestion
        // TODO Returns timeout value in seconds.
        public int? Timeout { get; set; }
```

src/Ocelot/DependencyInjection/OcelotBuilder.cs
```diff
@@ -53,7 +51,7 @@ public OcelotBuilder(IServiceCollection services, IConfiguration configurationRo
             Configuration = configurationRoot;
             Services = services;
             Services.Configure<FileConfiguration>(configurationRoot);
-
+
```
raman-m, line 54: Binding to global section is required ❗

raman-m (changes requested):
You can begin writing tests after implementing these suggestions. This way, the code will be quite comprehensive.

raman-m:
Zhannur,
Your PR is the final one pending for the current [release/milestone](https://github.com/ThreeMammals/Ocelot/milestone/6)❗

Could you please prepare for an intense session of pair programming this evening and tomorrow whole day? 🙏

I will help you for sure...
Please consider an extra DevOps to rebase the feature branch onto develop, if you have time. But it is not required, the history is pretty linear.

raman-m:
Hi @ggnaegi  Could you look into `MessageInvokerPoolTests` and review 2 failed tests please?
I can't get it why do they fail? See my try to fix: https://github.com/ThreeMammals/Ocelot/pull/2073/commits/a09760bdf0febae7273184770a0bc28bec8848e1 but with no luck 😢

raman-m:
@hogwartsdeveloper I'm going to rebase feature branch...

hogwartsdeveloper:
sorry I didn't have time this evening, I'll find time sorry

hogwartsdeveloper:
> @hogwartsdeveloper There is commit: [Merge remote-tracking branch 'origin/issues-1869' into issues-1869](https://github.com/ThreeMammals/Ocelot/pull/2073/commits/4e87d07d5d9bf1ad84944affefad8ce68081d1f6) Why was it committed? Was it conflict resolution, or something else?
> 
> Today I'm going to add another commit to develop, so seems rebasing your feature branch will be necessary, we will see...

it was a solution to the conflict

ggnaegi:
@raman @hogwartsdeveloper Are you sure about the rebase process, I can see some changes from other PR, such as Metadata or Version Policy? Otherwise, ok for me, the acceptance tests are still missing though

hogwartsdeveloper:
> @raman @hogwartsdeveloper Are you sure about the rebase process, I can see some changes from other PRs, such as Metadata or Version Policy? Otherwise, ok for me, the acceptance tests are still missing though. Acceptance tests for the default timeout, for the priority of the route timeout over global timeout and for the priority of the QoS timeout over route and global timeout should be foreseen imo.

Add tests to PollyQoSResiliencePipelineProvider add test too?

raman-m:
@hogwartsdeveloper Regarding unit tests...
I think we need tests in
- `MessageInvokerPoolTests`
- `PollyQoSResiliencePipelineProviderTests`

Regarding acceptance tests...
- `PollyQoSTests`
- I'm not sure which class is the best to write tests for `MessageInvokerPool`... Help, @ggnaegi! `ContentTests`❓

hogwartsdeveloper:
> @hogwartsdeveloper Regarding unit tests... I think we need tests in
> 
> * `MessageInvokerPoolTests`
> * `PollyQoSResiliencePipelineProviderTests`
> 
> Regarding acceptance tests...
> 
> * `PollyQoSTests`
> * I'm not sure which class is the best to write tests for `MessageInvokerPool`... Help, @ggnaegi! `ContentTests`❓

I added tests

- MessageInvokerPoolTests
- RoutesCreatorTests

you can check plz?

raman-m:
Hi @RaynaldM !
Are you OK if we do the following changes in lines 64-67
- branch [L64-L67](https://github.com/hogwartsdeveloper/Ocelot/blob/issues-1869/src/Ocelot.Provider.Polly/PollyQoSResiliencePipelineProvider.cs#L64-L67)
- PR [L64-L67](https://github.com/ThreeMammals/Ocelot/pull/2073/files#diff-a15546e836c83aa7eaeabbcb1661fb7254214dd72097d83f94611f56fb73a5d8R64-R67)
```C#
        var options = route.QosOptions;
        var timeout = options.TimeoutValue;
        if (timeout.HasValue && timeout.Value > 0)
        {
            builder.AddTimeout(TimeSpan.FromMilliseconds(timeout.Value));
        }
        else
        {
            builder.AddTimeout(TimeSpan.FromMilliseconds(_global.QoSOptions.TimeoutValue ?? DefaultQoSTimeoutMilliseconds));
        }
```
The change involves a new `else` block that initializes the TimeOut strategy with a global/default value, ensuring the TimeOut strategy is always enabled.
Looks good?

raman-m:
@ggnaegi If you have time, I would appreciate your review of the failed acceptance tests in the `Ocelot.AcceptanceTests.ServiceDiscovery` namespace. I've attempted to address them with the newly refactored code (refer to commit 02ab62a43e82be0917452ce2f83f1d2537e34674), but without success. The issues seem to be related to the Consul SD logic, the `IInternalConfiguration` interface, or QoS. Your fresh perspective might help identify the bug or any design flaws. Thank you!

P.S. I will continue with debugging and a deeper code analysis tomorrow...

raman-m:
### Attention ❗ ⚠️
@hogwartsdeveloper 
A new commit in the develop branch has caused numerous merge conflicts in the Polly project files. It's necessary to rebase the feature branch onto develop. I'm starting the rebase process now... your patience is appreciated!

raman-m:
@hogwartsdeveloper Don't push please❗
But if you want to commit more, then you have to delete branch, fetch all, checkout the branch locally 👉 
- `git switch develop`
- `git branch -D issues-1869`
- `git fetch --all`
- `git checkout issues-1869`

And you will be able to contribute more 😉

raman-m:
@ggnaegi @RaynaldM 
### Ready for final code review❗ 
- @RaynaldM You could pay attention to QoS changes &rarr; [PollyQoSResiliencePipelineProvider](https://github.com/ThreeMammals/Ocelot/pull/2073/files#diff-a15546e836c83aa7eaeabbcb1661fb7254214dd72097d83f94611f56fb73a5d8)
- @ggnaegi I need your help with broken acceptance tests related to Consul logic (total 3 failed tests) 👇 

image

raman-m:
Going to rebase onto develop...

RaynaldM (approved):

raman-m:
### Related to
- https://github.com/ThreeMammals/Ocelot/discussions/2072#discussioncomment-9469641

raman-m:
What's up, man?

raman-m:
Hello, @RaynaldM!
Your responsibility covers QoS & Polly. Could you please review the following?
- [src/Ocelot.Provider.Polly/PollyQoSResiliencePipelineProvider](https://github.com/ThreeMammals/Ocelot/pull/2073/files#diff-a15546e836c83aa7eaeabbcb1661fb7254214dd72097d83f94611f56fb73a5d8)

Additionally, if time permits, please examine the rest of the new code in the main library, `src/Ocelot/*`.

#### FYI Successor is your PR #2081❗

raman-m:
@ggnaegi, your primary focus should be on `MessageInvokerPool`. 😄
Kindly review it first at: [src/Ocelot/Requester/MessageInvokerPool](https://github.com/ThreeMammals/Ocelot/pull/2073/files#diff-dd34910b10381415777bc5516395102cece2d6fe6b98d22203f6e20a41d607a6).
The key method here is `DownstreamRoute.TimeoutMilliseconds()`, which contains all the logic for the Default Timeout feature. Please make sure to review all the changes!

src/Ocelot/Requester/MessageInvokerPool.cs
```diff
@@ -56,11 +46,13 @@ private HttpMessageInvoker CreateMessageInvoker(DownstreamRoute downstreamRoute)
             baseHandler = delegatingHandler;
         }
 
+        _timeoutMilliseconds ??= downstreamRoute.TimeoutMilliseconds();
```
ggnaegi, line 49: Why use a private variable here? Once it's set for an arbitrary route, then you will not be able to change it. Imagine that you have two downstream routes with different Timeout values... Could you review that please?

ggnaegi (changes requested):
We have an issue in the message invoker pool. Only one timeout value will be taken into account, the one of the downstream route called first. So please review that first

src/Ocelot/Requester/MessageInvokerPool.cs
raman-m: OK... Could you read [this diff](https://github.com/ThreeMammals/Ocelot/pull/2073/files#diff-dd34910b10381415777bc5516395102cece2d6fe6b98d22203f6e20a41d607a6) once again plz?
What I've done is:
```diff
- private int _requestTimeoutSeconds;
+ private int? _timeoutMilliseconds = null;
```
Let me know what is the cardinal change? I don't see any cardinal changes: only different member name.
If you wish I can rename the member to the old name 🤣 
So, asking me `Why use a private variable here?` a bit late because we've release 2-3 releases. 😄 So, why our tests don't fail?
Answering question: "It is used because of life-hack reflection logic for old acceptance test which ensures default time used" &rarr;  see [GivenIHackDefaultTimeoutValue](https://github.com/ThreeMammals/Ocelot/pull/2073/files#diff-9ac3fd8293493ed683c838dec776a103e81d9a5af9b948205a09c3e1f668181aL201-R204) method

raman-m: > So please review that first Keep in mind that the Message Invoker pool is a singleton guys!

Singletons! 😭

raman-m: >Once it's set for an arbitrary route, then you will not be able to change it. Imagine that you have two downstream routes with different Timeout values...

Good catch! You are right! You are Architect! 😄 

> Could you review that please?

Yes, sure!.. going to find another hacking solution for the test...

raman-m:
### TODO
- [ ] Work on fake changes

src/Ocelot/Requester/MessageInvokerPool.cs
ggnaegi: @raman-m please breathe! 😆

raman-m: @ggnaegi  I'm breathing... :triumph: and breathing more if Gui gives an approval 😍 

> build [4444](https://app.circleci.com/pipelines/github/ThreeMammals/Ocelot/2482/workflows/8933d0d3-0380-4df5-a99e-28a822bcf153/jobs/4444)

WTF numbers❓

raman-m:
### TODO
- [ ] Write major acceptance tests for linked issues &rarr; #1314 and #1869
- [ ] Optionally add more unit tests and check code coverage 
- [ ] Write docs: explain global static default timeout constants, and how to configure global timeouts

raman-m:
### TODO
- [ ] Rebase the branch due to merge conflicts

raman-m:
### TODO
- [ ] Write major acceptance tests for linked issues &rarr; #1314 and #1869
- [ ] Optionally add more unit tests and check code coverage 
- [ ] Write docs: explain global static default timeout constants, and how to configure global timeouts

src/Ocelot/Requester/MessageInvokerPool.cs
raman-m: Ok @ggnaegi, the service is still [a singleton](https://github.com/hogwartsdeveloper/Ocelot/blob/issues-1869/src/Ocelot/Requester/ServiceCollectionExtensions.cs#L10) 👇 
https://github.com/ThreeMammals/Ocelot/blob/8143d5608f354262d9a2a08e12baf36ef5e49eb0/src/Ocelot/Requester/ServiceCollectionExtensions.cs#L10

 and, it remains [a singleton](https://github.com/hogwartsdeveloper/Ocelot/blob/issues-1869/src/Ocelot/Requester/ServiceCollectionExtensions.cs#L10).

The original issue with the encapsulated default value has been resolved, so please review the current [diff for MessageInvokerPool](https://github.com/ThreeMammals/Ocelot/pull/2073/files#diff-dd34910b10381415777bc5516395102cece2d6fe6b98d22203f6e20a41d607a6) 🙏, where the `public constant DefaultRequestTimeoutSeconds = 90;` has been removed.
Currently, [CreateMessageInvoker correctly reads timeout values directly from DownstreamRoute](https://github.com/ThreeMammals/Ocelot/pull/2073/files#diff-dd34910b10381415777bc5516395102cece2d6fe6b98d22203f6e20a41d607a6R47) when creating the route's `HttpMessageInvoker` for caching. As a result, each cached object will have the correct timeout encapsulated within [TimeoutDelegatingHandler](https://github.com/ThreeMammals/Ocelot/pull/2073/files#diff-dd34910b10381415777bc5516395102cece2d6fe6b98d22203f6e20a41d607a6R52). The old, [inefficient testing logic has been completely removed](https://github.com/ThreeMammals/Ocelot/pull/2073/files#diff-dd34910b10381415777bc5516395102cece2d6fe6b98d22203f6e20a41d607a6L38-L45) from `MessageInvokerPool`, and the current timeout logic has zero impact on pooling in the singleton.

I believe the issue has been resolved—**do you agree**?

**FYI**: All timeout-related logic resides in the `DownstreamRoute` class as the [TimeoutMilliseconds() business-helper](https://github.com/ThreeMammals/Ocelot/pull/2073/files#diff-91f77268e2513800fea3711da0f151cc5563af89391413161726e1eb6e29fe83R150-R152). Thus, the default timeout is managed by the [DownstreamRoute.DefaultTimeoutSeconds](https://github.com/ThreeMammals/Ocelot/pull/2073/files#diff-91f77268e2513800fea3711da0f151cc5563af89391413161726e1eb6e29fe83R144) static property.
So, a developer can easily provide the setting based on their needs. It seems we need to apply a value constraint in the setter without breaking the behavior of the target `TimeoutDelegatingHandler`.

raman-m: Added setter constraint to check a `value >= 3`, see commit https://github.com/ThreeMammals/Ocelot/pull/2073/commits/b22743410dad80b564ce50bccacaea006235c83a

raman-m:
### TODO
- [ ] Write major acceptance tests for linked issues &rarr; #1314 and #1869
- [ ] Optionally add more unit tests and check code coverage 
- [ ] Write docs: explain global static default timeout constants, and how to configure global timeouts
- [ ] @ggnaegi [suggestions](https://github.com/ThreeMammals/Ocelot/pull/2073#pullrequestreview-2087596632) related to testing:
  >Some Acceptance tests are missing (Testing if the Global configuration Timeout is working as expected, that the QoS Timeout value has priority over the global / route Timeout value).

raman-m:
@ggnaegi @RaynaldM Hello from important pull request!
I have re-requested your review as this is a [![label: highest][~highest]](https://github.com/ThreeMammals/Ocelot/labels/highest) priority matter and the FIFO PR for the current milestone. I would greatly appreciate it if you could review this piece of coding work once more. I believe main development phase has been completed.

FYI, In the coming days, I plan to focus on user scenario tests, reviewing the existing tests, and writing documentation for the current code. So, during these days I am ready to fix something, and work on your suggestions 😉 

[~highest]: https://img.shields.io/badge/-highest-D93F0B.svg

Hello, @hogwartsdeveloper!
Do you still wish to contribute further to the project? Were you neither tired nor shocked?
I believe your PR will be merged within 2-3 days, I certainly hope so.

src/Ocelot/Configuration/File/FileRoute.cs
```diff
@@ -85,10 +85,20 @@ public FileRoute(FileRoute from)
...
+    ///   <item><see cref="DownstreamRoute.DefaultTimeoutSeconds"/></item>
+    /// </list>
+    /// </remarks>
+    /// <value>A <see cref="Nullable{T}"/> (T is <see cref="int"/>) value, in seconds.</value>
+    public int? Timeout { get; set; }
     public IDictionary<string, string> UpstreamHeaderTemplates { get; set; }
-    public Dictionary<string, string> UpstreamHeaderTransform { get; set; }
+    public IDictionary<string, string> UpstreamHeaderTransform { get; set; }
     public string UpstreamHost { get; set; }
     public List<string> UpstreamHttpMethod { get; set; }
```
RaynaldM, line 103: IList or IReadOnlyList should be better

src/Ocelot/Configuration/Creator/IRoutesCreator.cs
```diff
@@ -5,4 +5,12 @@ namespace Ocelot.Configuration.Creator;
 public interface IRoutesCreator
 {
     List<Route> Create(FileConfiguration fileConfiguration);
```
RaynaldM: if possible, use IList (or better, IReadOnlyList), as it's cleaner to use abstractions in interfaces

src/Ocelot/Configuration/Creator/IDynamicsCreator.cs
```diff
@@ -5,4 +5,12 @@ namespace Ocelot.Configuration.Creator;
 public interface IDynamicsCreator
 {
     List<Route> Create(FileConfiguration fileConfiguration);
```
RaynaldM: if possible, use IList (or better, IReadOnlyList), as it's cleaner to use abstractions in interfaces

src/Ocelot.Provider.Polly/PollyQoSResiliencePipelineProvider.cs
```diff
@@ -15,13 +17,16 @@ public class PollyQoSResiliencePipelineProvider : IPollyQoSResiliencePipelinePro
 {
     private readonly ResiliencePipelineRegistry<OcelotResiliencePipelineKey> _registry;
     private readonly IOcelotLogger _logger;
-
+    private readonly FileGlobalConfiguration _global;
```
RaynaldM:
```suggestion
    private readonly FileGlobalConfiguration _fileGlobalConfiguration;
```

src/Ocelot/Configuration/File/FileRoute.cs
raman-m: To `IList<KeyValuePair<string, string>>`? 😄 Why ❓ 
Disagree, sorry, because of File-model type.
Impossible to convert from dictionary to generic list. Here is `IDictionary<TK, TV>`:
```csharp
public interface IDictionary<TKey, TValue> : ICollection<KeyValuePair<TKey, TValue>>, IEnumerable<KeyValuePair<TKey, TValue>>, IEnumerable
```
The conversion is only possible to `ICollection<KeyValuePair<TKey, TValue>>`.

**And did you forget about your [previous suggestion](https://github.com/ThreeMammals/Ocelot/pull/2073#discussion_r1622329912) to convert to `IDictionary<TKey, TValue>`?** What's the goal of applying higher and higher level of abstraction here?
And, once more, what is the goal of refactoring File-models? They are not business models! We do not test File-models at all! We test business models like [QoSOptions](https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/Configuration/QoSOptions.cs), but not [FileQoSOptions](https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/Configuration/File/FileQoSOptions.cs) ❗

src/Ocelot/Configuration/Creator/IRoutesCreator.cs
raman-m: That was previously added by Tom in this commit https://github.com/ThreeMammals/Ocelot/commit/44f8e312a851a763126516736916fabbbbad0630#diff-e0efc616bfa2708ab9e2e42168a6fa870f1969e93d585e60b1d510eef50b24ba 
I've just renamed the interface. So, your suggestion is not related to timeouts, but if you wish I will change this returning type right today.

**FYI**
- `IReadOnlyList<T>` has indexer only &rarr; `T this[int index] { get; }`
- and `IReadOnlyCollection<T>` has count &rarr; `int Count { get; }`

Finally, I agree to define `IReadOnlyList<T>` to inform community that we use Lists. But keep in mind: it may require cascading changes in consumer classes.

src/Ocelot.Provider.Polly/PollyQoSResiliencePipelineProvider.cs
raman-m: Goal? Having longer names? Or is not the code readability good?

Fixed in https://github.com/ThreeMammals/Ocelot/pull/2073/commits/69ed9bec79fd835901cd4c5c7b52d799ff12f7a6 ✔️

src/Ocelot/Configuration/Creator/IDynamicsCreator.cs
raman-m: Fixed in https://github.com/ThreeMammals/Ocelot/pull/2073/commits/69ed9bec79fd835901cd4c5c7b52d799ff12f7a6 ✔️ 

We have made some cascading changes, together :)

src/Ocelot/Configuration/Creator/IRoutesCreator.cs
raman-m: Fixed in https://github.com/ThreeMammals/Ocelot/pull/2073/commits/69ed9bec79fd835901cd4c5c7b52d799ff12f7a6 ✔️

src/Ocelot/Configuration/File/FileRoute.cs
raman-m: This issue is disregarded as [it has already been resolved](https://github.com/ThreeMammals/Ocelot/pull/2073#discussion_r1622329912).

src/Ocelot/Requester/MessageInvokerPool.cs
```diff
@@ -56,11 +44,14 @@ private HttpMessageInvoker CreateMessageInvoker(DownstreamRoute downstreamRoute)
             baseHandler = delegatingHandler;
         }
 
+        int milliseconds = route.QosOptions.UseQos
+            ? route.QosOptions.TimeoutValue ?? DownstreamRoute.DefaultTimeoutSeconds // TODO Discuss this: seems we should not override timeout here because Polly Timeout strategy is responsible for this
+            : 1000 * (route.Timeout ?? DownstreamRoute.DefaultTimeoutSeconds);
+        var timeout = TimeSpan.FromMilliseconds(milliseconds);
```
raman-m: @RaynaldM, I would like to draw your attention to this TODO comment at line 48. It appears that Tom may have made a design mistake by proposing to set up the HTTP client with QoS options.
I'm confident that the responsibility for using QoS lies solely with the consumer library that acts as the `Ocelot.Provider.Polly` provider. **Therefore, only the Polly [Timeout strategy](https://www.pollydocs.org/strategies/timeout.html) should consume `QosOptions`.**

So, there are 2 user scenarios
- The Route timeout exceeds the QoS timeout. Therefore, everything should function correctly the Timeout strategy will manage all timeout events.
- When the route timeout is shorter than the QoS timeout, the [TimeoutDelegatingHandler](https://github.com/search?q=repo%3AThreeMammals%2FOcelot%20TimeoutDelegatingHandler&type=code) will trigger an exception, which Ocelot will process as a 499 status code. Consequently, the Polly Timeout strategy will not respond, leading to the overall request failing and potentially resulting in multiple errors being logged or traced.

What do you think?
Consequently, I believe the QoS timeout should be less than the Route timeout, with the ratio of QoS Timeout to Route ranging from 0 to 0.9. Ratios of 0.95 or higher appear to result in conflicts and critical failures.
I propose to remove this design gap right from the `MessageInvokerPool` class.

My final proposal:
```diff
-        int milliseconds = route.QosOptions.UseQos
-            ? route.QosOptions.TimeoutValue ?? DownstreamRoute.DefaultTimeoutSeconds // TODO Discuss this: seems we should not override timeout here because Polly Timeout strategy is responsible for this
-            : 1000 * (route.Timeout ?? DownstreamRoute.DefaultTimeoutSeconds);
+        int milliseconds = 1000 * (route.Timeout ?? DownstreamRoute.DefaultTimeoutSeconds);
```
as a minimum follow up action.
I propose implementing a simple validation mechanism that informs the user through a warning: log a warning if `1000 * route.Timeout < route.QosOptions.Timeout`
For sure, approval from @ggnaegi is required before proceeding.
Therefore, is essential to reach team's consensus in order to proceed with this refactoring

raman-m (commented):
@RaynaldM @ggnaegi  I need your attention to [this](https://github.com/ThreeMammals/Ocelot/pull/2073#discussion_r2161460165) ☝️

src/Ocelot/Requester/MessageInvokerPool.cs
raman-m: Implemented this idea in commit https://github.com/ThreeMammals/Ocelot/pull/2073/commits/aed8c944c6c63c0fc5c07efea72d98fa695f5643 ❕

ggnaegi:
> @ggnaegi Since you [haven’t responded](https://github.com/ThreeMammals/Ocelot/pull/2073#pullrequestreview-2949835973), I can’t wait for your approval. However, I’ve addressed all the suggestions from your previous code reviews ([review1](https://github.com/ThreeMammals/Ocelot/pull/2073#pullrequestreview-2084535534), [review2](https://github.com/ThreeMammals/Ocelot/pull/2073#pullrequestreview-2087596632), [review3](https://github.com/ThreeMammals/Ocelot/pull/2073#pullrequestreview-2373330500)). If anything critical comes up, we’ll handle it in the next PRs.

@raman-m Yep, very busy at the moment...

src/Ocelot/Requester/MessageInvokerPool.cs
ggnaegi: @raman-m I think, at the end we shouldn't mix the two concepts and use only one. From my side, I would just remove the DefaultTimeoutSeconds and provide default QosOptions and describe that in the documentation.

raman-m: @ggnaegi You're right, but the issue is that the Ocelot.Provider.Polly package acts as a consumer of the Ocelot package (has a reference to it). Therefore, Ocelot Core shouldn't have any knowledge of Polly's functionality. In theory, Ocelot Core would need its own QoS implementation independent of Polly.

> From my side, I would just remove the DefaultTimeoutSeconds and provide default QosOptions and describe that in the documentation.

Yes, you would. But what if the user doesn’t use Polly provider QoS? 😄 For C# users, `DefaultTimeoutSeconds` is known as the "[absolute timeout](https://ocelot.readthedocs.io/en/develop/features/configuration.html#default-timeout#:~:text=absolute%20timeout)" feature. Unfortunately, we’ve had to abstractly combine both approaches, and to resolve this, we’re hot-fixing the timeout for `TimeoutDelegatingHandler` to unblock Polly behavior. To be honest, this issue currently exists in release 24.0. Additionally, users have reported unacceptable behavior from Polly, which I believe happens when the route timeout is shorter than the Polly timeout strategy.

PR #2082, opened 2024-05-30
Update "Getting Started" docs

GeeSuth:
As new to Ocelot I found that convinced to understand when you mention minimal API and code seems not.

## Docs
  - [Getting Started](https://ocelot.readthedocs.io/en/latest/introduction/gettingstarted.html)


raman-m:
Hello, Abdallh!
I'd accept this PR to update docs if you update [Basic sample](https://github.com/ThreeMammals/Ocelot/tree/develop/samples/Basic) too.

raman-m:
> As new to Ocelot I found that convinced to understand when you **mention minimal API and code seems not**.

The documentation could be divided into two subsections: minimal API and non-minimal. If a minimal API design is not feasible, I would suggest omitting the phrase "minimal API" entirely from the documentation.

raman-m (changes requested):
I'm confused about this PR. Are you intending to update the documentation, or are you looking to update the Basic sample?

I hope your confusion regarding Minimal API app has been cleared up now.
However, this PR cannot be merged as it currently stands due to its quality.

raman-m:
You could return to us with more refined code in the feature branch.

PR #2086, opened 2024-06-03
#2085 Enforce Polly v8 Circuit Breaker as the primary strategy

### LINKED ISSUE #2085: Circuit Breaker behavior in Polly v7 vs v8
### Discussed in https://github.com/ThreeMammals/Ocelot/discussions/2083

<div type='discussions-op-text'>

<sup>Originally posted by **anonimousspider** June  1, 2024</sup>
**In Polly v8, we have observed that the circuit breaker functionality is not working as expected. We have experienced issues where the circuit breaker does not trip or reset correctly, leading to inconsistent fault handling. Conversely, in Polly v7, the circuit breaker also exhibits problems, failing to operate smoothly in certain scenarios. 
Circuit Breaker Not Tripping** 

### Example
Suppose you have configured an Ocelot route with a circuit breaker policy to trip after three consecutive HTTP request failures:
Route Configuration:
```json
{
  "Routes": [
    {
      "DownstreamPathTemplate": "/api/StudentIdentity/Verify ",
      "DownstreamScheme": "https",
      "DownstreamHostAndPorts": [
        {
          "Host": "localhost", "Port": 5001
        }
      ],
      "UpstreamPathTemplate": " api/student/Verify ",
      "QoSOptions": {
        "ExceptionsAllowedBeforeBreaking": 3,
        "DurationOfBreak": 10000,
        "TimeoutValue": 5000
      }
    }
  ]
}
```
Configuration in Program.cs for V7:
```C#
services.AddLogging(logging =>
 {
     logging.ClearProviders();
     logging.AddConsole();
                            
 });
 services.AddJwtAuthentication();
 services.AddOcelot().AddPollyV7();
```
Configuration in program.cs using Polly V8:
```C#
services.AddLogging(logging =>
 {
     logging.ClearProviders();
     logging.AddConsole();
                            
 });
 services.AddJwtAuthentication();
 services.AddOcelot().AddPolly();
```

### Expected Behaviour
During runtime, the downstream service fails five consecutive times. The circuit breaker should trip and stop the request to be forwarded , and move to open state as per configured. 

### Actual behavior
During runtime, the downstream service fails five consecutive times. Despite this, the circuit breaker does not trip and continues to forward requests, instead of moving to the open state after three failures as configured.

### Scenario 1
When service response is delayed from the service configured in ocelot.
**PollyV7:** Its working fine 
**PollyV8:** on applying pollyv8 in the gateway not Working.

### Scenario 2
When service is giving 503 error response .
**Polly v7** : Its working Fine 
**Polly v8:** Not working.

### Scenario 3
When Service is giving 404 error response.
**Polly v7 :** working fine.
**Polly v8:** Not working.

---
Can you please suggest, how to configure Polly v8 with the current Ocelot version 23.2.2?
</div>

RaynaldM:
### Fixes #2085 
- #2085 

### Discussion
- #2083 

### Proposed Changes
- Invert CircuitBreaker (first) and TimeOut (second)  be exactly the same as in V7 ([become order is important](https://www.pollydocs.org/migration-v8.html#policy-wrap-in-v8))
- Add more logs (on HalfOpened, Closed for circuit-breaker and on Timeout)
- Create 2 unit tests to improve TimeOut strategy


raman-m:
Let me check what acceptance tests do we have now...

raman-m:
I have no good clarity on this
https://github.com/ThreeMammals/Ocelot/blob/a034e8c1e3fc23a086ad10000c85615b9696a43e/src/Ocelot.Provider.Polly/PollyQoSResiliencePipelineProvider.cs#L36
What does the condition `options.TimeoutValue is int.MaxValue` check actually? Why max value?
Based on other line 52
https://github.com/ThreeMammals/Ocelot/blob/a034e8c1e3fc23a086ad10000c85615b9696a43e/src/Ocelot.Provider.Polly/PollyQoSResiliencePipelineProvider.cs#L52
I expect that line 36 should check existence of the option, right? So, I expect this check 👉 
```c#
// this means the option does NOT exist
options.TimeoutValue == int.MaxValue || options.TimeoutValue == 0
```
Am I right, @RaynaldM ?

raman-m:
Our official acceptance testing for Polly provider:
https://github.com/ThreeMammals/Ocelot/blob/a034e8c1e3fc23a086ad10000c85615b9696a43e/test/Ocelot.AcceptanceTests/PollyQoSTests.cs#L9
Reviewing...

src/Ocelot.Provider.Polly/PollyQoSResiliencePipelineProvider.cs
```diff
@@ -39,46 +61,65 @@ public ResiliencePipeline<HttpResponseMessage> GetResiliencePipeline(DownstreamR
...
 
-    private void PollyResiliencePipelineWrapperFactory(ResiliencePipelineBuilder<HttpResponseMessage> builder, DownstreamRoute route)
+    protected virtual void PollyResiliencePipelineWrapperFactory(ResiliencePipelineBuilder<HttpResponseMessage> builder, DownstreamRoute route)
     {
         var options = route.QosOptions;
 
-        // Add TimeoutStrategy if TimeoutValue is not int.MaxValue and greater than 0
-        if (options.TimeoutValue != int.MaxValue && options.TimeoutValue > 0)
+        // Add CircuitBreaker strategy only if ExceptionsAllowedBeforeBreaking is greater than 2
+        if (options.ExceptionsAllowedBeforeBreaking >= 2)
```
ggnaegi, line 74: Why not use the shortcut, `options.ExceptionsAllowedBeforeBreaking < 2` and return?

ggnaegi (approved):

src/Ocelot.Provider.Polly/PollyQoSResiliencePipelineProvider.cs
raman-m: We can't shortcut here because we must care about TimeOut strategy also: see [line 112](https://github.com/ThreeMammals/Ocelot/pull/2086/files#diff-a15546e836c83aa7eaeabbcb1661fb7254214dd72097d83f94611f56fb73a5d8R112)

The workflow is the following: if option exists, apply appropriate Polly strategy.
So, we have 2 major strategies now:
- 1st is CircuitBreaker 
- 2nd is TimeOut

Both of them have appropriate option and `if`-block.

ggnaegi: ok, maybe, because of the code block size to foresee a method and place the logic in this method.

raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/2086/commits/cfcb772d17c607634cee1919dbd717f523634ff2 ✔️

raman-m (approved):
We're ready to merge!
@RaynaldM, if the design meets your approval, feel free to hit the magical "**Squash and Merge**" button.

PR #1155, opened 2020-03-15
Add ability to register custom load balancers

DavidLievrouw:
New Feature #961

Recreation of #962 after request from @TomPallister.

## Proposed Changes

  - Add ability to register custom ILoadBalancer implementations during DI composition.
  - Refactor the LoadBalancerFactory class, so that it can create more than only the built-in load balancers.
  - Extend the IOcelotBuilder with AddCustomLoadBalancer() helper methods.

PR #2111, opened 2024-06-26
#2110 Review load balancing and independent fetching the list of services in `Kube` provider

### LINKED ISSUE #2110: After Upgrade to 23.3.3 from 23.2.2 about every 20th - 30th request finished with status code 500
Resolving a service address to route an API request

Sometimes, about every 20th - 30th request finished with status code 500

```
requestId: 0HN4ISPBSIV4V:000037FF, previousRequestId: No PreviousRequestId, 
message: 'Exception caught in global error handler, 
exception message: Object reference not set to an instance of an object.,
exception stack:
   at Ocelot.LoadBalancer.LoadBalancers.RoundRobin.Lease(HttpContext httpContext)
   at Ocelot.LoadBalancer.Middleware.LoadBalancingMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.DownstreamPathManipulation.Middleware.ClaimsToDownstreamPathMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.QueryStrings.Middleware.ClaimsToQueryStringMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.Headers.Middleware.ClaimsToHeadersMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.Authorization.Middleware.AuthorizationMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.Claims.Middleware.ClaimsToClaimsMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.Authentication.Middleware.AuthenticationMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.RequestId.Middleware.RequestIdMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.RateLimiting.Middleware.RateLimitingMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.Request.Middleware.DownstreamRequestInitialiserMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.Headers.Middleware.HttpHeadersTransformationMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.Security.Middleware.SecurityMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.Multiplexer.MultiplexingMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.DownstreamRouteFinder.Middleware.DownstreamRouteFinderMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.Responder.Middleware.ResponderMiddleware.Invoke(HttpContext httpContext)
   at Ocelot.Errors.Middleware.ExceptionHandlerMiddleware.Invoke(HttpContext httpContext) RequestId: 0HN4ISPBSIV4V:000037FF'
```

```json
{
  "Routes": [
    {
      "UpstreamPathTemplate": "/public/{everything}",
      "UpstreamHttpMethod": [
        "POST"
      ],
      "ServiceName": "serviceName",
      "DownstreamPathTemplate": "/public/{everything}",
      "DownstreamScheme": "http",
      "LoadBalancerOptions": {
        "Type": "RoundRobin"
      }
    }
  ],
  "GlobalConfiguration": {
    "ServiceDiscoveryProvider": {
      "Namespace": "namespace",
      "Type": "Kube"
    }
}
}
```

## Specifications

  - Version: 23.3.3 (NET6 and NET8)
  - Platform: Linux
  - Subsystem:

antikorol:
## Fixes #2110 
* [After Upgrade to 23.3.3 from 23.2.2 about every 20th - 30th request finished with status code 500 #2110](https://github.com/ThreeMammals/Ocelot/issues/2110)

## Proposed Changes

Move the creation of the services list from the class field to the method, to prevent list modification from different threads, which caused the partially modified list of services to be returned.


src/Ocelot.Provider.Kubernetes/Kube.cs
```diff
@@ -35,17 +33,14 @@ public virtual async Task<List<Service>> GetAsync()
...
-        else
+        if ((endpoint?.Subsets?.Count ?? 0) == 0)
         {
             _logger.LogWarning(() => $"K8s Namespace:{_configuration.KubeNamespace}, Service:{_configuration.KeyOfServiceInK8s}; Unable to use: it is invalid. Address must contain host only e.g. localhost and port must be greater than 0!");
+            return new(0);
         }
 
-        return _services;
+        return BuildServices(_configuration, endpoint)
+            .ToList();
```
raman-m, line 43: I'm uncertain, but it may be necessary to wrap this statement with a `lock`, assuming that the class is intended to act as a singleton.

src/Ocelot.Provider.Kubernetes/Kube.cs
```diff
@@ -35,17 +33,14 @@ public virtual async Task<List<Service>> GetAsync()
             .ResourceClient(client => new EndPointClientV1(client))
             .GetAsync(_configuration.KeyOfServiceInK8s, _configuration.KubeNamespace);
```
raman-m, line 34: The integrated endpoint is capable of generating subsets of varying lengths, which suggests a review of the design may be beneficial.

src/Ocelot.Provider.Kubernetes/Kube.cs
```diff
@@ -35,17 +33,14 @@ public virtual async Task<List<Service>> GetAsync()
...
-        _services.Clear();
-        if (endpoint?.Subsets.Count != 0)
-        {
-            _services.AddRange(BuildServices(_configuration, endpoint));
-        }
-        else
+        if ((endpoint?.Subsets?.Count ?? 0) == 0)
         {
             _logger.LogWarning(() => $"K8s Namespace:{_configuration.KubeNamespace}, Service:{_configuration.KeyOfServiceInK8s}; Unable to use: it is invalid. Address must contain host only e.g. localhost and port must be greater than 0!");
+            return new(0);
```
raman-m, line 39:
```diff
! This is a generic failed response processing.
```
Therefore, we should revert to the old collection if the response fails, i.e., `Count==0`.
However, "caching" the previous state necessitates synchronization logic.

raman-m (changes requested):
It appears that the significant concurrency issue in the `Kube` class mentioned in the https://github.com/ThreeMammals/Ocelot/issues/2110#issuecomment-2189064205 has been resolved, yet a genuine fixing solution for the `RoundRobin` class is still required.

Please note that our [development process](https://ocelot.readthedocs.io/en/latest/building/releaseprocess.html) necessitates both unit and acceptance testing. It is advisable to replicate the issue with an acceptance test before implementing the bug-fixing solution.
We need to replicate the bug using an acceptance test. I can draft a preliminary version of the test, and you can then refine it to create the final version.

src/Ocelot.Provider.Kubernetes/Kube.cs
antikorol: As part of bug fixing, I wouldn't add new logic like caching, as it wasn't here before. At the very least, the cache should not be just storing the previous state. For example, if I downscale the services to 0 pods, I expect that while the pods are terminating (especially when there is some logic on terminate), nobody can call this pod using cached host and port of pod.

But the idea of using an expiring cache would positively impact the `PollKube` logic because if, for some reason, we get 0 services, we will encounter many errors during the polling interval.

I guess it should be a new feature, during the planning of which various pros and cons will be considered.

src/Ocelot.Provider.Kubernetes/Kube.cs
antikorol: I don't think this is necessary because the builder simply transforms the EndpointsV1 model into an Ocelot Service.
I've changed to `.ToList();` but in general ToList and AddRange both try to cast to a ICollection and create an array of the required length

antikorol:
> It appears that the significant concurrency issue in the `Kube` class mentioned in the [#2110 (comment)](https://github.com/ThreeMammals/Ocelot/issues/2110#issuecomment-2189064205) has been resolved, yet a genuine fixing solution for the `RoundRobin` class is still required.
> 
> Please note that our [development process](https://ocelot.readthedocs.io/en/latest/building/releaseprocess.html) necessitates both unit and acceptance testing. It is advisable to replicate the issue with an acceptance test before implementing the bug-fixing solution. We need to replicate the bug using an acceptance test. I can draft a preliminary version of the test, and you can then refine it to create the final version.



> It appears that the significant concurrency issue in the `Kube` class mentioned in the [#2110 (comment)](https://github.com/ThreeMammals/Ocelot/issues/2110#issuecomment-2189064205) has been resolved, yet a genuine fixing solution for the `RoundRobin` class is still required.
> 
> Please note that our [development process](https://ocelot.readthedocs.io/en/latest/building/releaseprocess.html) necessitates both unit and acceptance testing. It is advisable to replicate the issue with an acceptance test before implementing the bug-fixing solution. We need to replicate the bug using an acceptance test. I can draft a preliminary version of the test, and you can then refine it to create the final version.

I'm not quite sure which fix you are referring to. In this case, there will no longer be a collection of the form [service, null] because the issue with modifying the array from different threads has been fixed. Or do you mean adding a check that if the service or the HostAndPort property is null, then log the error and return a 404 status code for that request?

src/Ocelot.Provider.Kubernetes/Kube.cs
raman-m: You are correct; it appears to be a new feature.

> For example, if I downscale the services to 0 pods, I expect that while the pods are terminating (especially when there is some logic on terminate), nobody can call this pod using cached host and port of pod.

Although I am not acquainted with scenarios involving pod termination, your point is significant in resolving this debate.
If the Kubernetes client can return zero service instances while a pod is terminating, then receiving an empty collection should not pose a problem.

raman-m:
> Or do you mean adding a check that if the service or the HostAndPort property is null, then log the error and return a 404 status code for that request?

Yes, I do. Could you include this check within the `RoundRobin` class plz?
Thank you for adding the unit test. I'm assessing the efficiency of writing an acceptance test, which is expected to be quite complicated. **Have you tested the bug fix in your environment?**

src/Ocelot.Provider.Kubernetes/Kube.cs
antikorol: Yes, I was wrong.  I've checked the API and everything should be okay  if there are no pods.

antikorol:
> > Or do you mean adding a check that if the service or the HostAndPort property is null, then log the error and return a 404 status code for that request?
> 
> Yes, I do. Could you include this check within the `RoundRobin` class plz? Thank you for adding [the unit test](https://github.com/ThreeMammals/Ocelot/pull/2111/commits/0e9a616e60950f0c8107361c2956bcabdf76930f). I'm assessing the efficiency of writing an acceptance test, which is expected to be quite complicated. **Have you tested the bug fix in your environment?**

Added new error to handle this case when service or HostAndPort is null.
The fix has been tested in dev/preprod environments, and the fix was checked with load testing at around 30 rps.

raman-m:
Thank you for conducting the tests!
Is the official hotfix release package, version 23.3.4, needed to update all environments and conduct the final bugfix test?
Can you update environments with manually built DLLs?

Additionally, I'm going to review the code once more...
Having unit/integration tests is likely sufficient to deem the code acceptable.

src/Ocelot/LoadBalancer/LoadBalancers/RoundRobin.cs
```diff
@@ -30,6 +30,17 @@ public async Task<Response<ServiceHostAndPort>> Lease(HttpContext httpContext)
                     }
 
                     var next = services[_last++];
+
+                    if (next == null)
+                    {
+                        return new ErrorResponse<ServiceHostAndPort>(new ServiceIsInvalidError($"The selected service was null in {nameof(RoundRobin)} during {nameof(Lease)} operation."));
```
raman-m, line 36: Very well... But adding of a new error-class is redundant!
Pay your attention that especially for this null-case we have already existed the [ServicesAreNullError](https://github.com/search?q=repo%3AThreeMammals%2FOcelot%20ServicesAreNullError&type=code) class.
Example of usage is here in the `LeastConnection` class ->
https://github.com/ThreeMammals/Ocelot/blob/8c0180a6461984ebd11ceca5544ca9d0e1564975/src/Ocelot/LoadBalancer/LoadBalancers/LeastConnection.cs#L27
so I suggest the following change:
```suggestion
                        return new ErrorResponse<ServiceHostAndPort>(new ServicesAreNullError($"The selected service was null in {nameof(RoundRobin)} during {nameof(Lease)} operation."));
```
Also it will be informative to specify total count and current failed index as an additional information.

src/Ocelot/LoadBalancer/LoadBalancers/RoundRobin.cs
```diff
@@ -30,6 +30,17 @@ public async Task<Response<ServiceHostAndPort>> Lease(HttpContext httpContext)
...
+                    }
+
+                    if (next.HostAndPort == null)
+                    {
+                        return new ErrorResponse<ServiceHostAndPort>(new ServiceIsInvalidError($"The {nameof(next.HostAndPort)} was null in the selected service in {nameof(RoundRobin)} during {nameof(Lease)} operation."));
```
raman-m, line 41: As previously discussed, the `ServicesAreNullError` class can also be repurposed for this scenario.

test/Ocelot.UnitTests/Responder/ErrorsToHttpStatusCodeMapperTests.cs
raman-m, line 1: The new changes become unnecessary if the `ServiceIsInvalidError` class is removed from the code base.

raman-m (changes requested):
The primary concern is the redundancy of the `ServiceIsInvalidError` class.

src/Ocelot.Provider.Kubernetes/Kube.cs
raman-m: What do you suggest here? I can't get you...

src/Ocelot/LoadBalancer/LoadBalancers/ServiceIsInvalidError.cs
```diff
@@ -0,0 +1,12 @@
+using Ocelot.Errors;
+
+namespace Ocelot.LoadBalancer.LoadBalancers
+{
+    public class ServiceIsInvalidError : Error
```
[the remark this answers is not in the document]
antikorol: Yes, I have seen that class and where it is used. It is used for errors when instead of a list of services, null is received (Even the name of the class is used in the plural form).
In this case, there will be a list, but one of the elements will be null, and the other will be valid.

But if you still think it's better to use it, then okay, I'll delete it.

antikorol:
> Thank you for conducting the tests! Is the official hotfix release package, version 23.3.4, needed to update your all environments and conduct the final bugfix test? Can you update environments with manually built DLLs?

At your discretion, I can release it with my fix for now and remove it in the future.

raman-m:
Great, Roman!
I'd like to merge it into the develop branch, after which you can create a temporary build from your forked repository and deploy it to the Prod environment, okay?

Could you provide more details about your Production environment?
Please include information on the hosting server (Docker, IIS, self-hosted), the number of Ocelot instances, CPU & memory resources per Ocelot instance, requests per second, and the technical stack (specific Ocelot features utilized), etc.

Is the Ocelot gateway deployed on a public or private network?
Also, could you specify the industry that the Prod product is part of?

src/Ocelot/LoadBalancer/LoadBalancers/ServiceIsInvalidError.cs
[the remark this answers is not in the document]
raman-m: The error message can be interpreted as "Some services are null." 😊

src/Ocelot.Provider.Kubernetes/Kube.cs
raman-m: Agreed! We'll keep it as it is. No problem at all.

src/Ocelot/LoadBalancer/LoadBalancers/ServiceIsInvalidError.cs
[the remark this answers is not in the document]
antikorol: Deleted custom error and now using ServicesAreNullError

antikorol:
> Is the Ocelot gateway deployed on a public or private network?

Unfortunately, the game is still in closed beta testing, and the loads on prod are much times lower than what we simulate during tests. Ocelot is deployed as a regular deployment in k8s.
But If I find any issues, I will let you know :)

src/Ocelot/LoadBalancer/LoadBalancers/ServiceIsInvalidError.cs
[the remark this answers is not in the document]
raman-m: Fixed in https://github.com/ThreeMammals/Ocelot/pull/2111/commits/bbed7d1aa195f425939cc5f3a69e245f6bdd565e ✔️

src/Ocelot/LoadBalancer/LoadBalancers/RoundRobin.cs
raman-m: Fixed in https://github.com/ThreeMammals/Ocelot/pull/2111/commits/bbed7d1aa195f425939cc5f3a69e245f6bdd565e

test/Ocelot.UnitTests/Responder/ErrorsToHttpStatusCodeMapperTests.cs
raman-m: Done in https://github.com/ThreeMammals/Ocelot/pull/2111/commits/bbed7d1aa195f425939cc5f3a69e245f6bdd565e

raman-m:
Hello Roman,
I've come back to your PR to complete the remaining work.
I have some news and will provide you with feedback either today or tomorrow...

raman-m:
@RaynaldM @ggnaegi Hello team!

Ray, I'm aware you're likely enjoying your time attending various Olympic events in Paris 🥇, which is an amazing lifestyle and form of relaxation, but I would appreciate your forthcoming review.

Gui, I hope for your understanding and not too harsh a judgment, given that it's the `Kube` provider, not `Consul`. Recognizing this as a significant [v23.3 Hotfixes](https://github.com/ThreeMammals/Ocelot/milestone/8) milestone we aim to achieve by September, your review will be vital for the release.

### P.S.
I will make only one more commit, which will include styling of tests and the division of tests into two theories, ensuring there will be no further changes to the core logic.

ggnaegi:
Ok, I will review this.

raman-m:
@antikorol Hello Roman,
In [build 4286](https://app.circleci.com/pipelines/github/ThreeMammals/Ocelot/2347/workflows/1b96a8a1-d1b1-414c-9490-c7c350951089/jobs/4286), as you can observe, the testing results for the [ShouldReturnServicesFromK8s_HighlyLoadOnUnStableProvider](https://github.com/ThreeMammals/Ocelot/pull/2111/files#diff-857f76beee3cb7c8488c15385322a6f55f272bc7d8e529ccc9f9bef5e1c692a9R151) acceptance test are inconsistent. While it passes most of the time in Visual Studio Test Explorer, it occasionally fails due to the K8s client integration with the `Kube` provider, particularly when using the "Run Until Failure" command. This leads to an empty list of services being sent to the load balancer, resulting in a `ServicesAreEmptyError` and a subsequent 404 error, causing the acceptance test to fail as it expects all statuses to be OK.
image
The instability's root cause is uncertain; it could lie within the `ServiceHandler` during parallel execution or the `IKubeApiClient` might also intermittently fail. It appears we are facing a typical integration issue that may not have a straightforward solution, and we need to decide on our approach for releasing the hotfix, considering the following options:

1. Address the unstable scenario acceptance test by considering failed 404 responses, although this might be a makeshift solution.
2. Amend the unstable scenario acceptance test to account for 404 statuses by initiating an additional request to achieve an OK 200 status, thereby increasing service counters for test assertions. This mirrors real-world scenarios where clients implement a Retry strategy upon receiving unsuccessful responses (404 or 500). While we will correct the test and issue a hotfix, developers will continue to observe 404 errors in the logs, necessitating the implementation of retry logic on the client side.
3. Implement the Retry pattern directly within the Ocelot core to stabilize the integration endpoint, either at the load balancer or provider level. Integrating retry logic at the load balancer level is suboptimal, hence an undesirable approach. Conversely, incorporating retry logic within the `Kube` provider is a viable solution. Although Polly is not an option, a manual Retry algorithm will be developed.

Please share your thoughts!
**P.S.** I believe the first option is entirely impractical, and I am deliberating between the second and third options, convinced that the Retry pattern must be adopted.

raman-m:
I've opted for the third option and implemented the Retry pattern. The previously unstable scenario test is now performing much better. I would consider it quite stable. So, I've resolved the issue with the 404 status, as mentioned above. However, sometimes Ocelot returns a 500 status, and the underlying cause is currently unclear to me.
image
Investigating this issue necessitates further development and execution of tests. However, this error is quite rare, occurring at a rate of 1/100, 1/500, or even 1/1000, which means it demands extensive testing in production environments, as production is certainly not the same as testing.
In conclusion, I believe we can proceed with the release at this hotfix quality. 😎

src/Ocelot.Provider.Kubernetes/Kube.cs
```diff
@@ -26,28 +30,39 @@ public class Kube : IServiceDiscoveryProvider
...
+    private Task<EndpointsV1> GetEndpoint() => _kubeApi
+        .ResourceClient(client => new EndPointClientV1(client))
+        .GetAsync(_configuration.KeyOfServiceInK8s, _configuration.KubeNamespace);
+
+    private bool Ensure(EndpointsV1 endpoint)
+    {
+        if ((endpoint?.Subsets?.Count ?? 0) == 0)
         {
-            _logger.LogWarning(() => $"K8s Namespace:{_configuration.KubeNamespace}, Service:{_configuration.KeyOfServiceInK8s}; Unable to use: it is invalid. Address must contain host only e.g. localhost and port must be greater than 0!");
+            _logger.LogWarning(() => GetMessage($"Endpoint ensuring has been failed! Endpoint object is null({endpoint == null}), or its subsets collection lenth is {endpoint?.Subsets?.Count ?? 0}."));
```
raman-m, line 57: Typo in `lenth is` 🤣

src/Ocelot.Provider.Kubernetes/Kube.cs
```diff
@@ -26,28 +30,39 @@ public class Kube : IServiceDiscoveryProvider
...
-            .GetAsync(_configuration.KeyOfServiceInK8s, _configuration.KubeNamespace);
+        var endpoint = await Retry.OperationAsync(GetEndpoint, Ensure, logger: _logger);
 
-        _services.Clear();
-        if (endpoint?.Subsets.Count != 0)
+        if (Ensure(endpoint))
         {
-            _services.AddRange(BuildServices(_configuration, endpoint));
+            _logger.LogWarning(() => GetMessage($"Unable to use bad result returned by {nameof(Kube)} integration endpoint because the final result is invalid/unknown after multiple retries!"));
+            return new(0);
```
ggnaegi, line 42: You could use collection expression ```[]``` here instead of ```new(0)```

src/Ocelot.Provider.Kubernetes/Kube.cs
```diff
@@ -26,28 +30,39 @@ public class Kube : IServiceDiscoveryProvider
...
     public virtual async Task<List<Service>> GetAsync()
     {
-        var endpoint = await _kubeApi
-            .ResourceClient(client => new EndPointClientV1(client))
-            .GetAsync(_configuration.KeyOfServiceInK8s, _configuration.KubeNamespace);
+        var endpoint = await Retry.OperationAsync(GetEndpoint, Ensure, logger: _logger);
 
-        _services.Clear();
-        if (endpoint?.Subsets.Count != 0)
+        if (Ensure(endpoint))
```
ggnaegi, line 39: You could invert the if, but that's a question of taste

src/Ocelot.Provider.Kubernetes/Kube.cs
```diff
@@ -26,28 +30,39 @@ public class Kube : IServiceDiscoveryProvider
...
+
+        return BuildServices(_configuration, endpoint)
+            .ToList();
+    }
+
+    private Task<EndpointsV1> GetEndpoint() => _kubeApi
+        .ResourceClient(client => new EndPointClientV1(client))
+        .GetAsync(_configuration.KeyOfServiceInK8s, _configuration.KubeNamespace);
+
+    private bool Ensure(EndpointsV1 endpoint)
```
ggnaegi, line 53: I'm not sure about this method. Do we really need it? First, we are writing two warning messages here, the first one in ```Ensure``` the second one in ```GetAsync```.  I don't like methods returning true or false explicitely...

src/Ocelot.Provider.Kubernetes/KubeServiceCreator.cs
```diff
@@ -34,6 +34,8 @@ public virtual IEnumerable<Service> CreateInstance(KubeRegistryConfiguration con
         return new Service[] { instance };
     }
 
+    protected IOcelotLogger Logger => _logger;
```
ggnaegi, line 37: Why not use auto-property instead?

src/Ocelot/Infrastructure/DesignPatterns/Retry.cs
```diff
@@ -0,0 +1,111 @@
...
+    /// <returns>A <typeparamref name="TResult"/> value as the result of the sync operation.</returns>
+    public static TResult Operation<TResult>(
+        Func<TResult> operation,
+        Predicate<TResult> predicate = null,
+        int retryTimes = DefaultRetryTimes, int waitTime = DefaultWaitTimeMilliseconds,
+        IOcelotLogger logger = null)
+    {
+        for (int i = 0; i < retryTimes - 1; i++)
+        {
+            TResult result = default;
```
ggnaegi, line 36: The default is not needed here... By the way, sometimes, I'm wondering why we are reeinventing the wheel everywhere when we could simply use polly for the retry...

src/Ocelot/LoadBalancer/LoadBalancers/CookieStickySessions.cs
```diff
@@ -1,85 +1,81 @@
...
+
+public class CookieStickySessions : ILoadBalancer
 {
-    public class CookieStickySessions : ILoadBalancer
+    private readonly int _keyExpiryInMs;
+    private readonly string _cookieName;
+    private readonly ILoadBalancer _loadBalancer;
+    private readonly IBus<StickySession> _bus;
+
+    private static readonly object _lock = new();
```
ggnaegi, line 16: Should be written ```Lock```

src/Ocelot/LoadBalancer/LoadBalancers/CookieStickySessions.cs
```diff
@@ -1,85 +1,81 @@
...
+    private static readonly ConcurrentDictionary<string, StickySession> _stored = new(); // TODO Inject instead of static sharing
```
ggnaegi, line 17: Should be ```Stored```

src/Ocelot/LoadBalancer/LoadBalancers/CookieStickySessions.cs
```diff
@@ -1,85 +1,81 @@
...
-        private readonly object _lock = new();
+        _bus = bus;
+        _cookieName = cookieName;
+        _keyExpiryInMs = keyExpiryInMs;
+        _loadBalancer = loadBalancer;
+        _bus.Subscribe(CheckExpiry);
+    }
 
-        public CookieStickySessions(ILoadBalancer loadBalancer, string key, int keyExpiryInMs, IBus<StickySession> bus)
+    private void CheckExpiry(StickySession sticky)
```
ggnaegi, line 28: Instead of mixing ```Lock``` and ```ConcurrentDictionary``` why not use the lock only, with a plain and simple dictionary eg.
```c#
private void CheckExpiry(StickySession sticky)
    {
        // TODO Get test coverage for this
        lock (Lock)
        {
            if (!Stored.TryGetValue(sticky.Key, out var session))
            {
                return;
            }

            if (session.Expiry >= DateTime.UtcNow)
            {
                return;
            }

            Stored.Remove(session.Key);
            _loadBalancer.Release(session.HostAndPort);
        }
    }
```

src/Ocelot/LoadBalancer/LoadBalancers/CookieStickySessions.cs
```diff
@@ -1,85 +1,81 @@
...
+            var ss = new StickySession(next.Data, DateTime.UtcNow.AddMilliseconds(_keyExpiryInMs), key);
+            Update(key, ss);
+            return Task.FromResult<Response<ServiceHostAndPort>>(new OkResponse<ServiceHostAndPort>(next.Data));
         }
+    }
 
-        public void Release(ServiceHostAndPort hostAndPort)
-        {
-        }
+    protected void Update(string key, StickySession value)
```
ggnaegi, line 72: Use ```Lock``` and plain and simple dictionary instead, eg:
```c#
protected void Update(string key, StickySession value)
    {
        lock (Lock)
        {
            Stored[key] = value;
            _bus.Publish(value, _keyExpiryInMs);
        }
    }
```

src/Ocelot/LoadBalancer/LoadBalancers/Lease.cs
```diff
@@ -1,16 +1,63 @@
...
+        Connections = connections;
     }
-}
+
+    public ServiceHostAndPort HostAndPort { get; }
+    public int Connections { get; set; }
+
+    public static Lease Null => new();
+
+    public override readonly string ToString() => string.Format("({0}+{1})", HostAndPort, Connections);
```
ggnaegi, line 36: string interpolation?

src/Ocelot/LoadBalancer/LoadBalancers/LeastConnection.cs
```diff
@@ -2,140 +2,85 @@
...
-
-                _leases.Add(leaseWithLeastConnections);
-
-                return new OkResponse<ServiceHostAndPort>(new ServiceHostAndPort(leaseWithLeastConnections.HostAndPort.DownstreamHost, leaseWithLeastConnections.HostAndPort.DownstreamPort));
-            }
+            return new ErrorResponse<ServiceHostAndPort>(new ServicesAreNullError($"Services were null in {nameof(LeastConnection)} for '{_serviceName}' during {nameof(Lease)} operation!"));
         }
-
-        public void Release(ServiceHostAndPort hostAndPort)
+        else if (services.Count == 0)
```
ggnaegi, line 29: ```else``` is not needed here

src/Ocelot/LoadBalancer/LoadBalancers/RoundRobin.cs
```diff
@@ -2,43 +2,108 @@
...
+
+    private static readonly Dictionary<string, int> LastIndices = new();
+    protected static readonly object SyncRoot = new();
 
-        private int _last;
+    public event EventHandler<LeaseEventArgs> Leased;
+    protected virtual void OnLeased(LeaseEventArgs e) => Leased?.Invoke(this, e);
 
-        public RoundRobin(Func<Task<List<Service>>> services)
+    public virtual async Task<Response<ServiceHostAndPort>> Lease(HttpContext httpContext)
```
ggnaegi, line 26: This method is too long, I would split it

src/Ocelot/Values/ServiceHostAndPort.cs
```diff
@@ -1,20 +1,54 @@
...
+    ///   <item><seealso href="https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-object-equals">System.Object.Equals method</seealso></item>
+    ///   <item><seealso href="https://learn.microsoft.com/en-us/dotnet/api/system.iequatable-1.equals?view=net-8.0">IEquatable&lt;T&gt;.Equals(T) Method</seealso></item>
+    ///   </list>
+    /// </remarks>
+    /// <param name="l">Left operand.</param>
+    /// <param name="r">Right operand.</param>
+    /// <returns><see langword="true"/> if both operands are equal; otherwise, <see langword="false"/>.</returns>
+    public static bool operator ==(ServiceHostAndPort l, ServiceHostAndPort r)
+        => (((object)l) == null || ((object)r) == null)
+            ? object.Equals(l, r)
```
ggnaegi, line 47: ```object``` is redundant

src/Ocelot/Values/ServiceHostAndPort.cs
```diff
@@ -1,20 +1,54 @@
...
+            : l.DownstreamHost == r.DownstreamHost && l.DownstreamPort == r.DownstreamPort && l.Scheme == r.Scheme;
+
+    public static bool operator !=(ServiceHostAndPort l, ServiceHostAndPort r)
+        => (((object)l) == null || ((object)r) == null)
+            ? !object.Equals(l, r)
```
ggnaegi, line 52: ```object``` is redundant

src/Ocelot/Values/ServiceHostAndPort.cs
```diff
@@ -1,20 +1,54 @@
...
+    public string Scheme { get; }
+
+    public override string ToString()
+        => $"{Scheme}:{DownstreamHost}:{DownstreamPort}";
+    public override int GetHashCode()
+        => Tuple.Create(Scheme, DownstreamHost, DownstreamPort).GetHashCode();
+
+    public bool Equals(ServiceHostAndPort other) => this == other;
+    public override bool Equals(object obj)
+        => obj != null && obj is ServiceHostAndPort o && this == o;
```
ggnaegi, line 32:
```c#
public override bool Equals(object obj)
        => obj is ServiceHostAndPort o && this == o;
```
merge into pattern?

ggnaegi (changes requested):

src/Ocelot.Provider.Kubernetes/Kube.cs
raman-m: Due to the lack of support for [Collection expressions](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-12#collection-expressions) in .NET 6/7, we will encounter warnings in our build logs. I've experienced this issue before with testing projects, where a large number of such warnings caused my confusion. [Collection expressions](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-12#collection-expressions) were introduced in [C# 12](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-12) for .NET 8. Opening the solution in Visual Studio 2018 or earlier versions will likely result in a build failure. Only Visual Studio 2022 version 17.8 or later can recognize **Collection Expressions** and compile them without warnings or errors. However, even with VS v17.8+, targeting `net7.0` and `net6.0` frameworks still generates warnings in the logs. It appears we must wait for the [end-of-support](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core) of .NET 6/7 to fully utilize this feature of C# 12. My apologies! We will only utilize .NET 6 features until November 2024. 😉 Following the release of .NET 9, we will begin using the new feature as we migrate our solution to .NET 8/9. Is that okay?

src/Ocelot.Provider.Kubernetes/Kube.cs
raman-m: Makes sense, but you mean readability issue?... I pass `Ensure` delegate to `Retry` method. so I can't negate delegate ))
But I can rename the method to `NotSure` or `Check` or `IsInvalid`.
Or I can negate `bool` constants inside the method, but it won't fix readability issue. So, I'd prefer to rename to `Check` without negation. Please note that according to `OperationAsync` definition "Retry an operation when **predicate is true**". So, I don't want to re-define `OperationAsync`.

src/Ocelot/Infrastructure/DesignPatterns/Retry.cs
raman-m: We cannot adopt Polly for several reasons:
- Polly has [PredicateBuilder](https://www.pollydocs.org/api/Polly.PredicateBuilder-1.html) feature but its another reinvention of the wheel for Polly fans who can't provide lite solutions. Also docs says in remarks: "Please be aware of the performance penalty if you register too many predicates with this builder. In such case, it's better to create your own predicate manually as a delegate." So I did manually! I believe my implementation has better performance.
- It is a large and complex library with numerous settings, which is unnecessary for simple tasks.
- Using Polly alongside other libraries can be risky for developers. We need to adhere to a single integration package, namely `KubeClient`, and referencing `Polly` is huge design mistake in my opinion.
Debugging the attached Polly pipeline will be a developer's nightmare for us! New `Retry` methods will be debugged easily!

In contrast, this `Retry` offers a lightweight implementation, giving us more control over the code compared to Polly, which has a substantial amount of partially "closed" code.
Sorry, I can't add `Polly` reference to [Ocelot.Provider.Kubernetes.csproj](https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot.Provider.Kubernetes/Ocelot.Provider.Kubernetes.csproj#L31-L36) package! And I wondered that you are asking me to do that 🤯

raman-m: >The default is not needed here... 

OK It is minor issue, we could still assign. But seems compiler does the same even without [default literal](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/default#default-literal)
Fair enough!

src/Ocelot/LoadBalancer/LoadBalancers/LeastConnection.cs
raman-m: Again, this is Tom's logic. )) But it's OK. Let's improve this.

src/Ocelot/Values/ServiceHostAndPort.cs
raman-m: `object.Equals(l, r)` &rarr; `Equals(l, r)` ? 🆗

src/Ocelot/Values/ServiceHostAndPort.cs
raman-m: **P.S.** Probably you mean: `object.Equals(l, r)` &rarr; `Equals(l, r)` ? Then 🆗I will fix this.

src/Ocelot.Provider.Kubernetes/Kube.cs
raman-m: > I don't like methods returning true or false explicitly...

Fair enough! I don't like seeing old Tom's code with the same code bool-pattern 🤣 
Fixed via warning logging removal.

raman-m (approved):
@ggnaegi I've addressed all issues from your [code review](https://github.com/ThreeMammals/Ocelot/pull/2111#pullrequestreview-2220577739). Awaiting your approval or next suggestions...
@RaynaldM Welcome!

src/Ocelot/Values/ServiceHostAndPort.cs
raman-m: > merge into pattern?

Swiss humor? Yeah! 🥲 I know you don't like my `Retry` pattern because you like Polly. 😏 
Let's ask another big fan of Polly... not @RaynaldM 🤣 ... and this is Mr. Seth @ks1990cn Hello Tanmay! 😉

src/Ocelot/Values/ServiceHostAndPort.cs
ggnaegi: @raman-m yes 😊

raman-m: Already fixed in https://github.com/ThreeMammals/Ocelot/pull/2111/commits/c4c7ae0ff2a891d64b4f81b6a6d051cc15e71211#diff-5baf30f066ec9efc9275e78bd80143b4c967bda577cbff011f1580598f8f80e5

src/Ocelot/Values/ServiceHostAndPort.cs
raman-m: Fixed ✔️  See https://github.com/ThreeMammals/Ocelot/pull/2111/commits/c4c7ae0ff2a891d64b4f81b6a6d051cc15e71211#diff-5baf30f066ec9efc9275e78bd80143b4c967bda577cbff011f1580598f8f80e5

src/Ocelot/LoadBalancer/LoadBalancers/RoundRobin.cs
raman-m: Fixed, the method's code was split into 3 sub-methods
See https://github.com/ThreeMammals/Ocelot/pull/2111/commits/c4c7ae0ff2a891d64b4f81b6a6d051cc15e71211#diff-cc104d57540132d29a89de66a0c0db287f8a85c2b7d5c20ed4d7958652db5847

src/Ocelot/LoadBalancer/LoadBalancers/LeastConnection.cs
raman-m: Fixed in favor of single if-statement. See https://github.com/ThreeMammals/Ocelot/pull/2111/commits/c4c7ae0ff2a891d64b4f81b6a6d051cc15e71211#diff-694338581b66de3c2a9c18a02e9431a9b13e1397f2532fa282cbc2751cc6fa08

src/Ocelot/LoadBalancer/LoadBalancers/Lease.cs
raman-m: Fixed, see https://github.com/ThreeMammals/Ocelot/pull/2111/commits/c4c7ae0ff2a891d64b4f81b6a6d051cc15e71211#diff-757cd170e7bbb0d07accf7058c4336d43cf007040ae3bbf00d7f1f4ecf5832cd

src/Ocelot/LoadBalancer/LoadBalancers/CookieStickySessions.cs
raman-m: I've followed all your suggestions for `CookieStickySessions` and refactored.
See combined changes in https://github.com/ThreeMammals/Ocelot/pull/2111/commits/c4c7ae0ff2a891d64b4f81b6a6d051cc15e71211#diff-75ae5165d973013bcfd7ad2d5776a4868cac5f7d22931687de8fc63c80e7e36b

src/Ocelot/Infrastructure/DesignPatterns/Retry.cs
raman-m: > It's just that we have several code blocks for retry and for timeouts. Probably using polly isn't the right strategy, but you got what I meant...

I understand your points; however, [my counterarguments](https://github.com/ThreeMammals/Ocelot/pull/2111#discussion_r1705420376) to Polly can be found above in the discussion.

Tiny issue with `default` is fixed, see https://github.com/ThreeMammals/Ocelot/pull/2111/commits/c4c7ae0ff2a891d64b4f81b6a6d051cc15e71211#diff-01ceec0d1ccf6008635b11e7d2c5a6d2ef2654eaba66371ffc488d7f1b0c7563  

I hope this thread can be closed.

src/Ocelot.Provider.Kubernetes/KubeServiceCreator.cs
raman-m: Fixed, see https://github.com/ThreeMammals/Ocelot/pull/2111/commits/c4c7ae0ff2a891d64b4f81b6a6d051cc15e71211#diff-aa904908dbabbf75561309a7e6be8cca8c423ac49d0c1d7a77bd8dd609316b17

src/Ocelot.Provider.Kubernetes/Kube.cs
raman-m: Your suggestions were followed, see changes https://github.com/ThreeMammals/Ocelot/pull/2111/commits/c4c7ae0ff2a891d64b4f81b6a6d051cc15e71211#diff-9b178d28d53bc91aa0d4b24da98f78648855928a824eabde6eb3a87acb115763

raman-m:
@ggnaegi What else?
Did I miss something from your code review?

raman-m (approved):

ggnaegi (approved):
Ok, fine for me, thanks!

raman-m:
### TODO
- [ ] Evaluate the initiation of a new project as an Ocelot package for [kubernetes-client/csharp](https://github.com/kubernetes-client/csharp), replacing the currently used [tintoy/dotnet-kube-client](https://github.com/tintoy/dotnet-kube-client) library due to its slow development pace. This would involve transitioning to the industry-standard library for Kubernetes service discovery.
- [ ] Undertake refactoring of load balancing following the identification of design bottlenecks in the [Custom Load Balancers](https://ocelot.readthedocs.io/en/latest/features/loadbalancer.html#custom-load-balancers) feature.
  References: release [15.0.3](https://github.com/ThreeMammals/Ocelot/releases/tag/15.0.3), feature commit https://github.com/ThreeMammals/Ocelot/commit/ccb5b841032ffc8033b87ea3f84f39f9f17edd88, PR #1155, Feature #961.

src/Ocelot.Provider.Kubernetes/Ocelot.Provider.Kubernetes.csproj
```diff
@@ -29,8 +29,8 @@
     <Compile Remove="KubeApiClientFactory.cs" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="KubeClient" Version="2.4.10" />
-    <PackageReference Include="KubeClient.Extensions.DependencyInjection" Version="2.4.10" />
+    <PackageReference Include="KubeClient" Version="2.5.8" />
+    <PackageReference Include="KubeClient.Extensions.DependencyInjection" Version="2.5.8" />
```
raman-m, line 33: ### TODO
The package has been updated to version [2.5.10](https://www.nuget.org/packages/KubeClient/#versions-body-tab), released on August 11, 2024. It is necessary to update to this version prior to the next release.

antikorol:
> **Please share your thoughts**❕
> **P.S.** I believe the 1st option is entirely impractical, and I am deliberating between the 2nd and 3rd options, convinced that the Retry pattern must be adopted.

Hello @raman-m Raman. 
Sorry. I was on vacation and didn't notice, among the many notifications, that there was a question directed at me. However, I see the solution has already been implemented :)

raman-m:
@antikorol Right! The hotfix has been implemented! However I've [asked](https://github.com/ThreeMammals/Ocelot/issues/2110#issuecomment-2273670573) you to start testing.
Will you wait for official package release?

antikorol:
> @antikorol Right! The hotfix has been implemented! However I've [asked](https://github.com/ThreeMammals/Ocelot/issues/2110#issuecomment-2273670573) you to start testing. Will you wait for official [23.3.4](https://github.com/ThreeMammals/Ocelot/releases/tag/23.3.4) package release?

@raman-m 
According to company policy, we are not allowed to publish packages to Artifactory that are not related to our project, and we can only use specific sources for NuGet packages. As soon as you release a new version, I will update and thoroughly test the fixes.

**However**, I copied your changes related to fix into my custom Kubernetes provider, which we are currently using as a fix, and so far everything is working well.

Thank you

src/Ocelot.Provider.Kubernetes/Ocelot.Provider.Kubernetes.csproj
raman-m: The packages was updated in release [23.3.4](https://github.com/ThreeMammals/Ocelot/releases/tag/23.3.4) on Oct 3, 2024

raman-m:
@antikorol FYI &rarr; [Ocelot 23.3.4](https://www.nuget.org/packages/Ocelot/23.3.4) Nuget package

PR #2120, opened 2024-07-12
#2084 Apply default config file paths in `GetMergedOcelotJson` when providing the `folder` argument of `AddOcelot`

### LINKED ISSUE #2084: After upgrade to Ocelot 23.2, `ocelot.global.json` file no longer gets merged correctly
## Expected Behavior

Data inside `ocelot.global.json` file should be taken into account when creating the `IInternalConfiguration`

## Actual Behavior

Defaults are being set, the data from the `ocelot.global.json` is completely ignored (we have a `RequestIdKey` set and a custom `ServiceDiscoveryProvider`)

  1. Create multiple json files inside a folder, each containing different routes
  2. Create `ocelot.global.json` file inside the same folder
  3. Use the `Configuration.AddOcelot(string folder, IWebHostEnvironment env, MergeOcelotJson.ToMemory)` overload
  4. At runtime, use the `IInternalConfigurationRepository` to get the Ocelot configuration. 
  5. The data found in the `InternalConfiguration` object regarding global configuration contains default values, instead of the ones from the `ocelot.global.json`

## Specifications

  - Version: 23.2.2 on .NET 8
  - Platform: Windows
 

After the upgrade to 23.2.2, the `ocelot.global.json` no longer gets interpreted when doing the merge, regardless if it's done InMemory or the old way, in the `ocelot.json` file. Since the deployment will be done in a Kubernetes cluster, we really want to use the InMemory option, but currently, it looks like there's an issue that was introduced with 23.2.2. 

Just to make it clear, I've tested with 23.1.0 and everything is working as before, without any other changes to the code outside of removing the MergeOcelotJson from the Program.cs file to make it compile

I'm more than happy to give more details in case it's needed

ben-bartholomew:
## Fixes #2084 
- #2084 

## Proposed Changes
  - Combining user provided folder path with the default names for global, primary, or environment config files, if those filenames are not provided using folder configuration with `AddOcelot()`
    - This is only done when the default names are used (i.e., when the user does not specify a filename)
  - Added unit test for this issue


raman-m:
Hello Ben!
Thank you for opening the PR!
I'll review today and possibly merge tomorrow.

src/Ocelot/DependencyInjection/ConfigurationBuilderExtensions.cs
```diff
@@ -97,16 +97,16 @@ public static IConfigurationBuilder AddOcelot(this IConfigurationBuilder builder
...
+                builder.AddJsonStream(new MemoryStream(Encoding.UTF8.GetBytes(json))) :
                 AddOcelotJsonFile(builder, json, primaryConfigFile, optional, reloadOnChange);
         }
 
         private static string GetMergedOcelotJson(string folder, IWebHostEnvironment env,
             FileConfiguration fileConfiguration = null, string primaryFile = null, string globalFile = null, string environmentFile = null)
         {
             var envName = string.IsNullOrEmpty(env?.EnvironmentName) ? "Development" : env.EnvironmentName;
-            environmentFile ??= string.Format(EnvironmentConfigFile, envName);
+            environmentFile ??= Path.Join(folder, string.Format(EnvironmentConfigFile, envName));
```
raman-m, line 109: I’m really wondering… Typically, `FileInfo` objects represent individual files and contain properties like the file name, size, and creation date. However, **don't they inherently include the folder information?**
We enumerate `FileInfo` objects on line 112.

test/Ocelot.UnitTests/DependencyInjection/ConfigurationBuilderExtensionsTests.cs
```diff
@@ -280,7 +302,7 @@ private void ThenTheConfigsAreMergedAndAddedInApplicationConfiguration(bool useC
             fc.Routes.ShouldContain(x => x.UpstreamHost == (useCombinedConfig ? _combinedFileConfiguration.Routes[1].UpstreamHost : _routeB.Routes[0].UpstreamHost));
             fc.Routes.ShouldContain(x => x.UpstreamHost == (useCombinedConfig ? _combinedFileConfiguration.Routes[2].UpstreamHost : _routeB.Routes[1].UpstreamHost));
 
-            fc.Aggregates.Count.ShouldBe(useCombinedConfig ? _combinedFileConfiguration.Aggregates.Count :_aggregate.Aggregates.Count);
+            fc.Aggregates.Count.ShouldBe(useCombinedConfig ? _combinedFileConfiguration.Aggregates.Count : _aggregate.Aggregates.Count);
```
raman-m, line 305: Fake change

src/Ocelot/DependencyInjection/ConfigurationBuilderExtensions.cs
```diff
@@ -97,16 +97,16 @@ public static IConfigurationBuilder AddOcelot(this IConfigurationBuilder builder
         private static IConfigurationBuilder ApplyMergeOcelotJsonOption(IConfigurationBuilder builder, MergeOcelotJson mergeTo, string json,
             string primaryConfigFile, bool? optional, bool? reloadOnChange)
         {
-            return mergeTo == MergeOcelotJson.ToMemory ? 
-                builder.AddJsonStream(new MemoryStream(Encoding.UTF8.GetBytes(json))) : 
+            return mergeTo == MergeOcelotJson.ToMemory ?
+                builder.AddJsonStream(new MemoryStream(Encoding.UTF8.GetBytes(json))) :
```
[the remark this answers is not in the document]
ben-bartholomew: You're right, I have format-on-save enabled, so this was just Visual Studio fixing whitespace issues. I'll revert this chunk.

test/Ocelot.UnitTests/DependencyInjection/ConfigurationBuilderExtensionsTests.cs
ben-bartholomew: You're right, I have format-on-save enabled, so this was just Visual Studio fixing whitespace issues. I'll revert this chunk.

src/Ocelot/DependencyInjection/ConfigurationBuilderExtensions.cs
ben-bartholomew: Yes, the `FileInfo` object contains the full path for the file. That's actually the problem.

The pattern used to check if one of the files matches one of these special names is:
```cs
if (file.Name.Equals(globalFileInfo.Name, StringComparison.OrdinalIgnoreCase) &&
    file.FullName.Equals(globalFileInfo.FullName, StringComparison.OrdinalIgnoreCase))
```

But without these changes `FullName` will never match when `globalFile` is using the default filename, because it lacks the user provided folder.

So, `file.FullName` will be something like "C:/PathToOcelotConfig/**UserValueForFolder**/ocelot.global.json" while `globalFileInfo.FullName` will be something like "C:/PathToOcelotConfig/ocelot.global.json".

That's why this is a simple fix, just prefixing the `folder` path to the default name.

This was all discussed in the original issue. You can also remove my changes and run the included test case.

src/Ocelot/DependencyInjection/ConfigurationBuilderExtensions.cs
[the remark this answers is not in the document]
ben-bartholomew: Whitespace removed.

test/Ocelot.UnitTests/DependencyInjection/ConfigurationBuilderExtensionsTests.cs
ben-bartholomew: Whitespace removed.

src/Ocelot/DependencyInjection/ConfigurationBuilderExtensions.cs
raman-m: Thank you for explanation!
The problem is string concatenation, not working with `FileInfo`.
It was better to use `FileInfo` argument for the method, but it is a bit late, and we have to have a deal with `string`.

ben-bartholomew: I'm not sure I would go that far, but either way making it explicit about what path is expected for arguments like `globalConfigFile` would do.

Ok to resolve?

ggnaegi (approved):

raman-m:
@ggnaegi Thanks for approving this, but there is no acceptance test(s) for the linked bug issue.
I'm working on acceptance test now...

src/Ocelot/DependencyInjection/ConfigurationBuilderExtensions.cs
```diff
@@ -91,6 +91,7 @@ public static IConfigurationBuilder AddOcelot(this IConfigurationBuilder builder
             string primaryConfigFile = null, string globalConfigFile = null, string environmentConfigFile = null, bool? optional = null, bool? reloadOnChange = null) // optional injections
         {
             var json = GetMergedOcelotJson(folder, env, null, primaryConfigFile, globalConfigFile, environmentConfigFile);
+            primaryConfigFile ??= Path.Join(folder, PrimaryConfigFile); // if not specified, merge & write back to the same folder
```
raman-m, line 94: Hi Ben,
We overlooked this change, and acceptance testing has revealed the issue. The folder information should be retained, and the file must be written back to the same folder.

raman-m (approved):
### Ready for delivery❗

raman-m:
Thank you, @ben-bartholomew, for your contribution!

PR #1389, opened 2020-12-07
Add support for other HTTP verbs for aggregation aka `UpstreamHttpMethod` enhancements

crb02005:
This changes the FileAggregateRoute to allow other verbs. 
It is for consumers not following traditional REST. Several APIs do not follow the REST guidelines and may use POST for getting data. In cases like these being able to specify a verb is helpful.

PR #2146, opened 2024-09-03
Downgrade the Warning to Information on missing `Content-Length` header in `MultiplexingMiddleware`

PaulARoy:
Aggregations log a lot of warnings due to missing content-headers (for aggregations without body).

Most of the time it's a completely normal behavior and it doesn't justify a warning message, we simply don't have a body. It might be interesting to keep an information log though.

Not sure tests are relevant for this minor fix as it's just logging.


raman-m:
Hello Paul,
It is related to Staging and Production environments only. For Development we don't care about logging levels because we specify usually `Debug` (level 1) or even `Trace` (level 0).

How many records are logged in the log?
Which environment is set to use the `Warning` level?

P.S. If the `Warning` level is set in the Testing (Staging) environment, then Production should be set to `Error` level, and your logs won't contain this message. However, if Production is set to `Warning` level, then your logs will indeed be full of these messages.

PaulARoy:
Hello raman,

I indeed see these logs in staging and production environment, and it's only a problem because it sends warning to my error tracking platform (Sentry).

We log one warning line per call of aggregate function that is missing a `Content-Length` header, but I initially thought it was an abnormality: it is not. All bodyless queries are missing the `Content-Length` header, which means all GET methods to an aggregate logs a Warning line, which is incorrect.

So my initial asumption to track it as a warning was wrong, this is a completely normal code path. Thus downgrading the log level to `Information` instead of `Warning`.

I still think it's relevant to log it as info.

Don't hesitate if you have any more questions.

PaulARoy:
Additional info: from this decision tree, I initially believed it was an unwanted state from which the process could continue (Warning), but we are not in an unwanted state (it's a normal situation), thus the Information level.

log_levels

raman-m:
Paul, this decision-making diagram is excellent! 😃 However, let's not suggest implementing a user's role in the configuration JSON. 🤣 Just kidding...

We will switch to the `Information` level to minimize log output. I would like to draw attention to [line 266](https://github.com/PaulARoy/Ocelot/blob/fix/remove-body-copy-warning/src/Ocelot/Multiplexer/MultiplexingMiddleware.cs#L266), where it may be prudent to also lower the logging level. @ggnaegi, what are your thoughts?

@PaulARoy, my plan is to merge it in the upcoming regular monthly release slated for [September'24](https://github.com/ThreeMammals/Ocelot/milestone/7). Our team will be quite busy in September as we have two pending releases: the [v23.3 Hotfixes](https://github.com/ThreeMammals/Ocelot/milestone/8) and the [Annual 2023](https://github.com/ThreeMammals/Ocelot/milestone/3). The latter is anticipated to be a swift release around mid-September, hopefully.
**How critical is this minor fix for you?** What's the priority?

PaulARoy:
Thanks raman

Regarding line 266 I'd thought about it and I'd keep the Warning level: this is an abnormal situation where somewhere someone has started reading the body. This shouldn't be happening and will cause the downstream to miss on the body (so continue on an abnormal state => Warning).

Honestly this is a bit critical for me as it logs a warning on each GET aggregate call (most of them). It saturates the logs and the bug tracking tool and I had to disable the tool entirely for now. While it doesn't break anything, it makes it a bit less production-ready.

Considering how little it is I'd appreciate if you could take this in the next release, but it is not completely blocking.

raman-m:
Understood regarding line 266. Let's keep it as Warning.

> Honestly this is a bit critical for me as it logs a warning on each GET aggregate call (most of them).

No problem to prioritize! The PR has been included in the ongoing [v23.3 Hotfixes](https://github.com/ThreeMammals/Ocelot/milestone/8) release.

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
```diff
@@ -276,7 +276,7 @@ protected virtual async Task<Stream> CloneRequestBodyAsync(HttpRequest request,
         }
         else
         {
-            Logger.LogWarning("Aggregation does not support body copy without Content-Length header!");
+            Logger.LogInformation("Aggregation does not support body copy without Content-Length header, skipping body copy.");
```
raman-m, line 279: I recommend implementing a minor decision-making process based on the HTTP verb
- If the verb is GET, then it should be logged at the `Information` level, as should HEAD, OPTIONS, and other verbs that do not have a body.
- However, if the verb is PUT, POST, etc., then we should log at the `Warning` level.

This approach should suffice to determine the appropriate logging level.

raman-m (commented):
Currently, the GET verb is permitted for the Aggregation feature. However, it appears that we will soon introduce additional verbs, see PR #1389, thus relaxing the strict GET requirement. Let's ensure we don't overlook this update.

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
PaulARoy: Hi raman,

Hard disagree for me on this, as the presence or absence of body is not specified in the HTTP spec.

It's an entirely normal and supported scenario to have a PUT or POST query without body and it stills holds a meaning (https://lists.w3.org/Archives/Public/ietf-http-wg/2010JulSep/0276.html)

So I wouldn't log it as a warning and I wouldn't make a verb-specific logic here.

raman-m: Fair enough! Not an issue.

raman-m, line 279: The message lacks route details. Including the route name would facilitate understanding of the failed route when reviewing the log.
Examples can be found in the repository here: [RouteName](https://github.com/search?q=repo%3AThreeMammals%2FOcelot%20RouteName&type=code).

raman-m (changes requested):

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
PaulARoy: I added it please tell me if it's OK for you.

raman-m (approved):
Paul, I appreciate your small contribution! 💟

src/Ocelot/Multiplexer/MultiplexingMiddleware.cs
raman-m: Fixed in https://github.com/ThreeMammals/Ocelot/pull/2146/commits/e76c3b44de70692468877b4409abf7b9d8253dbf ✔️

PR #2174, opened 2024-10-14
#2168 A brand-new `WatchKube` provider for Kubernetes service discovery

### LINKED ISSUE #163: Watch request fails with InvalidOperationException
There is a log call introduced in https://github.com/tintoy/dotnet-kube-client/commit/8cda77d93971eba77c3817c5a12f169442652f87 that uses PathAndQuery property of created request.
https://github.com/tintoy/dotnet-kube-client/blob/1e08492a6da6ce484ae4b9e5e2ec66a38fe802da/src/KubeClient/ResourceClients/KubeResourceClient.cs#L596-L598
But HttpRequest created by factory is relative URI, so on attempts to call ObserveLines I'm getting exception in Observable OnError:
```Unexpected error while streaming from the Kubernetes API to \"watch v1/Endpoints 'some-service' in namespace dev\".","exceptions":["System.InvalidOperationException: This operation is not supported for a relative URI.","   at System.Uri.get_PathAndQuery()","   at KubeClient.ResourceClients.KubeResourceClient.<>c__DisplayClass30_0.<<ObserveLines>b__0>d.MoveNext()```

### PR DESCRIPTION
**Pull request by kick2nick**
## Closes #2168 
- #2168 

## Proposed Changes
- `WatchKube` service discovery provider that utilizes [k8s watch api](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes) requests for efficient detection of changes.

It's proof of concept that needs improvements (see comments) and tests.
Tested locally, it works as expected, but with older versions of `KubeClient` which can be used with newer one after fix https://github.com/tintoy/dotnet-kube-client/issues/163.

p.s. sorry for mess with https://github.com/ThreeMammals/Ocelot/pull/2173
Had no intention to force downgrade of KubeClient, will resume after fix)

### CODE REVIEW

**Comment by raman-m** [comment]
I've rebased the branch. Don't push please❗
But if you want to commit more, then you have to delete branch, fetch all, checkout the branch locally 👉 
- `git switch develop`
- `git branch -D 2168-watchkube`
- `git fetch --all`
- `git pull`
- `git checkout 2168-watchkube`

And you will be able to contribute more 😉

**Comment by kick2nick** [comment]
Hi @raman-m , from my point of view all required stuff is done for moving to code review. Or am I missing something?

**Comment by raman-m** [comment]
The issue found at https://github.com/tintoy/dotnet-kube-client/issues/163 was resolved by https://github.com/tintoy/dotnet-kube-client/pull/164 and included in the release [v2.5.12](https://github.com/tintoy/dotnet-kube-client/releases/tag/v2.5.12). However, [this release has not been marked as the latest on the repository's releases page](https://github.com/tintoy/dotnet-kube-client/releases), where [v2.5.10](https://github.com/tintoy/dotnet-kube-client/releases/tag/v2.5.10) is still listed as the most recent version. Could you please explain why [v2.5.10](https://github.com/tintoy/dotnet-kube-client/releases/tag/v2.5.10) is considered the latest release?

**Comment by raman-m** [comment]
>from my point of view all required stuff is done for moving to code review. Or am I missing something?

If you're confident and have completed the main development phase, can't you press the "**Ready for Review**" button?

**Comment by kick2nick** [comment]
> Could you please explain why [v2.5.10](https://github.com/tintoy/dotnet-kube-client/releases/tag/v2.5.10) is considered the latest release?

Have no idea, let's wait for the author's answer to your question.

**Comment by raman-m** [comment]
Hello, Nikolay!

While it's true that the PR seems partially ready for review following the closure of https://github.com/tintoy/dotnet-kube-client/issues/163 and its subsequent release of [v2.5.12](https://github.com/tintoy/dotnet-kube-client/releases/tag/v2.5.12), I am concerned that reviewing this PR may be premature due to the following dependencies:
1. This PR relies on #2180, which needs to be released beforehand as a part of the next [23.3.6](https://github.com/ThreeMammals/Ocelot/releases/tag/23.3.6) patch.
2. My plan is to include it in the [November'24](https://github.com/ThreeMammals/Ocelot/milestone/9) milestone, which will prioritize the .NET 9 upgrade and migration to the `net9.0` framework, with a lower priority assigned to introducing new features.

**Review by raman-m — CHANGES_REQUESTED** [review]
**Testing the new `WatchKube` type in [ShouldReturnServicesFromK8s](https://github.com/ThreeMammals/Ocelot/pull/2174/files#diff-857f76beee3cb7c8488c15385322a6f55f272bc7d8e529ccc9f9bef5e1c692a9R51-R52) is necessary, but it is not sufficient for completing acceptance testing.**

While it is well-tested by unit tests, there are no new acceptance tests for the proposed feature. Please review the scenarios of the current `Kube` and `PollKube` providers for inspiration. If Not, I will share my ideas later, in 1-2 weeks, after the release of the v23.3.6 patch.

**Comment by kick2nick** [comment]
> This PR relies on https://github.com/ThreeMammals/Ocelot/pull/2180, which needs to be released beforehand as a part of the next [23.3.6](https://github.com/ThreeMammals/Ocelot/releases/tag/23.3.6) patch.

Not quite "relies", but conflicts on rebase will apear, I'll resolve them after https://github.com/ThreeMammals/Ocelot/pull/2180 merging.

> My plan is to include it in the [November'24](https://github.com/ThreeMammals/Ocelot/milestone/9) milestone, which will prioritize the .NET 9 upgrade and migration to the net9.0 framework, with a lower priority assigned to introducing new features.

No problem, I'll be in touch when you get back to this

**Review by RaynaldM — APPROVED** [review]

**Comment by kick2nick** [comment]
Hi Raman, great news!
I've resolved the conflicts. Regarding the feature itself, I think it's done =) But you had some ideas for improving the acceptance tests last time - do you want to discuss them?

**Comment by raman-m** [comment]
> @kick2nick commented [on May 31](https://github.com/ThreeMammals/Ocelot/pull/2174#issuecomment-2923516708)

### Acceptance
I saw that the [mentioned acceptance test](https://github.com/ThreeMammals/Ocelot/pull/2174#pullrequestreview-2391961313) was converted to `Theory` by adding the new provider's type, and it has passed. To be fair, I've lost track of my previous idea for improvements... My current thought is that we could convert the remaining acceptance Facts to `Theory` by simply adding the `[InlineData(nameof(WatchKube))]` attribute.
Regarding the new behavior, you should identify which parts of the new provider are critical and determine the best way to test them, since you are the author. I believe that one acceptance test with an advanced scenario, specifically for `WatchKube`, will be an essential part of integration testing.

### Unit 
For now, what I am afraid of is [degraded unit test coverage](https://github.com/ThreeMammals/Ocelot/pull/2174#issuecomment-2923509786), which has dropped by 0.07% ❗ This means that all the new code you've written is not 100% covered by unit tests. That is not good!
Here are the problematic source files being reported by build [73929220](https://coveralls.io/builds/73929220) &rarr; **Ocelot.Provider.Kubernetes** &rarr; 
- Changed [EndPointClientV1.cs](https://coveralls.io/builds/73929220/source?filename=Ocelot.Provider.Kubernetes%2FEndPointClientV1.cs) contains the new added method: `IObservable<TResource> Watch(string, string, CancellationToken)` which is [not covered](https://coveralls.io/builds/73929220/source?filename=Ocelot.Provider.Kubernetes%2FEndPointClientV1.cs#L38) at all ❗ So, [develop](https://coveralls.io/builds/73860727/source?filename=Ocelot.Provider.Kubernetes%2FEndPointClientV1.cs) 92% decreased to 58%
- Changed [KubernetesProviderFactory.cs](https://coveralls.io/builds/73929220/source?filename=Ocelot.Provider.Kubernetes%2FKubernetesProviderFactory.cs) has new [line 34](https://coveralls.io/builds/73929220/source?filename=Ocelot.Provider.Kubernetes%2FKubernetesProviderFactory.cs#L34) which is not covered ❕ So, [develop](https://coveralls.io/builds/73860727/source?filename=Ocelot.Provider.Kubernetes%2FKubernetesProviderFactory.cs) 100% decreased to 94%
- New added [ObservableExtensions.cs](https://coveralls.io/builds/73929220/source?filename=Ocelot.Provider.Kubernetes%2FObservableExtensions.cs) has 75.0% but it should be 100% ❕
- New added [WatchKube.cs](https://coveralls.io/builds/73929220/source?filename=Ocelot.Provider.Kubernetes%2FWatchKube.cs) has 93.62% but it should be 100% ❕ 

Kindly dedicate more effort to developing unit tests 🙏 

> P.S. With your permission, I will invite the library expert to review this PR...

**Comment by raman-m** [comment]
@tintoy
Hello, Adam!
We are continuously working to improve our [Ocelot.Provider.Kubernetes](https://github.com/ThreeMammals/Ocelot/tree/develop/src/Ocelot.Provider.Kubernetes) project, which is based on your excellent [KubeClient](https://www.nuget.org/packages/KubeClient) package.

This pull request introduces the new `WatchKube` provider type (in addition to the existing `Kube` and `PollKube`), utilizing reactive features and design principles from your library. We hope this will allow the package to leverage more of your library’s powerful functionality.

**We would greatly appreciate your feedback and a code review with any suggestions you may have.**
As an external reviewer, please provide a brief message indicating your readiness, and I will proceed to officially request a code review.

>Best regards,
>Raman, Ocelot team

**Comment by tintoy** [comment]
Sure - I’m happy to review it, although I can’t guarantee I’ll have the time to do so before Wednesday). Is that acceptable?

**Comment by kick2nick** [comment]
> @raman-m commented [on June 1](https://github.com/ThreeMammals/Ocelot/pull/2174#issuecomment-2927446898)

Acceptance
Converted tests to theories and created a new test that covers the case of a service endpoint change during concurrent client requests.

Unit tests
All points are fixed.

File: src/Ocelot.Provider.Kubernetes/ObservableExtensions.cs [hunk 1]
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Ocelot.Provider.Kubernetes;
+
+public static class ObservableExtensions
+{
+    public static IObservable<TSource> RetryAfter<TSource>(this IObservable<TSource> source,
+        TimeSpan dueTime,
+        IScheduler scheduler)
+    {
+        return RepeatInfinite(source, dueTime, scheduler).Catch();
```
tintoy: I thought `.Catch()` was supposed to take a result sequence or handler function that produces a result sequence to be used if the source sequence encounters an error; what is the behaviour of `.Catch()` with no arguments?

tintoy (approved):
This looks good to me; minor question about the `.Catch()` operator, but not a blocker.

src/Ocelot.Provider.Kubernetes/ObservableExtensions.cs
kick2nick: It's [overload](https://github.com/dotnet/reactive/blob/ebab5fc37e8c0888f7b96107852a8794e9af1735/Rx.NET/Source/src/System.Reactive/Linq/Observable.Multiple.cs#L235) of `Catch()` that takes `IEnumerable` of `Observable`'s as argument and in case of  exception takes next `Observable` from the sequence. Was found [here](https://stackoverflow.com/questions/18978523/write-an-rx-retryafter-extension-method#:~:text=So%20if%20we%20take%20this%20approach%2C%20we%20can%20just%20add%20a%20delay%20after%20the%20first%20yield.) ).

tintoy: Ah, fair enough - just wanted to confirm that it’s only taking a single argument (the source sequence) and not any handler delegate or alternative sequence as a second argument (as the StackOverflow examples seem to use).

raman-m: Good `catch`, Adam! 😄

raman-m:
> @tintoy approved these changes [on June 4, 2025](https://github.com/ThreeMammals/Ocelot/pull/2174#pullrequestreview-2897990183)

Thank you, Adam, for reviewing the code!

----
@ggnaegi Gui, any suggestions?
This is a new type of provider, so at the very least, you could take a look at this important PR...

src/Ocelot.Provider.Kubernetes/WatchKube.cs
```diff
@@ -0,0 +1,88 @@
...
+using Ocelot.Values;
+using System.Reactive.Concurrency;
+using System.Reactive.Linq;
+
+namespace Ocelot.Provider.Kubernetes;
+
+public class WatchKube : IServiceDiscoveryProvider, IDisposable
+{
+    internal const int FailedSubscriptionRetrySeconds = 5;
+    internal const int FirstResultsFetchingTimeoutSeconds = 3;
```
raman-m: Why 5? Why 3? Why internal?
How will the user be able to configure this?

src/Ocelot.Provider.Kubernetes/ObservableExtensions.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Reactive.Concurrency;
+using System.Reactive.Linq;
+
+namespace Ocelot.Provider.Kubernetes;
+
+public static class ObservableExtensions
+{
+    public static IObservable<TSource> RetryAfter<TSource>(this IObservable<TSource> source,
```
raman-m: I always wonder when people think in terms of a beloved library 😄 
FYI &rarr; [Retry](https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/Infrastructure/DesignPatterns/Retry.cs) in [Infrastructure](https://github.com/ThreeMammals/Ocelot/tree/develop/src/Ocelot/Infrastructure)/[DesignPatterns](https://github.com/ThreeMammals/Ocelot/tree/develop/src/Ocelot/Infrastructure/DesignPatterns)

src/Ocelot.Provider.Kubernetes/ObservableExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+    private static IEnumerable<IObservable<TSource>> RepeatInfinite<TSource>(IObservable<TSource> source,
+        TimeSpan dueTime,
+        IScheduler scheduler)
+    {
+        yield return source;
+
+        while (true)
+        {
+            yield return source.DelaySubscription(dueTime, scheduler);
+        }
```
raman-m, line 17: Will it block a thread indefinitely until the web server is shut down?
In what context is the `Task` running?
I cannot evaluate this code thoroughly, but appears quite unusual... a subscription is essentially based on deeply encapsulated `Timer`(s) within the `IScheduler`, correct?

src/Ocelot.Provider.Kubernetes/WatchKube.cs
```diff
@@ -0,0 +1,88 @@
...
+        SetFirstResultsCompletedAfterDelay();
+        _subscription = CreateSubscription();
+    }
+
+    public virtual async Task<List<Service>> GetAsync()
+    {
+        // wait for first results fetching
+        await _firstResultsCompletionSource.Task;
+
+        if (_services is not { Count: > 0 })
```
raman-m: A lengthy negated Pattern Matching sentence disregarding basic boolean operators ❕ 
```suggestion
        if (_services.Count == 0)
```
Note, it can't be `null` because it is initialized in constructor. So, null-check is not required here.

src/Ocelot.Provider.Kubernetes/WatchKube.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            _logger.LogWarning(() => GetMessage("Subscription to service endpoints gave no results!"));
+        }
+
+        return _services;
+    }
+
+    private void SetFirstResultsCompletedAfterDelay() => Observable
+        .Timer(TimeSpan.FromSeconds(FirstResultsFetchingTimeoutSeconds), _scheduler)
+        .Subscribe(_ => _firstResultsCompletionSource.TrySetResult());
```
raman-m: God bless! It is not Publish-Subscribe pattern 🤣 which in fact is pure C# delegate object 🤣 

Regarding `Subscribe`... It returns `IDisposable`. Shouldn't we register this disposable object somehow, as a class field?

src/Ocelot.Provider.Kubernetes/WatchKube.cs
```diff
@@ -0,0 +1,88 @@
...
+
+    private IDisposable CreateSubscription() =>
+        _kubeApi
+            .EndpointsV1()
+            .Watch(_configuration.KeyOfServiceInK8s, _configuration.KubeNamespace)
+            .Do(_ => { }, ex => _logger.LogError(() => GetMessage("Endpoints subscription error occured."), ex))
```
raman-m: Is `Do` responsible for exception logging only, not processing right?

src/Ocelot.Provider.Kubernetes/WatchKube.cs
```diff
@@ -0,0 +1,88 @@
...
+            .RetryAfter(TimeSpan.FromSeconds(FailedSubscriptionRetrySeconds), _scheduler)
+            .Subscribe(
+                onNext: endpointEvent =>
```
raman-m: I am not a fan of anonymous delegates, even when named parameters are used to slightly enhance code readability.
Better to utilize local functions to capture the context variables more effectively.
So, local functions or class methods guarantee that we use named delegates.

src/Ocelot.Provider.Kubernetes/WatchKube.cs
```diff
@@ -0,0 +1,88 @@
...
+                {
+                    _services = endpointEvent.EventType switch
+                    {
+                        ResourceEventType.Deleted or ResourceEventType.Error => new(),
+                        _ when (endpointEvent.Resource?.Subsets?.Count ?? 0) == 0 => new(),
+                        _ => _serviceBuilder.BuildServices(_configuration, endpointEvent.Resource).ToList(),
+                    };
```
raman-m: I believe it will never assign `null` to `_services`, right?

raman-m (changes requested):
The code looks good to me, but let's talk a bit more and then merge this PR.
> I'm going to push one commit with my review which will give my approval.

src/Ocelot.Provider.Kubernetes/WatchKube.cs
kick2nick: Right, just log and pass item further

src/Ocelot.Provider.Kubernetes/WatchKube.cs
kick2nick: Nope, it's not 'classic' `Disposable`, it's a way to cancel a subscription. There is no case where we need cancel this one.

src/Ocelot.Provider.Kubernetes/WatchKube.cs
kick2nick: Made `internal` for reference from Tests (with `InternalsVisibleTo` attribute). Numbers are made up =) Didn't want to make configuration complicated. We can move it to configuration, but I think `FailedSubscriptionRetrySeconds ` should stay as `const`, because `KubeClient` already [uses](https://github.com/tintoy/dotnet-kube-client/blob/64d2a8f1e27067182374bf8ec5155cb02d5b7d3c/src/KubeClient/ResourceClients/KubeResourceClient.cs#L738) retries. Our retry was added for **really** unexpected errors.

src/Ocelot.Provider.Kubernetes/ObservableExtensions.cs
kick2nick: > Will it block a thread indefinitely until the web server is shut down?

No, `yield return source.DelaySubscription(dueTime, scheduler)` get called only when previous Observable throws exception. Iteration on IEnumerable stops on shutdown or when current subscription cancelled (disposed).

> In what context is the Task running?

Which one? We have no tasks here.

> I cannot evaluate this code thoroughly, but appears quite unusual... a subscription is essentially based on deeply encapsulated Timer(s) within the IScheduler, correct?

This code is not about subscription, it's about Observables. Observable is deferred - it means that it starts 'working' only when someone create subscription( calls Subscribe). In this IEnumerable we create infinite sequence of identical observables but with delayed subscription (observable start 'working' after Subscribe() call + delay). `IScheduler` passed here (and everywhere) only for tests purposes =) With `IScheduler` we can [turn time back and forth](https://introtorx.com/chapters/testing-reactive-extensions-for-dotnet) 😈

tintoy: This code may look a little odd, in isolation, but it’s idiomatic for Rx. In Rx, almost everything is expressed as a sequence (with either “pull” semantics like `IEnumerable`, or “push” semantics like `IObservable`). Some of the operators for recovery take a source sequence and use a second sequence to express that recovery behaviour (say, for example, the second sequence being a sequence of sequences representing sequences to be appended to the original sequence based on some condition - which is where the error condition comes in); as long as that second sequence keeps yielding sequences, the original sequence continues. When it terminates, the original sequence terminates.

src/Ocelot.Provider.Kubernetes/ObservableExtensions.cs
raman-m: This class is a potential candidate for migration to Ocelot Infrastructure. Currently, its utility is limited to within the library itself.
Not an issue! ☑️

src/Ocelot.Provider.Kubernetes/WatchKube.cs
raman-m: Fixed in https://github.com/ThreeMammals/Ocelot/pull/2174/commits/81d38d247ffe25f14e125d5e97b26df748ba1eab ✔️ &rarr; [line R48](https://github.com/ThreeMammals/Ocelot/pull/2174/commits/81d38d247ffe25f14e125d5e97b26df748ba1eab#diff-b2a6fbc4016ce913723fc1c295b0d54bf18639cd68ad316926be90ac4678d3ad)

src/Ocelot.Provider.Kubernetes/WatchKube.cs
raman-m: 🆗 Got it! The infinite subscription.
If a registration for disposing is not required, then **it is not an issue!** ☑️

src/Ocelot.Provider.Kubernetes/WatchKube.cs
raman-m: Good! I've rewritten the line a bit in commit https://github.com/ThreeMammals/Ocelot/pull/2174/commits/81d38d247ffe25f14e125d5e97b26df748ba1eab in favor of named delegate, aka `OnException`.
Fixed in https://github.com/ThreeMammals/Ocelot/pull/2174/commits/81d38d247ffe25f14e125d5e97b26df748ba1eab ✔️ &rarr; [line R82](https://github.com/ThreeMammals/Ocelot/pull/2174/commits/81d38d247ffe25f14e125d5e97b26df748ba1eab#diff-b2a6fbc4016ce913723fc1c295b0d54bf18639cd68ad316926be90ac4678d3adR82)

src/Ocelot.Provider.Kubernetes/WatchKube.cs
raman-m: Added named delegates in commit https://github.com/ThreeMammals/Ocelot/pull/2174/commits/81d38d247ffe25f14e125d5e97b26df748ba1eab.
Fixed in https://github.com/ThreeMammals/Ocelot/pull/2174/commits/81d38d247ffe25f14e125d5e97b26df748ba1eab ✔️ &rarr; [line R84](https://github.com/ThreeMammals/Ocelot/pull/2174/commits/81d38d247ffe25f14e125d5e97b26df748ba1eab#diff-b2a6fbc4016ce913723fc1c295b0d54bf18639cd68ad316926be90ac4678d3adR84)

src/Ocelot.Provider.Kubernetes/ObservableExtensions.cs
raman-m: Well, guys... I want to eat oranges 🟠 and chew mint bubblegum... I'm listening...

The folks at Microsoft have reinvented a wheel called the [Observer design pattern](https://learn.microsoft.com/en-us/dotnet/standard/events/observer-design-pattern) with advanced features: essentially a refined version of the [Gang of Four](https://en.wikipedia.org/wiki/Design_Patterns)'s "[Observer](https://en.wikipedia.org/wiki/Observer_pattern)" behavioral pattern. However, it remains the classic [Publish-Subscribe](https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern) architectural pattern (PubSub), a projection of, with terminology setting them apart.

My understanding is that, over the past [10 years](https://www.nuget.org/packages/System.Reactive#versions-body-tab), Microsoft has been advocating the use of the [Observer design pattern](https://learn.microsoft.com/en-us/dotnet/standard/events/observer-design-pattern) as the next evolution of [Raising and Consuming Events](https://learn.microsoft.com/en-us/dotnet/standard/events/how-to-raise-and-consume-events) design model, which was introduced in the first release of .NET 1.0 in 2001-2002. This event-driven programming model is fundamentally built on two C# keywords: `event` and `delegate`.

The class that defines an **event** is the Publisher, and the class that adds a **delegate** object to the **event** is the Subscriber. A C# class with events is responsible for producing event data (aka `EventArgs`). Pretty simple, pretty clear… yet developers keep reinventing the wheel! 😄 And I’ve been wondering for 20 years, while working on outsourced projects, how more than 50% of them never defined even a single event.
Well… I always wonder why developers advocate for their favorite libraries instead of taking the time to learn a few new keywords in the programming language they use. Bingo! 💥

---
#### P.S.
When I asked Copilot AI:
> C# In what version (.NET release) the 'event' keyword was implemented?

It answered:
>The `event` keyword has been part of C# since its initial release in **.NET Framework 1.0** (2002). It is a fundamental feature of the language, enabling event-driven programming through **delegates** and **event handlers**.

So, true...

#### P.P.S.
A sad fact is that Ocelot doesn't utilize events, only delegates. However, this situation will change after clearing up [the backlog](https://github.com/ThreeMammals/Ocelot/issues). This year, I plan to incorporate an event-driven approach in Ocelot services more, as it has been ignored since the foundation of Ocelot project.

raman-m: Regarding the `while (true)` infinite loop...
I'm okay with leaving it as is after your explanations, because it implements the magic of both `IEnumerable` and `IObservable`. Since this infinite loop doesn't block any running threads, I would say **it's not an issue**.

src/Ocelot.Provider.Kubernetes/WatchKube.cs
raman-m: I have no objection to the Retry logic However, my concern lies with the unconfigurable numbers. I believe that a configuration design can be deferred for now and introduced at a later stage, in a separate pull request.
An acceptable compromise would be to enable configuration directly in C# code.

#### Next step 👇 
Therefore, I propose converting the internal constants into public static properties, allowing developers to set custom values for advanced user scenarios directly in C# code, given that configuration properties are not implemented now to be defined in `ocelot.json`, aka global [ServiceDiscoveryProvider](https://ocelot.readthedocs.io/en/latest/search.html?q=ServiceDiscoveryProvider) section.

#### The Q ❔ 
What I consider inefficient and worth discussing further is the Seconds dimension. Why not Milliseconds? As you may know, some deployed environments operate at the millisecond level — 10, 100, or 500 ms — clearly less than one second. For now, I propose decreasing the default values to one second as the minimum for definitions, but I believe that for highly rapid systems operating in milliseconds, one second won't significantly impact response time.
Ultimately, I would suggest that waiting 5 seconds between retries is excessive 😉 
_Do you have any use cases from your project, Kubernetes with the Ocelot environment in production, based on data from monitoring or traced logs?_
#### 2nd Q  
The 2nd question: Do both properties depend on each other? It seems the second property of fetching the first result is merely a delay in obtaining the initial result from the Observable layer, correct? This delay does not appear to be related to the retry logic, is that right?
In case they are both dependent, we have to think carefully about their values ​​in seconds, avoiding any disruptions.

raman-m: So, my proposal is this commit &rarr; https://github.com/ThreeMammals/Ocelot/pull/2174/commits/061f018bda255ddb3e550e49d2921b4fe0ad482f
However, I am uncertain about the following:
- Returning an `int` type that represents the value of the `TimeSpan.FromSeconds(double)` argument: it has the `double` parameter
- The default values of `5` and `3` seconds for the backing fields should be reduced, for instance, they could be set to `1`

kick2nick: > Therefore, I propose converting the internal constants into public static properties, allowing developers to set custom values for advanced user scenarios directly in C# code

reasonable

> What I consider inefficient and worth discussing further is the Seconds dimension. Why not Milliseconds?

10 ms for retries could create high pressure on k8s =) 10 ms = 100 rps. 100 rps * downstream services count * ocelot instances.
First result timeout is time that client can wait before first results fetching, and it can vary from couple milliseconds and up to defined timeout. So it's good candidate for configuration. Will try to find some numbers from prod env.

> Do you have any use cases from your project, Kubernetes with the Ocelot environment in production, based on data from monitoring or traced logs

No, as I said KubeClient already retries subcription in case of network failures. You can think about retry logic introduced in this PR like "try -> catch -> retry all nested stuff inside KubeClient whatever it was under the hood". This 'whatever' could be bug or anything else in KubeClient logic.

> This delay does not appear to be related to the retry logic, is that right?

Right,  there is no dependency between properties. First results is important part, as far as discovery provider instantiated during first call.

> So, my proposal is this commit.

Agree with your proposal.

src/Ocelot.Provider.Kubernetes/ObservableExtensions.cs
kick2nick: Comparing event/delegate to Observable is like comparing a hammer to a microscope 😄 

> This year, I plan to incorporate an event-driven approach in Ocelot services more, as it has been ignored since the foundation of Ocelot project.

Hmmmm, ask a Copilot "Why c# event api is not actively used in modern asp.net apps?" There is a bunch of reasons, like: poor subscriptions management, concurrency issues, etc.

raman-m: > ask a Copilot "Why c# event api is not actively used in modern asp.net apps?"

Good, clever, boy.

> it...  is like comparing a hammer to a microscope

At the very least, we can compare the sizes of coded files and the overall complexity of implementation, which is critical for the Ocelot project, where no one wants to implement efficient solutions, just simple and functional ones.

src/Ocelot.Provider.Kubernetes/WatchKube.cs
raman-m: FYI I have assigned def values (both to 1 second) in commit https://github.com/ThreeMammals/Ocelot/pull/2174/commits/28a3afe69b34c332a776b66cd5712d2f5babdc2d 💡

raman-m (approved):
### Ready for delivery ✅ 
- Code reviews ✔️✔️
- Unit testing ✔️
- Acceptance testing ✔️
- Updated docs ✔️

src/Ocelot.Provider.Kubernetes/WatchKube.cs
raman-m: > 10 ms for retries could create high pressure on k8s =) 10 ms = 100 rps. 100 rps * downstream services count * ocelot instances.
First result timeout is time that client can wait before first results fetching, and it can vary from couple milliseconds and up to defined timeout. So it's good candidate for configuration. **Will try to find some numbers from prod env.**

It appears this analysis is taking some time, and I am unable to wait any longer: the PR is [ready to merge](https://github.com/ThreeMammals/Ocelot/pull/2174#pullrequestreview-2932058669).
If there are further proposals for improvement, we will incorporate them into subsequent PRs, OK?

raman-m:
Merged as commit https://github.com/ThreeMammals/Ocelot/commit/8143d5608f354262d9a2a08e12baf36ef5e49eb0

PR #2180, opened 2024-10-21
#977 Enable validation of DI scopes in `Kube` and `PollKube` discovery providers

### LINKED ISSUE #977: Cannot resolve scoped service `KubeClient.IKubeApiClient` from root provider

redirect to service in k8s cluster

Follow this docs https://ocelot.readthedocs.io/en/latest/features/kubernetes.html

`services.AddOcelot(_configuration).AddKubernetes(false); // or true`

My service in k8s: devices-api.dev:10010

Config:
```json
{
    "ReRoutes": [
        {
            "DownstreamPathTemplate": "/{everything}",
            "DownstreamScheme": "http",
            "UpstreamPathTemplate": "/devices/{everything}",
            "ServiceName": "devices-api.dev",
            "UpstreamHttpMethod": [ "Get" ]
        }
    ],
    "GlobalConfiguration": {
        "ServiceDiscoveryProvider": {
            "Namespace": "dev",
            "Type": "kube"
        }
    }
}
```

## Specifications

  - Version: Ocelot 13.5.2, Ocelot.Provider.Kubernetes 13.5.2
  - Platform: Docker
  - Subsystem: ASP.NET Core that runs in docker container in k8s cluster

kick2nick:
## Fixes #977 
- #977 

## Proposed Changes

  - Added unit tests for verifying that `Kube` & `PollKube` providers can be resolved from DI with enabled scope validation.
  - Fixed Kubernetes acceptance tests. Used registered by Ocelot extension `KubeApiClient` instead of manually created one. Enabled scope validation for ServiceProvider
  - Made `IKubeApiClient` singleton lifetime explicit.

## Successor
- #2174


kick2nick:
@raman-m,
please see https://github.com/ThreeMammals/Ocelot/issues/977#issuecomment-2427168671

src/Ocelot.Provider.Kubernetes/Ocelot.Provider.Kubernetes.csproj
```diff
@@ -26,7 +26,6 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="KubeClient" Version="2.5.10" />
-    <PackageReference Include="KubeClient.Extensions.DependencyInjection" Version="2.5.10" />
```
raman-m, line 29: Why did you remove this pack reference?

src/Ocelot.Provider.Kubernetes/OcelotBuilderExtensions.cs
```diff
@@ -9,10 +11,23 @@ public static class OcelotBuilderExtensions
     public static IOcelotBuilder AddKubernetes(this IOcelotBuilder builder, bool usePodServiceAccount = true)
     {
         builder.Services
-            .AddKubeClient(usePodServiceAccount)
+            .AddSingleton<IKubeApiClient, KubeApiClient>(KubeApiClientFactory)
```
raman-m, line 14: **It appears to be a breaking change 🤯 Lovely singletons once more❗**
The main purpose of the `Ocelot.Provider.Kubernetes` NuGet package is to facilitate the integration of the `KubeClient` package with `Ocelot`. The proposed changes undermine this objective by removing `AddKubeClient`. This design direction is flawed because our solutions should be developed based on the official `KubeClient` package's functionality, not by altering the library's behavior to suit our needs!
You have to understand that the issue is not `AddKubeClient` extension, but singleton factory aka `KubernetesProviderFactory.Get` with bad scope management!

src/Ocelot.Provider.Kubernetes/OcelotBuilderExtensions.cs
```diff
@@ -9,10 +11,23 @@ public static class OcelotBuilderExtensions
...
             .AddSingleton(KubernetesProviderFactory.Get)
             .AddSingleton<IKubeServiceBuilder, KubeServiceBuilder>()
             .AddSingleton<IKubeServiceCreator, KubeServiceCreator>();
         return builder;
+
+        KubeApiClient KubeApiClientFactory(IServiceProvider sp)
```
raman-m, line 20: Factory, factory, factory... 🙈 
If it's a singleton factory, why not move this functionality to a singleton/static `KubernetesProviderFactory` class?
#### FYI
The static `KubernetesProviderFactory` class (aka singleton) will be refactored in the future to become a genuine `IServiceDiscoveryProvider` object through inheritance from the `IServiceDiscoveryProvider` interface.
**Please take note of this future change.**

test/Ocelot.AcceptanceTests/ServiceDiscovery/KubernetesServiceDiscoveryTests.cs
```diff
@@ -311,15 +309,24 @@ private void GivenThereIsAFakeKubernetesProvider(EndpointsV1 endpoints, bool isS
...
     private void WithKubernetes(IServiceCollection services) => services
-        .AddOcelot().AddKubernetes()
-        .Services.RemoveAll<IKubeApiClient>().AddSingleton(_clientFactory);
+        .AddOcelot().AddKubernetes(false)
+        .Services.Configure(_kubeClientOptionsConfigure)
+        .Replace(GetValidateScopesDescriptor());
 
     private void WithKubernetesAndRoundRobin(IServiceCollection services) => services
-        .AddOcelot().AddKubernetes()
+        .AddOcelot().AddKubernetes(false)
```
raman-m, line 317: Seems we can reuse `WithKubernetes` at the beginning of the method.
After that replace the rest of services.
```suggestion
        WithKubernetes(services);
        services.BlaBlaReplacements()...
```

test/Ocelot.AcceptanceTests/ServiceDiscovery/KubernetesServiceDiscoveryTests.cs
```diff
@@ -311,15 +309,24 @@ private void GivenThereIsAFakeKubernetesProvider(EndpointsV1 endpoints, bool isS
...
         .AddCustomLoadBalancer<RoundRobinAnalyzer>(GetRoundRobinAnalyzer)
         .Services
-        .RemoveAll<IKubeApiClient>().AddSingleton(_clientFactory)
-        .RemoveAll<IKubeServiceCreator>().AddSingleton<IKubeServiceCreator, FakeKubeServiceCreator>();
+        .Configure(_kubeClientOptionsConfigure)
+        .RemoveAll<IKubeServiceCreator>().AddSingleton<IKubeServiceCreator, FakeKubeServiceCreator>()
+        .Replace(GetValidateScopesDescriptor());
```
raman-m, line 322: It appears that the `Configure` and `Replace` methods ought to be positioned more closely together.
```suggestion
        .Configure(_kubeClientOptionsConfigure)
        .Replace(GetValidateScopesDescriptor());
        .RemoveAll<IKubeServiceCreator>().AddSingleton<IKubeServiceCreator, FakeKubeServiceCreator>()
```

test/Ocelot.UnitTests/Kubernetes/KubernetesProviderFactoryTests.cs
```diff
@@ -0,0 +1,62 @@
...
+
+    public KubernetesProviderFactoryTests()
+    {
+        _builder = new ServiceCollection().AddOcelot();
+    }
+
+    [Theory]
+    [Trait("Bug", "977")]
+    [InlineData(nameof(PollKube))]
+    [InlineData(nameof(Kube))]
```
raman-m, line 25: I suggest splitting it into two `Fact`s because long polling requires separate testing.

test/Ocelot.AcceptanceTests/ServiceDiscovery/KubernetesServiceDiscoveryTests.cs
raman-m, line 1: I'm unable to locate the repro-test that process the original `InvalidOperationException` from issue #977.

raman-m (changes requested):
1. I'm not as confident as you; the proposed changes seem to be breaking ones.
2. There's no acceptance test for the original user scenario in bug #977.
  In BDD, it's quite clear that we should test it first before any actual refactoring.
4. Why were so many changes proposed for a minor scope validation change? I don't understand.

I'm convinced that scope validation can be implemented with less code.

raman-m:
>@kick2nick commented [on Oct 21](https://github.com/ThreeMammals/Ocelot/pull/2180#issuecomment-2427657077)

Initially, I was uncertain if the `provider.CreateScope()` approach was ideal for our needs until I saw the `HttpContext.RequestServices` in PR #2179. To be honest, I'm not concerned with Sergio's solutions as they addressed issues in an older version of Ocelot, which may not be applicable to the current versions. We should consider your suggestions only if there is a consensus within the team.

> I have some doubts about this approach:
> - HttpClient MessageHandlers won't be reused - KubeClient [creates](https://github.com/tintoy/dotnet-kube-client/blob/84b055c885a2afe00781559ea400c0bd8cdfce6d/src/KubeClient/KubeApiClient.cs#L213) new one for each KubeApiClient instance, that can cause socket exhaustion issues on heavy loads.
> 

Excellent! How did you validate this hypothesis? Theoretical assumptions differ from empirical testing. Additionally, as previously mentioned in this message, we will pursue alternative solutions, setting aside my `provider.CreateScope()` design recommendation, as it appears there are superior design options based on `HttpContext`.

> - This scope is separate from request HttpContext scope and no services would be potentially reused.

Using `HttpContext.RequestServices` seems to be the appropriate method, though the consequences for the `PollKube` provider are still unclear to me.

> - There is no option to use long polling with this approach, HttpClient will be [disposed](https://github.com/tintoy/dotnet-kube-client/blob/84b055c885a2afe00781559ea400c0bd8cdfce6d/src/KubeClient/KubeApiClient.cs#L49C9-L49C30) with scope
> So my opinion is better leave KubeApiClient lifetime as is (it's now promoted to singleton), but make it explicit by registering it as singleton.
> 

If it's not implemented, why are we discussing it? You're correct to assume that we might encounter issues with disposed objects. Couldn't this be managed by `HttpContext`? I'm quite confident that using `HttpContext.RequestServices` would prevent any disposed objects, eliminating the need to manage scope lifetime.

> But if you insist on Sergio's approach than no problem, I'll implement it :)

Please disregard the previous approach for Sergio; we will devise an alternative solution. I find these debates to be unproductive.

test/Ocelot.UnitTests/Kubernetes/KubernetesProviderFactoryTests.cs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    private static ServiceProviderConfiguration GivenServiceProvider(string type) => new(
+        type: type,
+        scheme: string.Empty,
+        host: string.Empty,
+        port: 1,
+        token: string.Empty,
+        configurationKey: string.Empty,
+        pollingInterval: 1);
```
raman-m, line 59: Are you certain about enabling long polling for both `Kube` and `PollKube` providers? That is the reason I suggested dividing one theory into two facts. If the `pollingInterval` setting is disregarded during `Kube` testing, then disregard this notice.

src/Ocelot.Provider.Kubernetes/OcelotBuilderExtensions.cs
kick2nick: > It appears to be a breaking change 🤯

Why is it breaking change?
✔️  Public API hasn't changed
✔️  IKubeApiClient has the same singleton lifetime as it had before this fix(implicitly promoted)
Proposed change - make IKubeApiClient singleton lifetime explicit by registering it as singleton.

> our solutions should be developed based on the official KubeClient package's functionality, not by altering the library's behavior to suit our needs!

`KubeClient` package's functionality - talking to kube api as httpclient wrapper. I don't understand why it's lifetime can't be managed by `Ocelot.Provider.Kubernetes`. (like it was done for [consul](https://github.com/ThreeMammals/Ocelot/blob/ce0227c059f220761acc2d880e554174bdb3c544/src/Ocelot.Provider.Consul/ConsulClientFactory.cs)).

src/Ocelot.Provider.Kubernetes/Ocelot.Provider.Kubernetes.csproj
kick2nick: Because this fix propose IKubeApiClient to be registered in DI without helper from this package.

test/Ocelot.UnitTests/Kubernetes/KubernetesProviderFactoryTests.cs
kick2nick: Yep, it's disregarded.

kick2nick:
> until I saw the HttpContext.RequestServices in PR https://github.com/ThreeMammals/Ocelot/pull/2179.

There you are resolving `IConsulServiceBuilder` **once** from the scope of the **first** http request to downstream service and then cache it along with [loadBalancer](https://github.com/ThreeMammals/Ocelot/blob/414f63439d32ed9a3c09a77c86f035ed9c34aa56/src/Ocelot/LoadBalancer/LoadBalancers/LoadBalancerHouse.cs#L9).
`IConsulServiceBuilder` is not IDisposable and it works (but you don't have truly scoped `IConsulServiceBuilder` ). 
[`The framework creates a scope per request, and RequestServices exposes the scoped service provider. All scoped services are valid for as long as the request is active.`](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-8.0#:~:text=The%20framework%20creates%20a%20scope%20per%20request%2C%20and%20RequestServices%20exposes%20the%20scoped%20service%20provider.%20All%20scoped%20services%20are%20valid%20for%20as%20long%20as%20the%20request%20is%20active.)

> Using HttpContext.RequestServices seems to be the appropriate method, though the consequences for the PollKube provider are still unclear to me.

Usual `Kube` will not work if IKubeApiClientFactory is resolved from `HttpContext.RequestServices` because HttpClient will be disposed. We have another option - inject IHttpContextAccessor into `Kube` and on each request resolve IKubeApiClient, but for PollKube it won't work (it calls `Kube` on timer ticks without HttpContext).

> Excellent! How did you validate this hypothesis? Theoretical assumptions differ from empirical testing.

There is whole concept - [HttpClientFactory](https://learn.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests) that adresses this issues. Please read first 2 paragraphs.

src/Ocelot.Provider.Kubernetes/OcelotBuilderExtensions.cs
kick2nick: It's just helper method like [original](https://github.com/tintoy/dotnet-kube-client/blob/84b055c885a2afe00781559ea400c0bd8cdfce6d/src/KubeClient.Extensions.DependencyInjection/ClientRegistrationExtensions.cs#L47)used to avoid inlining delegate into AddSingleton. Should I move it to separate class?)

raman-m:
What's up, man? Are you online?

kick2nick:
> What's up, man? Are you online?

Yes, I'm here. Fixed review comments, pointed why `HttpContext.RequestServices` can't be used. Anything else?

kick2nick:
> Couldn't this be managed by HttpContext? I'm quite confident that using HttpContext.RequestServices would prevent any disposed objects, eliminating the need to manage scope lifetime.

No! Because IKubeApiClient will be resolved once, stored inside Kube singleton, and then disposed after HttpContext gone. Provided test as proof.
https://github.com/kick2nick/Ocelot/commit/bc2f7e6975e2d983cdfa79ca7f91cfdfbb5b1b21
results:
image

src/Ocelot.Provider.Kubernetes/Ocelot.Provider.Kubernetes.csproj
kick2nick: Right, but if DI helpers don't meet requirements or cause negative consequences then custom helpers is appropriate alternative, IMO.

raman-m:
> Yes, I'm here. Fixed review comments, **pointed why HttpContext.RequestServices can't be used**.

Well done, my astute developer! I concur, `HttpContext.RequestServices` is not suitable for `Kube`, and your evidence supports this. I'm weary of these debates as well. It seems the upcoming debate session will be unproductive.

Why suggest refactoring everything for a minor issue that throws an `InvalidOperationException` with the message `Cannot resolve scoped service 'KubeClient.IKubeApiClient' from root provider`?

I've [asked](https://github.com/ThreeMammals/Ocelot/pull/2180#pullrequestreview-2384499359) you:
> I'm convinced that scope validation can be implemented with less code.

You didn't respond but instead advocated for your ideal solution, which I believe is commendable and, in my view, should have been implemented 3-5 years ago, utilizing a design based on singleton factories. Writing less code is possible by creating a fake scope and using it during the operation of reading the list of services, as [suggested here](https://github.com/ThreeMammals/Ocelot/issues/977#issuecomment-641235558), but sure you didn't like it. This approach is quite straightforward and minimalistic. However, sometimes proposing ideal solutions can be a waste of time. As [I have previously mentioned](https://github.com/ThreeMammals/Ocelot/pull/2180#discussion_r1810362967), both `Consul` and `Kube` classes will be transitioned to DI converting them to `IServiceDiscoveryProvider` objects, and the `ServiceDiscoveryFinderDelegate` will be eliminated forever.

> Anything else?

Code review has done.
Please halt any further development at this point; I will proceed with the final review of the PR...

raman-m:
### Successor
- #2174

src/Ocelot.Provider.Kubernetes/Ocelot.Provider.Kubernetes.csproj
raman-m: In the next PR #2174 you've updated tte package version, and seems the pack is required. 😄

raman-m: Oh, my gosh! The pack version is outdated! 🤣  See [v2.5.12](https://www.nuget.org/packages/KubeClient.Extensions.DependencyInjection/2.5.12)

raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/2180/commits/b8cd234cfaab3e436d92b3f751802f6fcffd0965

src/Ocelot.Provider.Kubernetes/OcelotBuilderExtensions.cs
raman-m: Fair enough, not an issue.

src/Ocelot.Provider.Kubernetes/Ocelot.Provider.Kubernetes.csproj
kick2nick: Helper from there doesn't used anymore and not needed for next PR

raman-m:
@kick2nick Nikolay,
Thank you for your contribution! 💪

PR #2114, opened 2024-06-28
#842 #1414 Introduce `AllowAnonymous` authentication option with global `AuthenticationOptions` configuration

jlukawska:
## Closes #842 #1414 
- #842 
- #1414 

## Proposed Changes
It is possible to configure `AuthenticationOptions` in `GlobalConfiguration`. Then all routes use these settings. To configure an exception to this rule (e.g. for identity service), `AllowAnonymous` property should be set to true in the route `AuthenticationOptions`.

If a route has its own `AuthenticationOptions` with `AuthenticationProviderKey`/`AuthenticationProviderKeys` configured additionally, it has priority over the global one.

At the moment if a route uses a global `AuthenticationProviderKey`/`AuthenticationProviderKeys` (when `AuthenticationProviderKey`/`AuthenticationProviderKeys` is not configured for route explicitly), it uses also global `AllowedScopes`, even if `AllowedScopes` is configured for the route additionally.

### Note
This PR was prepared 4 years ago &rarr;
- #1215

The revival of the Ocelot brought many changes, also tightly connected with the PR code. Now there is additional property: `AuthenticationProviderKeys`. The merge was so hard that I created the PR from scratch and tried to apply all the code review comments.

Issue #1587, opened 2022-07-18
Claim-based rate limiting


impratikpatel:
## New Feature
Extracting claims from the token and use as a Client ID in the rate limiting.
Suggesting a new feature to implement the rate limiting based on the claims in the token.

## Motivation for New Feature
Current rate limiting is based on the Client ID passed in the header from the request. There are chances that anyone can manipulate the request by updating headers and using APIs without any restriction. 

So, instead of depending on the consumer to provide the `Client ID` in the header, we can use the claims from the token. Which is more secure and not modifiable. Considering this, a rate limit will be applicable for authenticated endpoints only. 

## Specifications

  - Version: 18.1.0
  - Platform: .NET 6

ks1990cn:
Have you tried this thing? I am trying to replicate if is there any flaw.

Its a security related thing you want to ask,

But clientId feature is well known to be such that it cannot be guessed and its external user login system can intergrated.

For example, if you integrate with Google Oauth, then we you will be signing In which will get you token, so in case of any manipulation you will be getting unauthorized before rate limit issue.

Read more about Identity Server intergation.
https://ocelot.readthedocs.io/en/latest/features/authentication.html

@raman-m @ggnaegi Its security question here, can we move it to discussion first, if required?

raman-m:
@ks1990cn  Tanmay,
Thanks for your message!
I'm busy now with release now. We will discuss this message after release...
Ping me once again after release...

raman-m:
> @ks1990cn commented [on Nov 27](#issuecomment-1827496271):
@raman-m @ggnaegi Its security question here, can we move it to discussion first, if required?

Tanmay, are you sure the author has an intention to communicate and contribute? 😉 
There is no problem to convert it to a discussion... But the author is silent! Do you see?

raman-m:
> @ArchPratik commented [on Jul 18, 2022](#issue-1308098573)

Hi Pratik!
Welcome to the Ocelot world! 🐅 

After team discussion we've designed to accept the issue! We agree that current Header solution is ugly. We must switch to Remote IP to detect client's identity for anonymous routes...

I have no idea when we will start development, I hope in January'24. We still have a couple of professional senior developers who write high quality code... So, any your contributions are really welcomed!

evilz:
Hello all,
Does someone work on this ?

What is the specification for the configuration if we want to use a claim value as clientId ?

Let me know if I should work on this ;p

ArchPratik:
[...]

Issue #1050, opened 2019-11-04
Rate limit with RemoteIpAddress


EngRajabi:
## New Feature
- add rate limit with ip address
- add StackBlockedRequests
- add custom rate limit store(sql server, redis,...)
read header key in config
add IpWhitelist in config
read StackBlockedRequests in config

catcherwong:
For a custom rate limit store, you can write a handler that implements `IRateLimitCounterHandler` interface. 
 
Then you can replace the default implementation `MemoryCacheRateLimitCounterHandler` with your own handler.

https://github.com/ThreeMammals/Ocelot/blob/fbd9c01a425a0a3bda46717efda423883543fb76/src/Ocelot/DependencyInjection/OcelotBuilder.cs#L108

m-hicks-OH:
Just do an UpsteamHeaderTransorm
`      "UpstreamHeaderTransform": {
        "ClientId": "{RemoteIpAddress}"
      },`

eddex:
If you also need the `X-Forwarded-For` header in your downstream systems, you can configure Ocelot to use this header instead of the default `ClientId` header for rate limiting.

```json
{
  "GlobalConfiguration": {
    "RateLimitOptions": {
      "DisableRateLimitHeaders": false,
      "HttpStatusCode": 429,
      "ClientIdHeader" : "X-Forwarded-For"
    }
  },
  "Routes": [
    {
    ... // route configuration
    "UpstreamHeaderTransform": {
        "X-Forwarded-For": "{RemoteIpAddress}"
      }
    }
  ]
}
```

raman-m:
> @EngRajabi commented [on Nov 4, 2019](https://github.com/ThreeMammals/Ocelot/issues/1050#issue-517084333)

>    add rate limit with ip address

Now not implemented. And this feature could be with high priority.

--- 
>    add StackBlockedRequests

Sorry? What are talking about?

----
>    add custom rate limit store(sql server, redis,...)

Seems, not implemented. Will you contribute?

----
>    read header key in config

Sorry? Isn't it already implemented by [ClientIdHeader](https://github.com/search?q=repo%3AThreeMammals%2FOcelot%20ClientIdHeader&type=code) option? 

----
>    add IpWhitelist in config

Isn't it implemented by [ClientWhitelist](https://github.com/search?q=repo%3AThreeMammals%2FOcelot+ClientWhitelist&type=code) option?
[...]

PR #2188, opened 2024-11-06
#2138 Introduce ASP.NET rate limiting

### LINKED ISSUE #1229: Global rate limiting for all, multiple routes instead of every route
Add rate limit options for all requests in common
```json
{
  "Routes": [],
  "RateLimitOptions": [
    {
      "RouteKeys": [ "posts", "users" ],
      "ClientWhitelist": [],
      "EnableRateLimiting": true,
      "Period": "10s",
      "PeriodTimespan": 3,
      "Limit": 1     
    }
  ]
}
```
We add rate-limiting options for every route :
For e.g:
```json
{
  "DownstreamPathTemplate": "/users",
  "DownstreamScheme": "https",
  "DownstreamHostAndPorts": [
    {
      "Host": "jsonplaceholder.typicode.com",
      "Port": 443
    }
  ],
  "UpstreamPathTemplate": "/users",
  "RateLimitOptions": {
    "ClientWhitelist": [],
    "EnableRateLimiting": true,
    "Period": "10s",
    "PeriodTimespan": 3,
    "Limit": 1
  }
}
```

### LINKED ISSUE #1915: Apply rate limiting globally
**this is my configuration that is not working globally**

```
"GlobalConfiguration": {
    "BaseUrl": "https://localhost:5002",
    "RateLimitOptions": {
      "DisableRateLimitHeaders": true,
      "QuotaExceededMessage": "Exceed ",
      "HttpStatusCode": 400,
      "ClientWhitelist": [],
      "EnableRateLimiting": true,
      "Period": "10s",
      "PeriodTimespan": 3,
      "Limit": 1
    }
  }
```
  
  **when i'm adding this to a specific route it is working**
  
```
  "RateLimitOptions": {
      "ClientWhitelist": [],
      "EnableRateLimiting": true,
      "Period": "10s",
      "PeriodTimespan": 3,
      "Limit": 1
    }
```
 
**is there any option to setup rate limiting globally that will affect to my all endpoints?**

yjorayev:
## Closes #2138 
- #2138 

### Discussion thread
- https://github.com/ThreeMammals/Ocelot/discussions/2138#discussioncomment-11116502 

## Proposed Changes

  - Includes changes to use .Net middleware
  - Middleware type for rate limiting can be specified in `ocelot.json`. It defaults to Ocelot's existing rate limiter if not specified
  - Also added some Unit Tests, Integration test and a sample

## Documentation block
- RST sources: [Rate Limiting | Future and ASP.NET Implementation](https://github.com/ThreeMammals/Ocelot/blob/develop/docs/features/ratelimiting.rst#future-and-aspnet-implementation)
- Html docs: [Rate Limiting | Future and ASP.NET Implementation](https://ocelot.readthedocs.io/en/latest/features/ratelimiting.html#future-and-asp-net-implementation)

## Predecessor
- #2294
  We need to merge this PR first due to the agreement on the JSON schema, which includes global configuration and specific `Metadata` options.


test/Ocelot.IntegrationTests/RateLimitingTests.cs
raman-m, line 1: These tests are actually acceptance tests, not integration tests. Integration tests typically meet the following criteria:
- They include at least one mocked interface object, also known as a fake, mock, or stub.
- They involve at least one integrated interface object that communicates based on either mocked or real testing data.

Since these tests lack mocked objects, they do not qualify as unit tests. However, because all integrated service interface objects communicate with stub or real objects, they are considered acceptance tests, also known as fitness or feature tests.

raman-m (changes requested):
**Thanks for PR creation!**

Please move current [test/Ocelot.IntegrationTests/RateLimitingTests](https://github.com/ThreeMammals/Ocelot/pull/2188/files#diff-917e7702b0ba318abb05c6b0d43712fa814313e15762dabd47b75d05dae91e1c) to [Ocelot.AcceptanceTests](https://github.com/ThreeMammals/Ocelot/tree/develop/test/Ocelot.AcceptanceTests) project, we already have a suitable class: [ClientRateLimitingTests](https://github.com/ThreeMammals/Ocelot/blob/develop/test/Ocelot.AcceptanceTests/RateLimiting/ClientRateLimitingTests.cs). However, these tests only cover the Ocelot feature "[Rate Limit by Client's Header](https://ocelot.readthedocs.io/en/latest/features/ratelimiting.html#rate-limit-by-client-s-header)." Since you are proposing a new feature, a new testing class should be created. 

**P.S.** I recommend (and require) inheriting from `Steps`, which offers many useful testing helpers, so there's no need to reinvent the wheel.

raman-m:
Yhlas, thank you once again. Here's the current status: I will request an official code review after transitioning tests to acceptance testing. Please remember to draft the preliminary [docs](https://github.com/ThreeMammals/Ocelot/blob/develop/docs/features/ratelimiting.rst). Regarding the delivery vision, I cannot include it in the current [Autumn'24](https://github.com/ThreeMammals/Ocelot/milestone/9) milestone due to anticipated multiple development rounds. Post-release, our team intends to upgrade Ocelot to .NET 9. It appears your feature may be incorporated into the release alongside this upgrade. Therefore, I suggest testing the PR code with the current .NET 9 RC. Is that okay? Consequently, we will target two frameworks in the project files: `net8.0` and `net9.0`

test/Ocelot.IntegrationTests/RateLimitingTests.cs
yjorayev: Thanks for pointing out about existence of Steps class. I used it as base class in my test and also moved it to AcceptanceTests project.

src/Ocelot/RateLimiting/Middleware/RateLimitingMiddleware.cs
```diff
@@ -35,6 +38,18 @@ public async Task Invoke(HttpContext httpContext)
                 return;
             }
 
+            #if NET7_0_OR_GREATER
+            if (options.RateLimitMiddlewareType == RateLimitMiddlewareType.DotNet)
+            {
+                //add EnableRateLimiting attribute to endpoint, so that .Net rate limiter can pick it up and do its thing
+                var metadata = new EndpointMetadataCollection(new EnableRateLimitingAttribute(options.RateLimiterPolicyName));
+                var endpoint = new Endpoint(null, metadata, "tempEndpoint");
```
raman-m, line 46: ### WTF? 😮 
We add endpoint temporarily? Common!

test/Ocelot.AcceptanceTests/RateLimiting/RateLimitingTests.cs
```diff
@@ -0,0 +1,110 @@
...
+#endif
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http;
+using Ocelot.Configuration;
+using Ocelot.Configuration.File;
+using Ocelot.DependencyInjection;
+
+namespace Ocelot.AcceptanceTests.RateLimiting
+{
+    public class RateLimitingTests: Steps
```
raman-m, line 14: Great! But you propose new feature for ASP.NET rate limiter...
```suggestion
    public class AspNetRateLimitingTests: Steps
```

raman-m (changes requested):
We can enhance the quality of the PR: my suggestions are outlined below 👇
**FYI**, I will revisit this PR once it has been reviewed by a team member as well.

**P.S.** Keep in mind [Line-Ending Gotchas aka EOL Fun](https://ocelot.readthedocs.io/en/latest/building/devprocess.html#line-ending-gotchas-aka-eol-fun). I saw some files have fake changes because of changed line endings.

test/Ocelot.AcceptanceTests/RateLimiting/RateLimitingTests.cs
ggnaegi: AspNet is an old concept in my opinion

ggnaegi:
@raman-m That's a great idea, but we are stuck with the .NET rate limiters limitations, you can't modify them on the fly, as you could with the current limiters. The rate limiters policies must be known, and therefore configured on startup

samples/RateLimiter/Program.cs
```diff
@@ -0,0 +1,26 @@
+using Microsoft.AspNetCore.RateLimiting;
+using Ocelot.DependencyInjection;
+using Ocelot.Middleware;
+using System.Threading.RateLimiting;
+
+var builder = WebApplication.CreateBuilder(args);
+builder.Configuration.AddJsonFile("ocelot.json");
+builder.Services.AddOcelot();
+
+builder.Services.AddRateLimiter(op =>
```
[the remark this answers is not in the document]
yjorayev: In config file, RateLimiterPolicyName must be set when using dotnet rate limiter. This policy must be configured at startup. Dotnet allows registering multiple policies. Each endpoint defines which policy to use. See [here](https://learn.microsoft.com/en-us/aspnet/core/performance/rate-limit?view=aspnetcore-8.0#fixed-window-limiter)
It is very similar to how to Authentication is setup with Ocelot.

src/Ocelot/Configuration/File/FileRateLimitRule.cs
```diff
@@ -65,11 +83,22 @@ public override string ToString()
             }
 
             var sb = new StringBuilder();
-            sb.Append(
+
+            sb.Append($"{nameof(RateLimitMiddlewareType)}:{RateLimitMiddlewareType},");
+
+            if (RateLimitMiddlewareType == RateLimitMiddlewareType.DotNet)
```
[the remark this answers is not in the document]
yjorayev: As [Guillaume Gnaegi](https://github.com/ggnaegi) pointed out, "AspNet is an old concept". However, if you insist I don't mind either way.

src/Ocelot/RateLimiting/Middleware/RateLimitingMiddleware.cs
yjorayev: The reason I had to resort for this hack is because all the classes .Net has for this feature are marked _internal_. See [RateLimitingMiddleware](https://github.com/dotnet/aspnetcore/blob/46ea84f2c3c53c51935210d794975dcfa024ee4e/src/Middleware/RateLimiting/src/RateLimitingMiddleware.cs) and [RateLimitingMetrics](https://github.com/dotnet/aspnetcore/blob/46ea84f2c3c53c51935210d794975dcfa024ee4e/src/Middleware/RateLimiting/src/RateLimitingMetrics.cs).

This was the only way I could think of to reuse what .Net already has. I would be glad explore ff you have a suggestion for a different approach.

The `httpContext.SetEndpoint(endpoint)` call is there so that we can add `EnableRateLimitingAttribute` with the policy name for the current http request.

test/Ocelot.AcceptanceTests/RateLimiting/RateLimitingTests.cs
raman-m: The naming is confusing due to the old Ocelot feature implementation. To clarify, I intend to rename the old classes (those without IBC) that should have the `Ocelot` prefix, and use `AspNet` as the new feature prefix to differentiate between Ocelot's own implementation and the ASP.NET native framework. This should be quite straightforward.

**P.S.** Gui, conducting a separate code review would be more efficient than debating over this minor styling issue, don't you agree?

raman-m:
>  but we are stuck with the .NET rate limiters specs, you can't modify them on the fly, as you could with the current limiters. 

@ggnaegi, what is your suggestion?
Could you please clarify the problem for the author?
Why not conduct a separate code review yourself?

test/Ocelot.AcceptanceTests/RateLimiting/RateLimitingTests.cs
raman-m: > AspNet is an old concept in my opinion

@ggnaegi In my view, it remains pertinent. It's an integral component of the ASP.NET Core framework, which now includes new functionalities for rate limiting. As a C# developer, it's essential to recognize the vastness of the .NET framework, encompassing countless classes. .NET encompasses numerous smaller frameworks and specific protocols. Indeed, ASP.NET Core is a subset of .NET, and it's also substantial in size. Thus, while the term 'DotNet' is accurate, it doesn't closely relate to the specific framework we utilize, which is ASP.NET Core.

To be honest, I'm puzzled by the shift in your mood over the past few months, especially when it comes to arguing over trivial matters. What's changed, Gui? What made you stop contributing?
Please, answer me in LinkedIn chat. Thanks!

raman-m: **Heads up, everyone!** A visit to the MS Learn Docs and a read through the [Fixed window limiter](https://learn.microsoft.com/en-us/aspnet/core/performance/rate-limit?view=aspnetcore-8.0#fixed-window-limiter) section reveals that:
```csharp
using Microsoft.AspNetCore.RateLimiting;
using System.Threading.RateLimiting;
```
Most classes fall under the `Microsoft.AspNetCore.RateLimiting` namespace. However, some are under the `System.Threading.RateLimiting` namespace. It appears the rate limiter is closely integrated with the `System.Threading` infrastructure, yet it seems that most classes belong to `Microsoft.AspNetCore.RateLimiting`.
In summary, the Rate Limiter feature is part of ASP.NET Core, rather than .NET as a whole.

src/Ocelot/RateLimiting/Middleware/RateLimitingMiddleware.cs
raman-m: I'm rephrasing my question...
**Is the setting added to the pipeline and made available for subsequent requests, or is it added only temporarily for the current request?**
----
> The reason I had to resort for this hack is because all the classes .Net has for this feature are marked internal. See [RateLimitingMiddleware](https://github.com/dotnet/aspnetcore/blob/46ea84f2c3c53c51935210d794975dcfa024ee4e/src/Middleware/RateLimiting/src/RateLimitingMiddleware.cs) and [RateLimitingMetrics](https://github.com/dotnet/aspnetcore/blob/46ea84f2c3c53c51935210d794975dcfa024ee4e/src/Middleware/RateLimiting/src/RateLimitingMetrics.cs).

OK, I see the problem! `internal` means the class can't be instantiated by constructor outside the MS lib assembly. `sealed` means the class can't be inhered by descendants. MS engineers don't wanna reshare internal implementations proposing to use only official extension methods to use the feature. This is old design dilemma for production projects when dev team don't want to process support enquiries from other developer whose projects consume the lib.
So, will you propose some else  solution if we will hack the problem of `internal sealed` qualifiers on the classes?

raman-m: > This was the only way I could think of to reuse what .Net already has. I would be glad explore if you have a suggestion for a different approach.

Currently, I haven't utilized the ASP.NET Rate Limiter. Please allow me some time to explore this feature...
As previously mentioned, anticipate several rounds of code review. My aim is to avert any issues by ensuring the feature is simple, flexible, and scalable. I'm open to coding shortcuts since our project, being an outdated ASP.NET product, already contains many. However, rest assured, we will discover a more efficient solution.
I'm confident that the injection, definition, and processing of rate limiting policies can be simplified.

raman-m: > The `httpContext.SetEndpoint(endpoint)` call is there so that we can add `EnableRateLimitingAttribute` with the policy name for the current http request.

It seems like a workaround! 😄
Okay, if I understand correctly, the issue is setting up the ASP.NET pipeline with the policy settings, and you're unsure how to do this efficiently. You're suggesting updating the pipeline and HTTP context services dynamically, correct?
Perhaps moving the setup to the application's startup phase would be better than a workaround within the middleware?
Since the policy setting is tied to DI services, any singleton/scoped DI service should suffice, or even moving the setup to the pipeline building stage, similar to [OcelotBuilder](https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/DependencyInjection/OcelotBuilder.cs) and [BuildOcelotPipeline](https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/Middleware/OcelotPipelineExtensions.cs), right❔

src/Ocelot/Configuration/File/FileRateLimitRule.cs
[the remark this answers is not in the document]
raman-m: > AspNet is an old concept

I have no idea why he said like that. 🤣 He loves dev jokes. 😉 
Now, let's examine this together: `RateLimitMiddlewareType.DotNet` appears to be ".NET middleware type for rate limiting". In the context of our project, where we employ ASP.NET middleware, it should be understood as "ASP.NET middleware type for rate limiting", applicable to both Ocelot and native implementations.
Are you familiar with other types of .NET middleware? Which ones are you referring to?

Generally, there is no problem: we distinguish the two implementations by type, yet there is a readability concern.
Please choose between: `AspNet` or `Native`
I believe this can be read without any confusion:
```C#
enum RateLimitMiddlewareType
{
    Ocelot,
    Native, // AspNet
}
```
But I am confused when defined as
```C#
enum RateLimitMiddlewareType
{
    Ocelot,
    DotNet,
}
```

samples/RateLimiter/Program.cs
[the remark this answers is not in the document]
raman-m:
```json
      "RateLimitOptions": {
        "EnableRateLimiting": true,
        "RateLimitMiddlewareType": "DotNet",
        "RateLimitPolicyName": "fixed"
      }
```
OK [I see now](https://github.com/ThreeMammals/Ocelot/pull/2188/files/617c85aafd1de808c91673d6f72c640dfb648e33#diff-cb4d86798213d830f3e239d6ff020de91ca1de6807dec9bcb77ca8a2e41789ffR28-R32)!

> It is very similar to how to Authentication is setup with Ocelot.

Yeap! I am familiar with the Auth policy concept.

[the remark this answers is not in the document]
raman-m: The concern is that the property names are overly long with unnecessary prefixes. My argument is that properties within `RateLimitOptions` should not include the `RateLimit` prefix, suffix, or be embedded within the name.

I suggest to rename the properties to more simplified names as shown below:
```diff
-        "RateLimitMiddlewareType": "DotNet",
-        "RateLimitPolicyName": "fixed"
+        "MiddlewareType": "DotNet",
+        "PolicyName": "fixed"
```
Alternatively, it may be even more concise to shorten them as follows:
```diff
-        "RateLimitMiddlewareType": "DotNet",
-        "RateLimitPolicyName": "fixed"
+        "Middleware": "DotNet",
+        "Policy": "fixed"
```
Which version do you prefer?

src/Ocelot/Configuration/File/FileRateLimitRule.cs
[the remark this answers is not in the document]
raman-m: I've come to realize that we might avoid using the proposed `enum` by deciding on empty and non-empty values for the policy name. If we define a policy name, we should configure the ASP.NET middleware automatically; otherwise, we retain the traditional Ocelot behavior. However, this needs to be clearly documented in the User's Docs.

Team consensus is essential here. Personally, I think introducing the enumeration is overengineering. We can safely rely on the decision logic based on the `PolicyName`.

src/Ocelot/Configuration/RateLimitOptions.cs
```diff
@@ -18,73 +19,79 @@ public RateLimitOptions(bool enableRateLimiting, string clientIdHeader, Func<Lis
...
+
         /// <summary>
         /// Disables X-Rate-Limit and Rety-After headers.
         /// </summary>
         /// <value>
         /// A boolean value for disabling X-Rate-Limit and Rety-After headers.
         /// </value>
         public bool DisableRateLimitHeaders { get; }
+
+        public RateLimitMiddlewareType RateLimitMiddlewareType { get; }
```
raman-m, line 93: I'm quite certain this is overengineering. We can simply determine the `PolicyName` value to be either empty or non-empty.
Let me to explain... If we had more than 3 middleware types, then defining an enum would be necessary. However, since we only have 2 types (Ocelot and AspNet), the enumeration is not required. So, we can decide on policy value (defined or not defined).

samples/RateLimiter/Program.cs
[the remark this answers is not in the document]
yjorayev: Sure. I prefer the second option.

src/Ocelot/Configuration/File/FileRateLimitRule.cs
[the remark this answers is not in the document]
yjorayev: I agree that enum is not necessary here with the note that this must be documented clearly.

src/Ocelot/Configuration/RateLimitMiddlewareType.cs
```diff
@@ -0,0 +1,7 @@
+namespace Ocelot.Configuration;
+
+public enum RateLimitMiddlewareType
+{
+    Ocelot,
+    DotNet,
```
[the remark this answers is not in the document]
yjorayev: Can you clarify whether you are saying xml dev docs must be added or you are against of having them anywhere in the project? I ask, because it is not consistent throughout the project.

src/Ocelot/DependencyInjection/Features.cs
```diff
@@ -16,10 +24,29 @@ public static class Features
     /// Read The Docs: <see href="https://ocelot.readthedocs.io/en/latest/features/ratelimiting.html">Rate Limiting</see>.
     /// </remarks>
     /// <param name="services">The services collection to add the feature to.</param>
+    /// <param name="configurationRoot">Root configuration object.</param>
     /// <returns>The same <see cref="IServiceCollection"/> object.</returns>
-    public static IServiceCollection AddRateLimiting(this IServiceCollection services) => services
-        .AddSingleton<IRateLimiting, RateLimiting.RateLimiting>()
-        .AddSingleton<IRateLimitStorage, MemoryCacheRateLimitStorage>();
+    public static IServiceCollection AddRateLimiting(this IServiceCollection services, IConfiguration configurationRoot)
```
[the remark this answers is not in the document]
yjorayev: I will add a new method instead of overriding existing one. I assumed this method is for internal purposes, but now I realize that it was a wrong assumption given that this method is public.

src/Ocelot/RateLimiting/Middleware/RateLimitingMiddleware.cs
yjorayev: > Is the setting added to the pipeline and made available for subsequent requests, or is it added only temporarily for the current request?

It is added temporarily. Since HttpContext is a scoped, nothing added/modified will persist into subsequent requests.

> If we consider bypassing the internal sealed restrictions, do you have an alternative solution in mind?

I don't have an alternative solution in mind at this moment.

> It seems like a workaround! 😄

It definitely is a workaround. I am trying to "trick" .Net RateLimiting middleware that it is currently executing and `Endpoint` so that the following line will not return null ([source](https://github.com/dotnet/aspnetcore/blob/46ea84f2c3c53c51935210d794975dcfa024ee4e/src/Middleware/RateLimiting/src/RateLimitingMiddleware.cs)):
`var endpoint = context.GetEndpoint();`
**and** this endpoint has `EnableRateLimiting` attribute.
`var enableRateLimitingAttribute = endpoint?.Metadata.GetMetadata<EnableRateLimitingAttribute>();`

> You're suggesting updating the pipeline and HTTP context services dynamically, correct?

Yes.

> Perhaps moving the setup to the application's startup phase would be better than a workaround within the middleware?
> Since the policy setting is tied to DI services, any singleton/scoped DI service should suffice, or even moving the setup to the pipeline building stage, similar to [OcelotBuilder](https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/DependencyInjection/OcelotBuilder.cs) and [BuildOcelotPipeline](https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/Middleware/OcelotPipelineExtensions.cs), right❔

I am not sure I understand this part. Are you suggesting to encapsulate following lines in a service, registering this new service in DI and invoking the service from the middleware?

 ```
var metadata = new EndpointMetadataCollection(new EnableRateLimitingAttribute(options.RateLimiterPolicyName));
var endpoint = new Endpoint(null, metadata, "tempEndpoint");
httpContext.SetEndpoint(endpoint);
```

File: src/Ocelot/Configuration/RateLimitMiddlewareType.cs (hunk 6 above)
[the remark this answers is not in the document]
Comment (raman-m) [reply src/Ocelot/Configuration/RateLimitMiddlewareType.cs:6]: Not an issue, because we've agreed to delete this `enum`. It is not required.

File: src/Ocelot/DependencyInjection/Features.cs (hunk 7 above)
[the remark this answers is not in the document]
Comment (raman-m) [reply src/Ocelot/DependencyInjection/Features.cs:29]: The decision hinges on the ultimate logic! Perhaps separating the internal extension methods is not essential.
I advise to revisit this once all the logic has been developed.

File: samples/RateLimiter/Program.cs (hunk 3 above)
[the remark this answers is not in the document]
Comment (yjorayev) [reply samples/RateLimiter/Program.cs:10]: renamed as discussed.

File: src/Ocelot/Configuration/RateLimitMiddlewareType.cs [hunk 8]
```diff
@@ -0,0 +1,7 @@
+namespace Ocelot.Configuration;
```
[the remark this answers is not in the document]
yjorayev: this enum is deleted

src/Ocelot/DependencyInjection/OcelotBuilder.cs
```diff
@@ -103,7 +103,7 @@ public OcelotBuilder(IServiceCollection services, IConfiguration configurationRo
             Services.TryAddSingleton<IDownstreamRouteProviderFactory, DownstreamRouteProviderFactory>();
             Services.TryAddSingleton<IHttpResponder, HttpContextResponder>();
             Services.TryAddSingleton<IErrorsToHttpStatusCodeMapper, ErrorsToHttpStatusCodeMapper>();
-            Services.AddRateLimiting(); // Feature: Rate Limiting
+            Services.AddRateLimiting(configurationRoot); // Feature: Rate Limiting
```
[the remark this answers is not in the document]
yjorayev: separated

src/Ocelot/Middleware/OcelotPipelineExtensions.cs
```diff
@@ -79,6 +79,14 @@ public static RequestDelegate BuildOcelotPipeline(this IApplicationBuilder app,
             // We check whether the request is ratelimit, and if there is no continue processing
             app.UseRateLimiting();
 
+            //use .Net rate limiter
+#if NET7_0_OR_GREATER
+            app.UseWhen(UseDotNetRateLimiter, rateLimitedApp =>
+            {
+                rateLimitedApp.UseRateLimiter();
+            });
+#endif
```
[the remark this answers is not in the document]
yjorayev: done

samples/RateLimiter/Ocelot.Samples.RateLimiter.http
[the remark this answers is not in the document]
yjorayev: Little "play" icon shows near each request definition that can be clicked to send http request. Below screenshot is from Rider, but VS has this feature too.

Screenshot

[Docs for .http](https://learn.microsoft.com/en-us/aspnet/core/test/http-files?view=aspnetcore-8.0)

ggnaegi:
> > but we are stuck with the .NET rate limiters specs, you can't modify them on the fly, as you could with the current limiters.
> 
> @ggnaegi, what is your suggestion? Could you please clarify the problem for the author? Why not conduct a separate code review yourself?

I'am wondering if we could replace all the current rate limiters in the request pipeline, like this (haven't checked), then it would make sense imho:

```c#
public class CustomRateLimiterMiddleware
{
    private readonly RequestDelegate _next;
    private readonly TokenBucketRateLimiter _rateLimiter;

    public CustomRateLimiterMiddleware(RequestDelegate next)
    {
        _next = next;
        
        // Create a rate limiter with a custom configuration
        _rateLimiter = new TokenBucketRateLimiter(
            new TokenBucketRateLimiterOptions
            {
                TokenLimit = 10,
                QueueProcessingOrder = QueueProcessingOrder.OldestFirst,
                QueueLimit = 5,
                ReplenishmentPeriod = TimeSpan.FromSeconds(1),
                TokensPerPeriod = 1,
                AutoReplenishment = true
            });
    }

    public async Task InvokeAsync(HttpContext context)
    {
        // Attempt to lease a token for each request
        var lease = await _rateLimiter.WaitAsync();

        if (lease.IsAcquired)
        {
            try
            {
                await _next(context);
            }
            finally
            {
                lease.Dispose();
            }
        }
        else
        {
            context.Response.StatusCode = StatusCodes.Status429TooManyRequests;
        }
    }
}
```

yjorayev:
@ggnaegi, @raman-m  I think we can make this work. Couple things:

1.  We will need to be able to get rate limiter policy by name. Simple dictionary might suffice. 
2.  If we go this route, we might find ourselves re-implementing(copying?)  `CombinedAcquire` and `CombinedWaitAsync` methods of [RateLimitingMiddleware](https://github.com/dotnet/aspnetcore/blob/46ea84f2c3c53c51935210d794975dcfa024ee4e/src/Middleware/RateLimiting/src/RateLimitingMiddleware.cs#L211). 

What do you think?

ggnaegi:
> @ggnaegi, @raman-m I think we can make this work. Couple things:
> 
> 1. We will need to be able to get rate limiter policy by name. Simple dictionary might suffice.
> 2. If we go this route, we might find ourselves re-implementing(copying?)  `CombinedAcquire` and `CombinedWaitAsync` methods of [RateLimitingMiddleware](https://github.com/dotnet/aspnetcore/blob/46ea84f2c3c53c51935210d794975dcfa024ee4e/src/Middleware/RateLimiting/src/RateLimitingMiddleware.cs#L211).
> 
> What do you think?

Yes, that's the main concern; we essentially need to reinvent the wheel for the middleware part. I've tried to "dynamize" the policies myself, but I believe it’s not possible. However, at least we can retain Microsoft's RateLimiter implementations and replace the current ones.

I’d go down that path, and if you look at the code, there isn’t much to modify.

@yjorayev @raman-m

raman-m:
@ggnaegi @yjorayev 
### Hello, guys! What's up? 😄

samples/RateLimiter/Ocelot.Samples.RateLimiter.http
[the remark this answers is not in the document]
raman-m: But this is Rider IDE! Have you used the same feature in Visual Studio?

raman-m:
@yjorayev commented [on Nov 14](https://github.com/ThreeMammals/Ocelot/pull/2188#issuecomment-2475320853)
>2.  If we go this route, we might find ourselves re-implementing(copying?)  `CombinedAcquire` and `CombinedWaitAsync` methods of [RateLimitingMiddleware](https://github.com/dotnet/aspnetcore/blob/46ea84f2c3c53c51935210d794975dcfa024ee4e/src/Middleware/RateLimiting/src/RateLimitingMiddleware.cs#L211). 

I don't recommend to copy & paste code from MS codebase, for a long term purposes because the internal implementation can change with .NET version increasing. But it can be acceptable in short term until we developed real handy design solution.

### Let's hack this private method?! :sunglasses: Why not to call it with a help of reflection technology? :speak_no_evil:

raman-m:
@ggnaegi commented [on Nov 14](https://github.com/ThreeMammals/Ocelot/pull/2188#issuecomment-2475603444)

> Yes, that's the main concern; **we essentially need to reinvent the wheel for the middleware part**.

Reflection to call internal parts of the MS middleware? 🤣 I don't think that reflection code is a wheel. 🤣 

> I've tried to "dynamize" the policies myself, but I believe it’s not possible.

Really? Have a repo to deep dive? 😉 

>However, at least we can retain Microsoft's RateLimiter implementations and replace the current ones.

"No, no, no," the ladies say. 😄 We will add the ugly duck MS feature as a sidecar module for the current Ocelot's feat, and our own Ocelot's feat should remain and evolve.

@ggnaegi **FYI**, we are going to implement our own rate limiting by Client's IPs. See issue #1050 by our Iranian friend, Mohsen. And, the current Rate Limiting scope contains the following issues:
- #1915: I guess this is a must-have feature.
- #1050: I guess the community will love this feature because the configuration will be much simpler than MS policies.
- #1587: I guess this awesome Ocelot feature will be a state-of-the-art feature, making our fans just happy.
- and many more.

ggnaegi:
> @ggnaegi commented [on Nov 14](https://github.com/ThreeMammals/Ocelot/pull/2188#issuecomment-2475603444)
> 
> > Yes, that's the main concern; **we essentially need to reinvent the wheel for the middleware part**.
> 
> Reflection to call internal parts of the MS middleware? 🤣 I don't think that reflection code is a wheel. 🤣
> 
> > I've tried to "dynamize" the policies myself, but I believe it’s not possible.
> 
> Really? Have a repo to deep dive? 😉
> 
> > However, at least we can retain Microsoft's RateLimiter implementations and replace the current ones.
> 
> "No, no, no," the ladies say. 😄 We will add the ugly duck MS feature as a sidecar module for the current Ocelot's feat, and our own Ocelot's feat should remain and evolve.
> 
> @ggnaegi **FYI**, we are going to implement our own rate limiting by Client's IPs. See issue #1050 by our Iranian friend, Mohsen. And, the current Rate Limiting scope contains the following issues:
> 
> * [Apply rate limiting globally  #1915](https://github.com/ThreeMammals/Ocelot/issues/1915): I guess this is a must-have feature.
> * [Rate limit with RemoteIpAddress #1050](https://github.com/ThreeMammals/Ocelot/issues/1050): I guess the community will love this feature because the configuration will be much simpler than MS policies.
> * [Claim-based rate limiting #1587](https://github.com/ThreeMammals/Ocelot/issues/1587): I guess this awesome Ocelot feature will be a state-of-the-art feature, making our fans just happy.
> * and many more.

@raman-m as you wish, I think it's not a good idea, since we already can't maintain everything, but do as you wish.

raman-m:
> @ggnaegi wrote:
> - "as you wish, I think it's not a good idea, since we already can't maintain everything, but do as you wish."
> - Yes but still, we can reuse the Rate Limiters from Microsoft and avoid ours... But then the configuration etc. we can use our configuration

Gui, please refrain from saying "we can't" if you're not willing to contribute. As mentioned, we'll preserve the old feature design and continue to develop it. We cannot simply remove the old design features because the community will react negatively, as some production projects depend on them. Moreover, it appears we can't adapt the current Client's Headers limitation based on the MS feature engine... I believe that's not the right approach. It would be more prudent to keep both designs (both feature modes) separate to prevent future bug reports. At the moment, I'm considering integrating both modes into a single middleware, but I need to review the final design with the author. Following that, we'll decide whether to have a single middleware or two separate ones.

Concerning the outstanding issues for previous feature design enhancements, Mohsen has opened numerous PRs. I cannot request more from him at this moment since the existing PRs need to be advanced to the 'Dev Complete' state and merged. Subsequently, I will approach Mohsen's team to take on some tickets. FYI, Mohsen is aware of the new ASP.NET feature middleware and is incorporating it into his projects, yet I am confident his team can still assist us. Maybe other contributors will help us.

ggnaegi:
> > @ggnaegi wrote:
> > 
> > * "as you wish, I think it's not a good idea, since we already can't maintain everything, but do as you wish."
> > * Yes but still, we can reuse the Rate Limiters from Microsoft and avoid ours... But then the configuration etc. we can use our configuration
> 
> Gui, please refrain from saying "we can't" if you're not willing to contribute. As mentioned, we'll preserve the old feature design and continue to develop it. We cannot simply remove the old design features because the community will react negatively, as some production projects depend on them. Moreover, it appears we can't adapt the current Client's Headers limitation based on the MS feature engine... I believe that's not the right approach. It would be more prudent to keep both designs (both feature modes) separate to prevent future bug reports. At the moment, I'm considering integrating both modes into a single middleware, but I need to review the final design with the author. Following that, we'll decide whether to have a single middleware or two separate ones.
> 
> Concerning the outstanding issues for previous feature design enhancements, Mohsen has opened numerous PRs. I cannot request more from him at this moment since the existing PRs need to be advanced to the 'Dev Complete' state and merged. Subsequently, I will approach Mohsen's team to take on some tickets. FYI, Mohsen is aware of the new ASP.NET feature middleware and is incorporating it into his projects, yet I am confident his team can still assist us. Maybe other contributors will help us.

@raman-m Mmmh, do as you wish... I'm not sure about the quality of the code that was delivered by our colleagues, and I don't want to have two types of rate limiters maintained in the solution. I'm willing to contribute, but we don't have enough ressources and time to ensure the stability. So my moto: Less is more... And finally, you ask us to review code and do things, but the inherent bashing is just not fair, at the end people are going to give up.

yjorayev:
> I don't recommend copying and pasting code from the MS codebase for long-term purposes because the internal implementation can change with .NET version updates. However, it can be acceptable in the short term until we develop a more practical design solution based on `public` methods.
> #### Let's hack this private method?! 😎 Why not to call it with a help of reflection technology? 🙊

I disagree about calling private methods, because private method signature can(probably will) change even with minor version upgrades. 

What are the advantages of calling the private method via reflection in contrast to my original proposal in this PR? I understand that my proposal is also sort of a hack, but it only interacts with public methods and classes which is less likely to change.

ggnaegi:
> > I don't recommend copying and pasting code from the MS codebase for long-term purposes because the internal implementation can change with .NET version updates. However, it can be acceptable in the short term until we develop a more practical design solution based on `public` methods.
> > #### Let's hack this private method?! 😎 Why not to call it with a help of reflection technology? 🙊
> 
> I disagree about calling private methods, because private method signature can(probably will) change even with minor version upgrades.
> 
> What are the advantages of calling the private method via reflection in contrast to my original proposal in this PR? I understand that my proposal is also sort of a hack, but it only interacts with public methods and classes which is less likely to change.

@yjorayev @raman-m  Mmmh, i haven't read this bit of trolling from Raman. My point was: Use the classes FixedWindowRateLimiter, SlidingWindowRateLimiter etc. and avoid our constructs that are less reliable.

raman-m:
@yjorayev Hi Yhlas!
Do you have the enthusiasm proceed with development? Could you begin by addressing the merge conflicts, please?

If you have no objections, I would like to invite you to review the following PR &rarr;
- #2294 
where we will address the issue of various types of limiters.

src/Ocelot/Configuration/Validator/RouteFluentValidator.cs
```diff
@@ -120,7 +128,7 @@ private static bool IsValidPeriod(FileRateLimitRule rateLimitOptions)
             var minutesRegEx = new Regex("^[0-9]+m");
```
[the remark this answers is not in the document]
raman-m: https://github.com/ThreeMammals/Ocelot/blob/d5e6d9aef3265495a23e380d2145a769a0d5a7c3/src/Ocelot/Configuration/Validator/RouteFluentValidator.cs#L99-L112

Not an issue anymore!

src/Ocelot/Configuration/Validator/RouteFluentValidator.cs
```diff
@@ -85,28 +88,33 @@ public RouteFluentValidator(IAuthenticationSchemeProvider authenticationSchemePr
...
-            if (string.IsNullOrEmpty(options.AuthenticationProviderKey)
+            if (string.IsNullOrEmpty(options.AuthenticationProviderKey)
                 && options.AuthenticationProviderKeys.Length == 0)
             {
                 return true;
             }
 
             var schemes = await _authenticationSchemeProvider.GetAllSchemesAsync();
-            var supportedSchemes = schemes.Select(scheme => scheme.Name).ToList();
+            var supportedSchemes = schemes.Select(scheme => scheme.Name).ToList();
```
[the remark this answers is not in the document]
raman-m: This logic was moved to the [FileAuthenticationOptionsValidator](https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/Configuration/Validator/FileAuthenticationOptionsValidator.cs) class in PR #2114 (commit in develop is https://github.com/ThreeMammals/Ocelot/commit/265f9b4d0fcd9bed5fca2ed2c920842bb460e13c)
After refactoring the `supportedSchemes` variable is `IEnumerable<T>` now &rarr;
https://github.com/ThreeMammals/Ocelot/blob/d5e6d9aef3265495a23e380d2145a769a0d5a7c3/src/Ocelot/Configuration/Validator/FileAuthenticationOptionsValidator.cs#L28

OK, not an issue!

samples/RateLimiter/Program.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    op.AddFixedWindowLimiter(policyName: "fixed", options =>
+    {
+        options.PermitLimit = 2;
+        options.Window = TimeSpan.FromSeconds(12);
+        options.QueueProcessingOrder = QueueProcessingOrder.OldestFirst;
+        options.QueueLimit = 0;
+    });
+});
```
raman-m, line 19: I disagree! In this case, we haven't proposed anything new.
I think we should provide the user with an option to choose between configuration approaches. We could consider having a separate Ocelot rate-limiting rule to manage app auto-configurations.

src/Ocelot/Configuration/File/FileRateLimitRule.cs
```diff
@@ -74,6 +75,14 @@ public FileRateLimitRule(FileRateLimitRule from)
     /// <value>A <see cref="string"/> object which value defaults to "Ocelot.RateLimiting", see the <see cref="RateLimitOptions.DefaultCounterPrefix"/> property.</value>
     public string KeyPrefix { get; set; }
 
+    /// <summary>
+    /// Rate limit policy name. It only takes effect if rate limit middleware type is set to DotNet.
+    /// </summary>
+    /// <value>
+    /// A string of rate limit policy name.
+    /// </value>
+    public string Policy { get; set; }
```
raman-m, line 84: You might want to consider removing this property and using `FileRateLimitByAspNetRule.Policy` instead.

src/Ocelot/Configuration/Validator/RouteFluentValidator.cs
```diff
@@ -56,7 +56,9 @@ public RouteFluentValidator(
                 .WithMessage("{PropertyName} {PropertyValue} contains scheme");
         });
 
-        When(route => route.RateLimitOptions != null && route.RateLimitOptions.EnableRateLimiting != false, () =>
+        When(route => route.RateLimitOptions != null
+                && route.RateLimitOptions.EnableRateLimiting != false
+                && string.IsNullOrWhiteSpace(route.RateLimitOptions.Policy), () =>
```
raman-m: We still need to validate the rule algorithm properties, even if the ASP.NET policy is in place.

src/Ocelot/Configuration/RateLimitOptions.cs
```diff
@@ -63,6 +65,7 @@ public RateLimitOptions(FileRateLimitByHeaderRule fromRule)
             fromRule.Period.IfEmpty(RateLimitRule.DefaultPeriod),
             fromRule.PeriodTimespan.HasValue ? $"{fromRule.PeriodTimespan.Value}s" : fromRule.Wait,
             fromRule.Limit ?? RateLimitRule.ZeroLimit);
+        Policy = fromRule.Policy;
```
raman-m: An option of a flawed rule 😄 Consider removing!

src/Ocelot/Configuration/RateLimitOptions.cs
```diff
@@ -77,6 +80,7 @@ public RateLimitOptions(RateLimitOptions fromOptions)
         QuotaMessage = fromOptions.QuotaMessage.IfEmpty(DefaultQuotaMessage);
         KeyPrefix = fromOptions.KeyPrefix.IfEmpty(DefaultCounterPrefix);
         Rule = fromOptions.Rule ?? RateLimitRule.Empty;
+        Policy = fromOptions.Policy;
```
raman-m: It's fine because of the copy constructor ✔️

src/Ocelot/DependencyInjection/Features.cs
```diff
@@ -7,6 +10,11 @@
 using Ocelot.RateLimiting;
 using FluentValidation;
 
+#if NET7_0_OR_GREATER
```
raman-m: The TMFs for our current project are `net8.0` and `net9.0`. Eliminate the preprocessor directive!

src/Ocelot/DependencyInjection/Features.cs
```diff
@@ -30,10 +38,39 @@ public static IServiceCollection AddConfigurationValidators(this IServiceCollect
...
+    /// <summary>
+    /// Ocelot feature: <see href="">AspNet Rate Limiting</see>.
+    /// </summary>
+    /// <remarks>
+    /// Read The Docs: <see href="">Rate Limiting</see>.
+    /// </remarks>
+    /// <param name="services">The services collection to add the feature to.</param>
+    /// <param name="configurationRoot">Root configuration object.</param>
+    /// <returns>The same <see cref="IServiceCollection"/> object.</returns>
+    public static IServiceCollection AddAspNetRateLimiting(this IServiceCollection services, IConfiguration configurationRoot)
```
raman-m, line 81: I'm not sure, but I have a strong feeling that this adder should be part of the method on line 43. I'm open to discussions.
The goal is that the end user shouldn't need to write any code for adding services, as the current `AddRateLimiting` is an internal Ocelot feature already included in the DI container by default.

src/Ocelot/DependencyInjection/Features.cs
```diff
@@ -30,10 +38,39 @@ public static IServiceCollection AddConfigurationValidators(this IServiceCollect
...
+    {
+        var globalRateLimitOptions = configurationRoot.Get<FileConfiguration>()?.GlobalConfiguration?.RateLimitOptions;
+        var rejectStatusCode = globalRateLimitOptions?.HttpStatusCode ?? StatusCodes.Status429TooManyRequests;
+        var rejectedMessage = globalRateLimitOptions?.QuotaExceededMessage ?? "API calls quota exceeded!";
```
raman-m, line 85: Merging in the wrong place can be risky.
We've assigned the `IRateLimitOptionsCreator` service specifically for merging options. Please refer to the [RateLimitOptionsCreator](https://github.com/ThreeMammals/Ocelot/blob/develop/src/Ocelot/Configuration/Creator/RateLimitOptionsCreator.cs) class for details.

src/Ocelot/DependencyInjection/Features.cs
```diff
@@ -30,10 +38,39 @@ public static IServiceCollection AddConfigurationValidators(this IServiceCollect
...
+        services.AddRateLimiter(options =>
+        {
+            options.OnRejected = async (rejectedContext, token) =>
+            {
+                rejectedContext.HttpContext.Response.StatusCode = rejectStatusCode;
+                await rejectedContext.HttpContext.Response.WriteAsync(rejectedMessage, token);
```
raman-m, line 92: It's amusing that we implement the same approach in Ocelot's middleware. Perhaps adding a virtual method in the middleware to connect to the handler could work? 😉

src/Ocelot/DependencyInjection/OcelotBuilder.cs
```diff
@@ -111,7 +110,10 @@ public OcelotBuilder(IServiceCollection services, IConfiguration configurationRo
         Services.AddOcelotCache();
         Services.AddOcelotMetadata();
         Services.AddOcelotMessageInvokerPool();
-
+        Services.AddRateLimiting(); // Feature: Rate Limiting
+#if NET7_0_OR_GREATER
+        Services.AddAspNetRateLimiting(configurationRoot); // Feature: AspNet Rate Limiting
```
raman-m: I strongly believe we shouldn't introduce a new feature, whether public or internal. Since the method is already auto-enabled internally, it seems we can leave it as-is. However, we definitely need to review this design.

src/Ocelot/RateLimiting/RateLimitingMiddlewareExtensions.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http;
+using Ocelot.Middleware;
+
+namespace Ocelot.RateLimiting.Middleware;
+
+public static class RateLimitingMiddlewareExtensions
+{
+    public static IApplicationBuilder UseRateLimiting(this IApplicationBuilder builder)
```
raman-m, line 9: Great! It looks like the number of references is zero.

Image

So, how are we supposed to use this extension method? Just mentioning in the documentation that users should call it?

test/Ocelot.AcceptanceTests/RateLimiting/AspNetRateLimitingTests.cs
```diff
@@ -0,0 +1,76 @@
...
+public class AspNetRateLimitingTests: RateLimitingSteps
+{
+    private const string FixedWindowPolicyName = "RateLimitPolicy";
+    private int _rateLimitLimit;
+    private int _rateLimitWindow;
+    private string _quotaExceededMessage;
+    
+    [Fact]
+    [Trait("Feat", "2138")]
+    public async Task Should_RateLimit()
```
raman-m, line 21: Currently, this test fails ❗

raman-m (changes requested):
Hi Yhlas!
Do you recall what you were doing a year ago? **Do you plan to continue contributing further?**

With the merge conflicts resolved and #2294 successfully integrated, it seems like the perfect time to move forward with development. Unfortunately, the [acceptance test](https://github.com/ThreeMammals/Ocelot/pull/2188/files#diff-ad7ded98f5f70cdb779cdf79edfaacc6644813b24d47bcd51c963ab6781a28f7) is currently failing.

src/Ocelot/DependencyInjection/Features.cs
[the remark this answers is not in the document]
raman-m: Apologies, I was mistaken! There is no IBC since `AddRateLimiting` is an internal feature of `OcelotBuilder` that has already been included. Feel free to refactor this method as needed.
Not an issue!

samples/RateLimiter/Program.cs
ggnaegi: @raman-m it's basically what i mentioned:
_Yes, that's the main concern; we essentially need to reinvent the wheel for the middleware part. I've tried to "dynamize" the policies myself, but I believe it’s not possible._
The policies are setup during startup, you can't modify them after that, and you can't mix them (as of 2024, maybe now it's possible).

raman-m: We can't mix policies, but we can define as many as needed, right?
From your words, did you mean that the "fixed" policy is global in this code?

ggnaegi: Yes, it's global and they can't be modified.

test/Ocelot.AcceptanceTests/RateLimiting/AspNetRateLimitingTests.cs
raman-m: ## PR [build 854](https://github.com/ThreeMammals/Ocelot/actions/runs/33899963390?pr=2188) failed ❗

raman-m:
@yjorayev Are you available to continue development?

PR #2213, opened 2024-11-22
#2212 Incorrect calculation of placeholders, only the last placeholder should be allowed to match several segments

### LINKED ISSUE #2212: Incorrect calculation of placeholders
## Expected Behavior
Path: /dati-registri/v1.0/operatore/R80QQ5J9600/valida
Template: /dati-registri/{version}/{everything}

Expected:
version = v1.0
everything = operatore/R80QQ5J9600/valida

## Actual Behavior 

version = v1.0/operatore/R80QQ5J9600
everything = valida

in the previous version, the behavior was as expected

Set the configuration in the gateway with an ocelot.settings.json file with an UpstreamPathTemplate as indicated in the first point and create an endpoint in the API with the path indicated in the first point.

## Specifications

  - Version: 23.4.0.0
  - Platform: .net 9 win

ggnaegi:
## Fixes #2212 
- #2212 
 
## Proposed Changes
- Ensuring that only the last placeholder is allowed to match multiple segments (separated by slashes).
- Accomplishing this by counting the placeholders and using an index during the replacement process.
- If the current index is not equal to the placeholder count, the placeholder cannot match forward slashes; if the index equals the placeholder count, it can match slashes.

### DIFF
--- a/src/Ocelot/DownstreamRouteFinder/UrlMatcher/UrlPathPlaceholderNameAndValueFinder.cs
+++ b/src/Ocelot/DownstreamRouteFinder/UrlMatcher/UrlPathPlaceholderNameAndValueFinder.cs
@@ -67,10 +67,12 @@ public partial class UrlPathPlaceholderNameAndValueFinder : IPlaceholderNameAndV
     private static List<Group> FindGroups(string path, string query, string template)
     {
         template = EscapeExceptBraces(template);
-        var regexPattern = $"^{RegexPlaceholders().Replace(template, match => $"(?<{match.Groups[1].Value}>[^&]*)")}";
+        var regexPattern = GenerateRegexPattern(template);
         var testedPath = ShouldSkipQuery(query, template) ? path : $"{path}{query}";
+        
         var match = Regex.Match(testedPath, regexPattern);
         var foundGroups = match.Groups.Cast<Group>().Skip(1).ToList();
+        
         if (foundGroups.Count > 0 || !IsCatchAllPath(template))
         {
             return foundGroups;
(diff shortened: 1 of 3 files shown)


ggnaegi:
@raman-m ready for review

ggnaegi:
@raman-m could you please review this, so we can close it?

src/Ocelot/DownstreamRouteFinder/UrlMatcher/UrlPathPlaceholderNameAndValueFinder.cs
```diff
@@ -80,6 +82,33 @@ private static List<Group> FindGroups(string path, string query, string template
...
+        
+        int index = 0;
+        
+        // we know that the replace process will be started from the beginning of the url
+        // so we can use a simple counter to determine the last placeholder
+        var regexPattern = $@"^{RegexPlaceholders().Replace(escapedTemplate, match =>
+        {
+            var groupName = match.Groups[1].Value;
+            index++;
+            return index == placeHoldersCount ? $"(?<{groupName}>[^&]*)" : $"(?<{groupName}>[^/|&]*)";
```
raman-m, line 107: So, this `[^/|&]` negative matching group filters out all slashes `/` after closing brace `}`, before catch all placeholder, right?
And, the last RX-match, for the last Catch-All, just accepts uncounted number of slashes before query string, right?
Query string processing is quite unclear... What's the final generated pattern?

test/Ocelot.AcceptanceTests/Routing/RoutingTests.cs
```diff
@@ -501,6 +501,30 @@ public void ShouldNotMatchComplexQueriesCaseSensitive(string downstream, string
             .BDDfy();
     }
 
+    [Theory]
+    [Trait("Bug", "2212")]
+    [InlineData("/data-registers/{version}/it/{everything}", "/dati-registri/{version}/{everything}", "/dati-registri/v1.0/operatore/R80QQ5J9600/valida", "/data-registers/v1.0/it/operatore/R80QQ5J9600/valida")]
```
raman-m, line 506: Is this test data from ##2212, right?

raman-m (commented):
I will add one small commit with style-fixes, while you answering my questions...

src/Ocelot/DownstreamRouteFinder/UrlMatcher/UrlPathPlaceholderNameAndValueFinder.cs
```diff
@@ -67,10 +67,12 @@ public Response<List<PlaceholderNameAndValue>> Find(string path, string query, s
     private static List<Group> FindGroups(string path, string query, string template)
     {
         template = EscapeExceptBraces(template);
-        var regexPattern = $"^{RegexPlaceholders().Replace(template, match => $"(?<{match.Groups[1].Value}>[^&]*)")}";
+        var regexPattern = GenerateRegexPattern(template);
```
raman-m, line 70: Seems this is dynamic `Regex`, so the pattern is dynamically generated, but the previous version aka `RegexPlaceholders()` returns static object.
Well... is it possible to convert somehow to static object to improve the performance?

src/Ocelot/DownstreamRouteFinder/UrlMatcher/UrlPathPlaceholderNameAndValueFinder.cs
```diff
@@ -80,6 +82,29 @@ private static List<Group> FindGroups(string path, string query, string template
...
+        // We know that the replace process will be started from the beginning of the url,
+        // so we can use a simple counter to determine the last placeholder
+        string MatchEvaluator(Match match)
+        {
+            var groupName = match.Groups[1].Value;
+            index++;
+            return index == placeHoldersCount ? $"(?<{groupName}>[^&]*)" : $"(?<{groupName}>[^/|&]*)";
+        }
+
+        return $@"^{RegexPlaceholders().Replace(escapedTemplate, MatchEvaluator)}";
```
raman-m, line 106: Is it possible to move this replacement to the static object? It appears not, due to the varying `escapedTemplate` that represents the current path template. This is unfortunate 😢 as it seems there are no options available.

raman-m (approved):
### Ready for Delivery ✅ 
I am agreeable to merging; however, we will undoubtedly experience a loss in performance, a little bit... Nevertheless, intelligent logic must remain dynamic.

raman-m:
I will release a patch after your pressing magic buttons here, and I needn't your approval 😝

ggnaegi:
> ### Ready for Delivery ✅ 
> I am agreeable to merging; however, we will undoubtedly experience a loss in performance, a little bit... Nevertheless, intelligent logic must remain dynamic.

Yes, I think a few benchmarks are needed...

PR #2254, opened 2025-01-22
Setting up CI/CD with GitHub Actions using a standard .NET workflow

raman-m:
This PR introduces a new CI/CD setup based on the [Building and Testing .NET](https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-net) workflow.

## Proposed Changes
  - Removed CircleCI configuration
  - Added GitHub Actions configuration
  - Updated Visual Studio solution files

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -1,0 +1,28 @@
+# This workflow will build a .NET project
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-net
+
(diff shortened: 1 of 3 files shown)


TomPallister (commented):
Hi Raman, I would suggest continuing to use the ocelot container and cake build script because it also does the nuget publish and code coverage stuff. But the main advantage is you can build it locally in the same was as the CI server and know that there are no differences between your local environment and CI environment.

PR #2264, opened 2025-03-08
Removed ambiguity from ISecurityPolicy.

Nor-Wei:
# Streamlined ISecurityPolicy interface

## Proposed Changes

  - Removes synchronous function from ISecurityPolicy for clarity.
  - Adjusts implementation in IPSecurityPolicy class accordingly.
  - Updates affected unit tests.

### DIFF
--- a/src/Ocelot/Security/IPSecurity/IPSecurityPolicy.cs
+++ b/src/Ocelot/Security/IPSecurity/IPSecurityPolicy.cs
@@ -7,13 +7,13 @@ namespace Ocelot.Security.IPSecurity;
 
 public class IPSecurityPolicy : ISecurityPolicy
 {
-    public Response Security(DownstreamRoute downstreamRoute, HttpContext context)
+    public ValueTask<Response> SecurityAsync(DownstreamRoute downstreamRoute, HttpContext context)
     {
         var clientIp = context.Connection.RemoteIpAddress;
         var options = downstreamRoute.SecurityOptions;
         if (options == null || clientIp == null)
         {
-            return new OkResponse();
+            return ValueTask.FromResult(OkResponse);
         }
 
         if (options.IPBlockedList?.Count > 0)
@@ -21,7 +21,7 @@ public class IPSecurityPolicy : ISecurityPolicy
             if (options.IPBlockedList.Contains(clientIp.ToString()))
             {
                 var error = new UnauthenticatedError($"This request rejects access to {clientIp} IP");
-                return new ErrorResponse(error);
+                return ValueTask.FromResult<Response>(new ErrorResponse(error));
             }
         }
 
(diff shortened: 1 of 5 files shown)


raman-m:
Dear Norman,
What do you propose here?
I don't see any improvements or enhancements.

> Removed **ambiguity** from ISecurityPolicy.

Could you explain your use case? Have you encountered compilation errors?

Please note that having synchronous and asynchronous versions of an interface method is good practice for maintaining legacy software. Ultimately, the developer decides which version of the method to call on the context.

Nor-Wei:
Hi,

I need to implement the asynchronous function for ISecurityPolicy due to asynchronous calls in my implemetation.
I faced following problem:

What to do with the synchronous version of the two functions?
In my case there is no synchronous implementation that I could provide. Also, getting an asynchronous function executed in an synchronous context seems to be more hassle and not really a good practise.
Also, looking at the source, it seems the asynchronous version is never called anyway; the default implementation just wraps the synchronous version in a Task.Run - call.

Just my opinion:
If you want to support synchronous functions for it I would recommend to split the interfaces (e.g. like IDisposal and IAsyncDisposal).
If supporting synchronous calls for legacy reason I would still opt for only having an asynchronous version in ISecurityPolicy because it is way easier to call synchronous stuff without any "awaiting" in an asynchronous version than vice versa.

Hope that makes,
Norman.

src/Ocelot/Security/ISecurityPolicy.cs
```diff
@@ -6,6 +6,5 @@ namespace Ocelot.Security;
 
 public interface ISecurityPolicy
 {
-    Response Security(DownstreamRoute downstreamRoute, HttpContext context);
-    Task<Response> SecurityAsync(DownstreamRoute downstreamRoute, HttpContext context);
+    ValueTask<Response> SecurityAsync(DownstreamRoute downstreamRoute, HttpContext context);
```
raman-m, line 9: Potential IBC changes because of removal.
Please keep both versions.
And call async version from sync one in implementations.

src/Ocelot/Security/IPSecurity/IPSecurityPolicy.cs
```diff
@@ -30,13 +30,12 @@ public Response Security(DownstreamRoute downstreamRoute, HttpContext context)
...
+                return ValueTask.FromResult<Response>(new ErrorResponse(error));
             }
         }
 
-        return new OkResponse();
+        return ValueTask.FromResult(OkResponse);
     }
 
-    public Task<Response> SecurityAsync(DownstreamRoute downstreamRoute, HttpContext context)
-        => Task.Run(() => Security(downstreamRoute, context));
```
raman-m, line 41: That is not good. I know. But don't remove!
And call async version from sync one aka `SecurityAsync().GetAwaiter().GetResult()`

raman-m (changes requested):
Please work more!
We should not have any interface breaking changes.

Nor-Wei:
Hi,

Thanks for your response. I just decided to call my asynchronous implementation from the synchronous function for now.

PR #2257, opened 2025-01-24
#2255 #2256 Overload `AddKubernetes` method with `KubeClientOptions` param for the discovery provider

raman-m:
## Closes #2255 #2256 
- #2255 
- #2256 

## Proposed Changes
  - Added new overloaded `AddKubernetes(this IOcelotBuilder, Action<KubeClientOptions>, ...)` method
  - Updated Kubernetes docs
  - Updated Kubernetes sample

PR #2266, opened 2025-03-15
Upgrade to KubeClient v3 and log failed Kubernetes API requests

tintoy:
KubeClient v3 should simplify dependency-management, from Ocelot's perspective:

- KubeClient no longer has an external dependency (HTTPlease) for its HttpClient support. The internalised version of this support is now located in the `KubeClient.Http` namespace.
- KubeClient v3 is multi-targeted for `net9.0`, as well as a couple of older versions that some consumers are still using (support for these older versions will be tailed off over the next couple of minor releases)

Related discussion at:

tintoy/dotnet-kube-client#166
tintoy/dotnet-kube-client#167

CC: @raman-m (as previously discussed; sorry for the extended delay!)

## Proposed Changes

  - Update `KubeClient` packages to `v3.0.0`
  - Add support for logging detailed error responses from the Kubernetes API, in case of failure (only enabled at the Debug log level, or higher)

### DIFF
--- a/src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
+++ b/src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
@@ -1,17 +1,20 @@
-﻿using HTTPlease;
+﻿using KubeClient.Http;
 using KubeClient.Models;
 using KubeClient.ResourceClients;
+using Microsoft.Extensions.Logging;
 using Ocelot.Provider.Kubernetes.Interfaces;
 
 namespace Ocelot.Provider.Kubernetes;
 
 public class EndPointClientV1 : KubeResourceClient, IEndPointClient
 {
-    private readonly HttpRequest _collection;
+    private static readonly HttpRequest Collection = KubeRequest.Create("api/v1/namespaces/{Namespace}/endpoints/{ServiceName}");
+
+    private readonly ILogger _logger;
 
     public EndPointClientV1(IKubeApiClient client) : base(client)
     {
-        _collection = KubeRequest.Create("api/v1/namespaces/{Namespace}/endpoints/{ServiceName}");
+        _logger = client.LoggerFactory.CreateLogger<EndPointClientV1>();
     }
 
     public async Task<EndpointsV1> GetAsync(string serviceName, string kubeNamespace = null, CancellationToken cancellationToken = default)
@@ -21,7 +24,7 @@ public class EndPointClientV1 : KubeResourceClient, IEndPointClient
             throw new ArgumentNullException(nameof(serviceName));
         }
 
-        var request = _collection
+        var request = Collection
             .WithTemplateParameters(new
             {
                 Namespace = kubeNamespace ?? KubeClient.DefaultNamespace,
@@ -30,8 +33,28 @@ public class EndPointClientV1 : KubeResourceClient, IEndPointClient
 
         var response = await Http.GetAsync(request, cancellationToken);
 
-        return response.IsSuccessStatusCode
-            ? await response.ReadContentAsAsync<EndpointsV1>()
-            : null;
+        if (!response.IsSuccessStatusCode)
+        {
+            StatusV1 errorResponse = await response.ReadContentAsAsync<StatusV1>();
+
+            if (_logger.IsEnabled(LogLevel.Debug))
+            {
+                (string resourceKind, string resourceApiVersion) = KubeObjectV1.GetKubeKind<EndpointsV1>();
+                _logger.LogDebug("Failed to retrieve {ResourceApiVersion}/{ResourceKind} {ResourceName} in namespace {ResourceNamespace} ({HttpStatusCode}/{Status}/{StatusReason}): {StatusMessage}",
+                    resourceApiVersion,
+                    resourceKind,
+                    serviceName,
+                    kubeNamespace,
+                    response.StatusCode,
+                    errorResponse.Status,
+                    errorResponse.Reason,
+                    errorResponse.Message
+                );
+            }
+
+            return null;
+        }
+
+        return await response.ReadContentAsAsync<EndpointsV1>();
     }
 }
--- a/src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
+++ b/src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
@@ -13,7 +13,7 @@ public class EndPointClientV1 : KubeResourceClient, IEndPointClient
     {
     }
 
-    public async Task<EndpointsV1> GetAsync(string serviceName, string kubeNamespace = null, CancellationToken cancellationToken = default)
+    public Task<EndpointsV1> GetAsync(string serviceName, string kubeNamespace = null, CancellationToken cancellationToken = default)
     {
         if (string.IsNullOrEmpty(serviceName))
         {
@@ -27,7 +27,7 @@ public class EndPointClientV1 : KubeResourceClient, IEndPointClient
                 ServiceName = serviceName,
             });
 
-        return await Http.GetAsync(request, cancellationToken)
+        return Http.GetAsync(request, cancellationToken)
             .ReadContentAsObjectV1Async<EndpointsV1>(operationDescription: $"get {nameof(EndpointsV1)}");
     }
 }
--- a/src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
+++ b/src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
@@ -35,10 +35,10 @@ public class EndPointClientV1 : KubeResourceClient, IEndPointClient
 
         if (!response.IsSuccessStatusCode)
         {
-            StatusV1 errorResponse = await response.ReadContentAsAsync<StatusV1>();
-
             if (_logger.IsEnabled(LogLevel.Debug))
             {
+                StatusV1 errorResponse = await response.ReadContentAsAsync<StatusV1>();
+
                 (string resourceKind, string resourceApiVersion) = KubeObjectV1.GetKubeKind<EndpointsV1>();
                 _logger.LogDebug("Failed to retrieve {ResourceApiVersion}/{ResourceKind} {ResourceName} in namespace {ResourceNamespace} ({HttpStatusCode}/{Status}/{StatusReason}): {StatusMessage}",
                     resourceApiVersion,
--- a/src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
+++ b/src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
@@ -1,7 +1,6 @@
 ﻿using KubeClient.Http;
 using KubeClient.Models;
 using KubeClient.ResourceClients;
-using Microsoft.Extensions.Logging;
 using Ocelot.Provider.Kubernetes.Interfaces;
 
 namespace Ocelot.Provider.Kubernetes;
@@ -10,11 +9,8 @@ public class EndPointClientV1 : KubeResourceClient, IEndPointClient
 {
     private static readonly HttpRequest Collection = KubeRequest.Create("api/v1/namespaces/{Namespace}/endpoints/{ServiceName}");
 
-    private readonly ILogger _logger;
-
     public EndPointClientV1(IKubeApiClient client) : base(client)
     {
-        _logger = client.LoggerFactory.CreateLogger<EndPointClientV1>();
     }
 
     public async Task<EndpointsV1> GetAsync(string serviceName, string kubeNamespace = null, CancellationToken cancellationToken = default)
@@ -31,30 +27,7 @@ public class EndPointClientV1 : KubeResourceClient, IEndPointClient
                 ServiceName = serviceName,
             });
 
-        var response = await Http.GetAsync(request, cancellationToken);
-
-        if (!response.IsSuccessStatusCode)
-        {
-            if (_logger.IsEnabled(LogLevel.Debug))
-            {
-                StatusV1 errorResponse = await response.ReadContentAsAsync<StatusV1>();
-
-                (string resourceKind, string resourceApiVersion) = KubeObjectV1.GetKubeKind<EndpointsV1>();
-                _logger.LogDebug("Failed to retrieve {ResourceApiVersion}/{ResourceKind} {ResourceName} in namespace {ResourceNamespace} ({HttpStatusCode}/{Status}/{StatusReason}): {StatusMessage}",
-                    resourceApiVersion,
-                    resourceKind,
-                    serviceName,
-                    kubeNamespace,
-                    response.StatusCode,
-                    errorResponse.Status,
-                    errorResponse.Reason,
-                    errorResponse.Message
-                );
-            }
-
-            return null;
-        }
-
-        return await response.ReadContentAsAsync<EndpointsV1>();
+        return await Http.GetAsync(request, cancellationToken)
+            .ReadContentAsObjectV1Async<EndpointsV1>(operationDescription: $"get {nameof(EndpointsV1)}");
     }
 }
--- a/src/Ocelot.Provider.Kubernetes/Kube.cs
+++ b/src/Ocelot.Provider.Kubernetes/Kube.cs
@@ -70,16 +70,16 @@ public class Kube : IServiceDiscoveryProvider
                     httpStatusCode = httpRequestError.StatusCode.ToString();
                 }
 
-                return $"Failed to retrieve {EndPointsKubeKind.ResourceApiVersion}/{EndPointsKubeKind.ResourceKind} {serviceName} in namespace {kubeNamespace} ({httpStatusCode}/{errorResponse.Status}/{errorResponse.Reason}): {errorResponse.Message}";
+                return $"Failed to retrieve {EndPointsKubeKind.ResourceApiVersion}/{EndPointsKubeKind.ResourceKind} '{serviceName}' in namespace '{kubeNamespace}' (HTTP.{httpStatusCode}/{errorResponse.Status}/{errorResponse.Reason}): {errorResponse.Message}";
             }, kubeApiError);
         }
         catch (HttpRequestException unexpectedRequestError)
         {
-            _logger.LogError(() => $"Failed to retrieve {EndPointsKubeKind.ResourceApiVersion}/{EndPointsKubeKind.ResourceKind} {serviceName} in namespace {kubeNamespace} ({unexpectedRequestError.HttpRequestError}/{unexpectedRequestError.StatusCode}).", unexpectedRequestError);
+            _logger.LogError(() => $"Failed to retrieve {EndPointsKubeKind.ResourceApiVersion}/{EndPointsKubeKind.ResourceKind} '{serviceName}' in namespace '{kubeNamespace}' ({unexpectedRequestError.HttpRequestError}/HTTP.{unexpectedRequestError.StatusCode}).", unexpectedRequestError);
         }
         catch (Exception unexpectedError)
         {
-            _logger.LogError(() => $"Failed to retrieve {EndPointsKubeKind.ResourceApiVersion}/{EndPointsKubeKind.ResourceKind} {serviceName} in namespace {kubeNamespace} (an unexpected error occurred).", unexpectedError);
+            _logger.LogError(() => $"Failed to retrieve {EndPointsKubeKind.ResourceApiVersion}/{EndPointsKubeKind.ResourceKind} '{serviceName}' in namespace '{kubeNamespace}' (an unexpected error occurred).", unexpectedError);
         }
 
         return null;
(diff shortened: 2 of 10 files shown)


src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
```diff
@@ -30,8 +33,28 @@ public async Task<EndpointsV1> GetAsync(string serviceName, string kubeNamespace
...
-            ? await response.ReadContentAsAsync<EndpointsV1>()
-            : null;
+        if (!response.IsSuccessStatusCode)
+        {
+            if (_logger.IsEnabled(LogLevel.Debug))
+            {
+                StatusV1 errorResponse = await response.ReadContentAsAsync<StatusV1>();
+
+                (string resourceKind, string resourceApiVersion) = KubeObjectV1.GetKubeKind<EndpointsV1>();
+                _logger.LogDebug("Failed to retrieve {ResourceApiVersion}/{ResourceKind} {ResourceName} in namespace {ResourceNamespace} ({HttpStatusCode}/{Status}/{StatusReason}): {StatusMessage}",
```
raman-m, line 43: Since you log error responses, the more appropriate level is `LoggingLevel.Error`.
```suggestion
                _logger.LogError("Failed to retrieve {ResourceApiVersion}/{ResourceKind} {ResourceName} in namespace {ResourceNamespace} ({HttpStatusCode}/{Status}/{StatusReason}): {StatusMessage}",
```
Examples: [LogError](https://github.com/search?q=repo%3AThreeMammals%2FOcelot%20LogError&type=code)
We can also discuss the use of `LogWarning` due to `IsSuccessStatusCode` being `false`. However, this encompasses a wide range of status codes aka 4xx and 5xx.

src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
```diff
@@ -30,8 +33,28 @@ public async Task<EndpointsV1> GetAsync(string serviceName, string kubeNamespace
 
         var response = await Http.GetAsync(request, cancellationToken);
 
-        return response.IsSuccessStatusCode
-            ? await response.ReadContentAsAsync<EndpointsV1>()
-            : null;
+        if (!response.IsSuccessStatusCode)
+        {
+            if (_logger.IsEnabled(LogLevel.Debug))
```
raman-m, line 38: I am not certain, but if we use the native Ocelot logger, then all `Log*` methods have internal check to determine whether logging is enabled. Therefore, this `if` clause may be redundant.

src/Ocelot.Provider.Kubernetes/Ocelot.Provider.Kubernetes.csproj
```diff
@@ -25,8 +25,8 @@
     <NoWarn>1591</NoWarn>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="KubeClient" Version="2.5.12" />
-    <PackageReference Include="KubeClient.Extensions.DependencyInjection" Version="2.5.12" />
+    <PackageReference Include="KubeClient" Version="3.0.0" />
```
raman-m, line 28: Actually, it should be [3.0.1](https://www.nuget.org/packages/KubeClient/3.0.1) 😄

raman-m (changes requested):
Adam, overall, it is satisfactory, but the logging could be improved. Another concern is the absence of tests, despite the fact that the current tests are passing.

src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
tintoy: Yeah, ILogger does the same, but this check was more to decide if we want to check whether to try deserialising the error response;  it’s a pattern I tend to use when we perform some non-zero-cost operation to collect info purely for logging purposes. Some rationale from the runtime folks here:

https://github.com/dotnet/runtime/issues/45290#issue-752502603

tintoy: Having said that, as designed, this method swallows any error and returns null (logging being the only place errors are surfaced) so perhaps a warning is more appropriate? just didn’t want to fill up the log with noise… 🤷🏻

src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
tintoy: I think that’s called “message template” format; those expressions are similar to `{0}`, `{1}`, etc, but the names inside the braces become properties of the logged message:

https://learn.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line#log-message-template

tintoy: I hadn’t seen the delegate syntax before; I am guessing it’s shorthand for the log-level check I’m doing outside that statement?

tintoy:
> Another concern is the absence of tests, despite the fact that the current tests are passing.

Yes, I think I can probably whip up a couple of integration tests for this, if nothing else (KubeClient has a testability framework that can be used to mock out the Kubernetes API for tests using a custom `HttpMessageHandler` or a `TestServer` so no actual sockets required).

I’ll try to get the tests done sometime today, if I can (it’s Sunday here, and next week is already starting to look pretty heavy).

src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
tintoy: Oh, I see - this is `IOcelotLogger`, which expects the log message to already have been formatted (and therefore does not take format arguments).

No problem, I can switch to using that instead `ILogger`, if you prefer.

src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
```diff
@@ -31,30 +27,7 @@ public async Task<EndpointsV1> GetAsync(string serviceName, string kubeNamespace
...
-                    errorResponse.Message
-                );
-            }
-
-            return null;
-        }
-
-        return await response.ReadContentAsAsync<EndpointsV1>();
+        return await Http.GetAsync(request, cancellationToken)
+            .ReadContentAsObjectV1Async<EndpointsV1>(operationDescription: $"get {nameof(EndpointsV1)}");
```
tintoy, line 31: This wrapper method (part of KubeClient) will take care of deserialising the response from the server (if the status code does not indicate a success, it will attempt to deserialise the response body as a Kubernetes `StatusV1` and wrap that up in a `KubeApiException`.

src/Ocelot.Provider.Kubernetes/Kube.cs
```diff
@@ -46,9 +48,42 @@ public virtual async Task<List<Service>> GetAsync()
             .ToList();
     }
 
-    private Task<EndpointsV1> GetEndpoint() => _kubeApi
-        .ResourceClient<IEndPointClient>(client => new EndPointClientV1(client))
-        .GetAsync(_configuration.KeyOfServiceInK8s, _configuration.KubeNamespace);
+    private async Task<EndpointsV1> GetEndpoint()
```
tintoy, line 51: I've moved this logic here because it keeps the resource-client implementation cleaner (since it wouldn't otherwise be clear to someone who's not seen this code before what the lifetime of the resource client is, vs the lifetime of the containing type or its logger).

tintoy:
@raman-m I've reworked the logic a bit (see comments on the commit). Can you let me know if you're ok with those changes? If so, I can get started on the tests.

src/Ocelot.Provider.Kubernetes/Ocelot.Provider.Kubernetes.csproj
tintoy: Yeah, fair call - I'd only just published that; will pull in the new version 🙂

tintoy: Hmm - just before I push this change, I thought I'd better check - `KubeClient` `3.0.1` requires `Microsoft.Extensions.*` `9.0.3` (Ocelot's test projects currently use `9.0.1`;  tried upgrading, and all are tests still passing, but it's your call if you want me to upgrade).

tintoy:
Regarding the tests - although there are existing tests for the Kubernetes provider, and I can see how to extend them to test handling of error responses, I'm not confident I can test the logging behaviour without risking reinventing the wheel (I see a couple of of internal-only/test-specific implementations of `IOcelotLogger` scattered across the test codebase, and I'm not sure I have time to consolidate them). I can probably create a stub implementation for the existing tests but I'd definitely prefer to keep the additional complexity to a minimum if you're wanting to release in the next couple of days.

src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
tintoy: Switched to `IOcelotLogger.LogError(Func<string>)`.

src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
tintoy: I've moved that logic out of `EndPointClientV1`, to maintain separation of concerns; it now lives in the caller (`Kube`).

tintoy:
> KubeClient has a testability framework that can be used to mock out the Kubernetes API for tests using a custom HttpMessageHandler or a TestServer so no actual sockets required

Ah, good - I see there's already a framework for serving up a fake Kubernetes API; will use that, instead.

src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
raman-m: That's OK

raman-m:
> @raman-m I've reworked the logic a bit (see comments on the commit). Can you let me know if you're ok with those changes? If so, I can get started on the tests.

Please proceed. It is to upgrade the tests in the unit testing project: [Kubernetes](https://github.com/ThreeMammals/Ocelot/tree/develop/test/Ocelot.UnitTests/Kubernetes).
Additionally, I believe that extensive upgrades to the acceptance tests are not required, as they utilize a testing server. Therefore, I require the upgrading and testing of new internal implementation through unit tests.

> Regarding the tests - although there are existing tests for the Kubernetes provider, and I can see how to extend them to test handling of error responses, I'm not confident I can test the logging behaviour without risking reinventing the wheel (I see a couple of of internal-only/test-specific implementations of IOcelotLogger scattered across the test codebase, and I'm not sure I have time to consolidate them). I can probably create a stub implementation for the existing tests but I'd definitely prefer to keep the additional complexity to a minimum if you're wanting to release in the next couple of days.

Please update [unit tests](https://github.com/ThreeMammals/Ocelot/tree/develop/test/Ocelot.UnitTests/Kubernetes) only. There is no need to rewrite or refactor acceptance tests. They should be solid, meaning they require minimal updating if they fail. Therefore, you need to test all new implementations with [unit tests](https://github.com/ThreeMammals/Ocelot/tree/develop/test/Ocelot.UnitTests/Kubernetes).

> Ah, good - I see there's already a framework for serving up a fake Kubernetes API; will use that, instead.

Indeed, we utilize `Microsoft.AspNetCore.TestHost` for acceptance testing, we generate actual requests directed at the test server endpoints.
FYI I have no objections if you use the native framework you proposed for Kube objects, but only for integration/unit testing.

raman-m:
Adam FYI, you probably saw PR #2257 and commit https://github.com/ThreeMammals/Ocelot/commit/4b6b96af6061ff2ea817c5408074baa1ab387735
I am afraid we will need to update the [K8s feature doc](https://github.com/ThreeMammals/Ocelot/blob/develop/docs/features/kubernetes.rst) well. However, I will reconsider this after the Dev Complete stage.

tintoy:
> Therefore, I require the upgrading and testing of new internal implementation through unit tests.

Just to confirm - by this, do you mean unit tests verifying the underlying implementation (using Moq or similar) or do you mean unit tests for the behaviour changes resulting from the implementation change?

tintoy:
> Just to confirm - by this, do you mean unit tests verifying the underlying implementation (using Moq or similar) or do you mean unit tests for the behaviour changes resulting from the implementation change?

Never mind - I see the tests you're referring to; I'll extend those.

test/Ocelot.UnitTests/Kubernetes/KubeTests.cs
```diff
@@ -54,6 +56,66 @@ public async Task Should_return_service_from_k8s()
...
+        string expectedLogMessage = $"Failed to retrieve v1/Endpoints '{given.ProviderOptions.KeyOfServiceInK8s}' in namespace '{given.ProviderOptions.KubeNamespace}' (HTTP.{expectedStatusCode}/Failure/{expectedStatusCode}): {expectedKubeApiErrorMessage}";
+        _logger.Setup(logger => logger.LogError(It.IsAny<Func<string>>(), It.IsAny<Exception>()))
+            .Callback((Func<string> messageFactory, Exception exception) =>
+            {
+                messageFactory.ShouldNotBeNull();
+
+                string logMessage = messageFactory();
+                logMessage.ShouldNotBeNullOrWhiteSpace();
+
+                // This is a little fragile, as it may change if other entries are logged due to implementation changes.
```
tintoy, line 91: @raman-m small issue, as noted in these comments.

tintoy:
> Adam FYI, you probably saw PR #2257 and commit [4b6b96a](https://github.com/ThreeMammals/Ocelot/commit/4b6b96af6061ff2ea817c5408074baa1ab387735) I am afraid we will need to update the [K8s feature doc](https://github.com/ThreeMammals/Ocelot/blob/develop/docs/features/kubernetes.rst) well.

No worries, happy to discuss / fill in details if required.

Potentially useful KubeClient docs for registering `IKubeApiClient` / `KubeClientOptions`:

https://github.com/tintoy/dotnet-kube-client/?tab=readme-ov-file#make-the-client-available-for-dependency-injection

test/Ocelot.UnitTests/Kubernetes/KubeTests.cs
raman-m: Wow!... 😮

raman-m:
@tintoy commented [on March 17](https://github.com/ThreeMammals/Ocelot/pull/2266#issuecomment-2727733771)
> 
> No worries, happy to discuss / fill in details if required.
> 
> Potentially useful KubeClient docs for registering `IKubeApiClient` / `KubeClientOptions`:
> 
> https://github.com/tintoy/dotnet-kube-client/?tab=readme-ov-file#make-the-client-available-for-dependency-injection

No, I meant: How will your changes be reflected in the current [K8s doc](https://ocelot.readthedocs.io/en/develop/features/kubernetes.html) for the development version? Please note that the development version is still unreleased. Are these documents relevant? Please review them.

tintoy:
> No, I meant: How will your changes be reflected in the current [K8s doc](https://ocelot.readthedocs.io/en/develop/features/kubernetes.html) for the development version? Please note that the development version is still unreleased. Are these documents relevant? Please review them

Oh - sure, will do.

raman-m (approved):
I believe the development is complete. The final step is review the documentation.

src/Ocelot.Provider.Kubernetes/Kube.cs
```diff
@@ -46,9 +48,42 @@ public virtual async Task<List<Service>> GetAsync()
...
+            return await _kubeApi
+                .ResourceClient<IEndPointClient>(client => new EndPointClientV1(client))
+                .GetAsync(serviceName, kubeNamespace);
+        }
+        catch (KubeApiException kubeApiError)
+        {
+            _logger.LogError(() =>
+            {
+                StatusV1 errorResponse = kubeApiError.Status;
+                string httpStatusCode = "Unknown";
```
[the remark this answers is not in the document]
raman-m: Fixed in https://github.com/ThreeMammals/Ocelot/pull/2266/commits/c853a6f25ffaef8c2174b4e6ac6fdb8ce064ad2d

src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
```diff
@@ -21,17 +20,14 @@ public async Task<EndpointsV1> GetAsync(string serviceName, string kubeNamespace
...
                 Namespace = kubeNamespace ?? KubeClient.DefaultNamespace,
                 ServiceName = serviceName,
             });
 
-        var response = await Http.GetAsync(request, cancellationToken);
-
-        return response.IsSuccessStatusCode
-            ? await response.ReadContentAsAsync<EndpointsV1>()
-            : null;
+        return await Http.GetAsync(request, cancellationToken)
```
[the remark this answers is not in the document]
tintoy: I’ll check the `ReadContentAsObjectV1Async` method when I get back to my computer; as long as it doesn’t return one of the result types that have special casting behaviour that’s probably fine.

test/Ocelot.UnitTests/Kubernetes/KubeTests.cs
tintoy: The problem (as it looks to me) is that the test can't know for sure what other components may log messages as well (in this case, the retry facility uses the same logger and trying to verify those logged messages would make the whole thing too fragile by making this unit test coupled to the implementation of other components such as the retry).

I've always aimed to make unit tests, in particular, to test behaviour without being coupled to the implementation; otherwise the tests may have to be changed each time the implementation changes which removes one of the (in my opinion) best benefits of unit testing.

But I'm not that familiar with the code-base and so I'm open to changing it if you believe that would be better 🙂

src/Ocelot.Provider.Kubernetes/EndPointClientV1.cs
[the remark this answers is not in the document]
tintoy: Yes, you're right - we can skip the await here and make this method a non-async wrapper.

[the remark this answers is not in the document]
tintoy: Done.

tintoy:
Regarding the documentation, it looks to me like no changes are required as a result of this PR's changes.

Later, however, it might be worth adding some information to the docs about KubeClient's support for:
* [Dependency-injection](https://github.com/tintoy/dotnet-kube-client/?tab=readme-ov-file#make-the-client-available-for-dependency-injection)
* [`~/.kube/config`](https://github.com/tintoy/dotnet-kube-client/blob/ef065dd8069c721e4772ec104b9cc063a8bd885c/test/KubeClient.Extensions.KubeConfig.Tests/OptionsTests.cs#L38) 

For example, the code from step 2:

```csharp
Action<KubeClientOptions> configureKubeClient = opts =>
 {
     opts.ApiEndPoint = new UriBuilder("https", "my-host", 443).Uri;
     opts.AccessToken = "my-token";
     opts.AuthStrategy = KubeAuthStrategy.BearerToken;
     opts.AllowInsecure = true;
 };
 builder.Services
     .AddOptions<KubeClientOptions>()
     .Configure(configureKubeClient); // manual binding options via IOptions<KubeClientOptions>
```

Could also be written like this:

```csharp
services.AddKubeClientOptions(clientOptions =>
{
    clientOptions.ApiEndPoint = new UriBuilder("https", "my-host", 443).Uri;
    clientOptions.AuthStrategy = KubeAuthStrategy.BearerToken;
    clientOptions.AccessToken = "my-token";
    clientOptions.AllowInsecure = true;
});
```

test/Ocelot.UnitTests/Kubernetes/KubeTests.cs
raman-m: So, the issue lies in the retry logic of the upper context, correct? It seems we have been debating this extensively. If the problem truly exists, I will identify it in the next round of refactoring the logic. For now, we can omit this.

raman-m:
@tintoy commented [on March 18](https://github.com/ThreeMammals/Ocelot/pull/2266#issuecomment-2731215776)

> Regarding the documentation, it looks to me like no changes are required as a result of this PR's changes.

Great! If no changes needed when upgrading to v3 version then I have a plan to release current docs with this extra method:
- [AddKubernetes(Action&lt;KubeClientOptions&gt;) method](https://ocelot.readthedocs.io/en/develop/features/kubernetes.html#addkubernetes-action-kubeclientoptions-method) (doc in develop)

---

> Later, however, it might be worth adding some information to the docs about KubeClient's support for:
> * [Dependency-injection](https://github.com/tintoy/dotnet-kube-client/?tab=readme-ov-file#make-the-client-available-for-dependency-injection)
> * [`~/.kube/config`](https://github.com/tintoy/dotnet-kube-client/blob/ef065dd8069c721e4772ec104b9cc063a8bd885c/test/KubeClient.Extensions.KubeConfig.Tests/OptionsTests.cs#L38) 
> 
> For example, the code from step 2:
> XXX
> 
> Could also be written like this:
> YYY
> 

Good suggestion. However, the Ocelot documentation recommends adding Kubernetes support using the native Ocelot `AddKubernetes` method with two overloaded versions (see docs). Therefore, as a team, we encapsulate all native client settings and pass only high-level required options to the internal setup.

Switching to `services.AddKubeClientOptions` method will provide no benefirts due to the internally encapsulated setup of the integrated client. Another point regarding Step 2 is explaining to users how it works in the case of manual setup. However, currently we do not draw users' attention to the use of the native helper with the Action-delegate argument. Therefore, I will explain this briefly in Note 2 of the doc...

raman-m (approved):
### Ready for delivery ✔️ 
@tintoy Congrats! 🎉

PR #2308, opened 2025-08-06
remove write only filed _upstreamHttpMethod of DownstreamRouteBuilder

harris2012:
Fixes / New Feature #

## Proposed Changes

  -
  -
  -


src/Ocelot/Configuration/Creator/RoutesCreator.cs
raman-m, line 1: There are a lot of fictional changes caused by the [EOL issue](https://ocelot.readthedocs.io/en/latest/building/devprocess.html#eol-gotchas).

src/Ocelot/Configuration/Creator/RoutesCreator.cs
```diff
@@ -117,16 +117,15 @@ private DownstreamRoute SetUpDownstreamRoute(FileRoute fileRoute, FileGlobalConf
...
 
         var cacheOptions = _cacheOptionsCreator.Create(fileRoute.FileCacheOptions, globalConfiguration, fileRoute.UpstreamPathTemplate, fileRoute.UpstreamHttpMethod);
 
-        var metadata = _metadataCreator.Create(fileRoute.Metadata, globalConfiguration);
+        var metadata = _metadataCreator.Create(fileRoute.Metadata, globalConfiguration);
 
         var route = new DownstreamRouteBuilder()
             .WithKey(fileRoute.Key)
             .WithDownstreamPathTemplate(fileRoute.DownstreamPathTemplate)
-            .WithUpstreamHttpMethod(fileRoute.UpstreamHttpMethod)
```
raman-m, line 129: Alright, did you take out this line because it was your objective?
RoutesCreator is responsible for creating both internal Ocelot Core business objects, `Route` and `DownstreamRoute`. The proposal to remove information from `DownstreamRoute` is based on the fact that the property is not used in further logic. However, there doesn't seem to be a strong motivation for this removal. Since it is already implemented, it could theoretically be used in subsequent logic involving the `DownstreamRoute` object.

src/Ocelot/Configuration/Builder/DownstreamRouteBuilder.cs
```diff
@@ -89,14 +88,6 @@ public DownstreamRouteBuilder WithUpstreamPathTemplate(UpstreamPathTemplate inpu
         return this;
     }
 
-    public DownstreamRouteBuilder WithUpstreamHttpMethod(List<string> input)
```
raman-m, line 92: There is a single reference to RoutesCreator in the Core, along with 28 references to testing classes. My intention has been clarified, so let's keep this code unchanged.

raman-m (commented):
This PR is not linked to any existing issues or discussions, which deviates from our development process. While exceptions might be made for bug fixes, typo corrections, or genuinely valuable improvements, I don’t see any tangible benefits in this case. I also don’t see a reason to suggest changes or improvements, so I’m unable to approve it.
Apologies, but this PR will be closed...

src/Ocelot/Configuration/Creator/RoutesCreator.cs
harris2012: all right.

src/Ocelot/Configuration/Builder/DownstreamRouteBuilder.cs
harris2012: It's ok.

PR #2314, opened 2025-09-15
#1040 Add authentication and authorization to web socket pipeline

### LINKED ISSUE #1040: Authentication with SignalR
Thanks for the good work!

##  New Feature
Is authentication for signalR on the plans? We are using signalR with IdentityServer4, to send data over to authenticated users, but I don't think Ocelot is supporting it currently, as stated in documentation.

## Motivation for New Feature
I am not sure of people who would use signalR without user authentication.

fplorenzetti:
## Closes #1040 
- #1040 

## Proposed Changes
  - Add support to authentication and authorization for [Web sockets](https://github.com/ThreeMammals/Ocelot/blob/develop/docs/features/websockets.rst)
  - Add [PreQueryStringBuilderMiddleware](https://github.com/search?q=repo%3AThreeMammals%2FOcelot+PreQueryStringBuilderMiddleware&type=code) support

### DIFF
--- a/src/Ocelot/Middleware/OcelotPipelineExtensions.cs
+++ b/src/Ocelot/Middleware/OcelotPipelineExtensions.cs
@@ -34,16 +34,23 @@ public static class OcelotPipelineExtensions
         app.UseMiddleware<ExceptionHandlerMiddleware>();
 
         // If the request is for websockets upgrade we fork into a different pipeline
-        app.MapWhen(httpContext => httpContext.WebSockets.IsWebSocketRequest,
+        app.MapWhen(
+            httpContext => httpContext.WebSockets.IsWebSocketRequest,
             ws =>
             {
                 ws.UseMiddleware<DownstreamRouteFinderMiddleware>();
                 ws.UseMiddleware<MultiplexingMiddleware>();
                 ws.UseMiddleware<DownstreamRequestInitialiserMiddleware>();
+                ws.UseIfNotNull(configuration.PreAuthenticationMiddleware);
+                ws.UseIfNotNull<AuthenticationMiddleware>(configuration.AuthenticationMiddleware);
+                ws.UseIfNotNull(configuration.PreAuthorizationMiddleware);
+                ws.UseIfNotNull<AuthorizationMiddleware>(configuration.AuthorizationMiddleware);
+                ws.UseIfNotNull(configuration.PreQueryStringBuilderMiddleware);
                 ws.UseMiddleware<LoadBalancingMiddleware>();
                 ws.UseMiddleware<DownstreamUrlCreatorMiddleware>();
-                ws.UseMiddleware<WebSocketsProxyMiddleware>();
-            });
+                ws.UseMiddleware<WebSocketsProxyMiddleware>(); // websockets pipeline stops here.
+            }
+        );
 
         // Allow the user to respond with absolutely anything they want.
         app.UseIfNotNull(configuration.PreErrorResponderMiddleware);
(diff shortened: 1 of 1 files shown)


src/Ocelot/Middleware/OcelotPipelineExtensions.cs
```diff
@@ -40,10 +40,16 @@ public static RequestDelegate BuildOcelotPipeline(this IApplicationBuilder app,
...
+                ws.UseIfNotNull(configuration.PreAuthenticationMiddleware);
+                ws.UseIfNotNull<AuthenticationMiddleware>(configuration.AuthenticationMiddleware);
+                ws.UseIfNotNull(configuration.PreAuthorizationMiddleware);
+                ws.UseIfNotNull<AuthorizationMiddleware>(configuration.AuthorizationMiddleware);
+                ws.UseIfNotNull(configuration.PreQueryStringBuilderMiddleware);
                 ws.UseMiddleware<LoadBalancingMiddleware>();
                 ws.UseMiddleware<DownstreamUrlCreatorMiddleware>();
-                ws.UseMiddleware<WebSocketsProxyMiddleware>();
-            });
+                ws.UseMiddleware<WebSocketsProxyMiddleware>(); // websockets pipeline stops here.
```
raman-m, line 50: Interesting observation...
Let's conduct some code research. I've considered the position of this pipeline branching, and it seems the most appropriate location would be at the very beginning of the current method.
**Could you try moving this branch to the start of the method?**
This would correspond to the current line 29.

fplorenzetti: Actually, I think it is possible to apply all the ocelot middlewares to the websockets, so it could be moved to the line 129:

```cs
// If the request is for websockets, we fork into a different pipeline
app.MapWhen(httpContext => httpContext.WebSockets.IsWebSocketRequest,
ws =>
{
   ws.UseMiddleware<WebSocketsProxyMiddleware>(); // websockets pipeline stops here.
}

// Not sure if this is the best place for this but we use the downstream url
// as the basis for our cache key.
app.UseMiddleware<OutputCacheMiddleware>();

//We fire off the request and set the response on the scoped data repo
app.UseMiddleware<HttpRequesterMiddleware>();
```

I need to know your expectations for this,  can we reuse the authentication / authorization configuration from aspnet pipeline or do you want this inside ocelot pipeline?

raman-m: Don't assume, rather develop.

raman-m:
@fplorenzetti commented [on Sep 17](https://github.com/ThreeMammals/Ocelot/pull/2314#issuecomment-3303329409)

> Your idea is to have all the implementation of the JWT bearer authentication inside the ocelot pipeline? With the changes in this PR, the users can configure their application as the following:

Should I be congratulating you on this code?

fplorenzetti:
> @fplorenzetti commented [on Sep 17](https://github.com/ThreeMammals/Ocelot/pull/2314#issuecomment-3303329409)
> 
> > Your idea is to have all the implementation of the JWT bearer authentication inside the ocelot pipeline? With the changes in this PR, the users can configure their application as the following:
> 


> Should I be congratulating you on this code?



> @fplorenzetti commented [on Sep 17](https://github.com/ThreeMammals/Ocelot/pull/2314#issuecomment-3303329409)
> 
> > Your idea is to have all the implementation of the JWT bearer authentication inside the ocelot pipeline? With the changes in this PR, the users can configure their application as the following:
> 

Look, my idea was to allow the usage of Aspnet authentication middleware for websockets doing the minimum modification in the Ocelot code. If it is not enough, go ahead with your suggestions.
 
> Should I be congratulating you on this code?

I don't know what is your point by writing these kind of random comments but I don't think it helps improving the library that your are managing.

fplorenzetti:
Ok, I understand, good luck with your fascinating and with a rich history library!

PR #2326, opened 2025-10-09
Fix #2325: Prioritize service instance address over Consul node address

### LINKED ISSUE #2325: Gateway routes traffic to Consul address instead of service instance address in Ocelot 23.3+
This issue is particularly problematic in Kubernetes environments where downstream services and Consul are deployed in different locations, but affects any environment with separated Consul and service deployments.

## Describe the bug
After upgrading Ocelot from version 23.2.x to 23.3.x or newer (including 24.x), the gateway incorrectly routes traffic to the Consul node address instead of the actual service instance address registered in Consul. This issue is particularly problematic in Kubernetes environments where downstream services and Consul are deployed in different locations.

## Steps to reproduce
1. Set up Ocelot 23.3+ with Consul service discovery in a Kubernetes environment
2. Deploy Consul and downstream services in different locations/pods
3. Register the microservices with Consul
4. Configure Ocelot to use Consul for service discovery
5. Make a request through Ocelot to a downstream service
6. Observe that Ocelot attempts to connect to the Consul node address instead of the service instance address
**Note:** This issue does NOT occur in Ocelot 23.2.x and earlier versions.

## Expected behavior
Ocelot should route requests to the service instance address provided by Consul (`entry.Service.Address`) rather than using the Consul node address (`node.Address`).

## Actual behavior
Ocelot routes requests to the Consul node address (`node.Address`) instead of the service instance address (`entry.Service.Address`), causing connection failures when the Consul nodes are not reachable from the gateway or when the service is deployed on different nodes than Consul.

## Environment
- **Ocelot Version**: 23.3.0 and later
- **Platform**: Kubernetes (any distribution)
- **Consul Version**: Any version
- **.NET Version**: Not relevant (affects all .NET versions)

## Root Cause Analysis
The regression was introduced in Ocelot 23.3 in the `DefaultConsulServiceBuilder.GetDownstreamHost` method. The current implementation incorrectly prioritizes the Consul node address over the service instance address:

```csharp
// Current broken implementation (Ocelot 23.3+)
protected virtual string GetDownstreamHost(ServiceEntry entry, Node node)
    => node != null ? node.Name : entry.Service.Address;
```

## Proposed Fix
The logic should be reversed to prioritize the service instance address, with node address as fallback only when service address is unavailable:

```csharp
// Fixed implementation
protected virtual string GetDownstreamHost(ServiceEntry entry, Node node)
    => !string.IsNullOrEmpty(entry?.Service?.Address)
        ? entry.Service.Address 
        : node?.Address ?? node?.Name;
```

This change:
1. **Restores previous behavior** from Ocelot 23.2.x and earlier
2. **Maintains backward compatibility** for environments where service address might be empty
3. **Fixes Kubernetes deployments** where services and Consul run in separate pods
4. **Preserves fallback behavior** for edge cases where service address is unavailable by using node address as backup

## Additional context
This issue is particularly critical in modern microservices architectures where:
- Services are containerized and dynamically scheduled across nodes
- Consul runs as a separate cluster or stateful set
- Network segmentation exists between service discovery and application layers
- Service mesh architectures rely on accurate endpoint discovery

## Related to #2208 
- #2208

rongjieaaa:
## Description
Fix for issue #2325: Gateway routes traffic to Consul address instead of service instance address in Ocelot 23.3+

## Problem
After upgrading to Ocelot 23.3+, the gateway incorrectly routes traffic to Consul node addresses instead of service instance addresses registered in Consul. This breaks service discovery in Kubernetes and other distributed environments where Consul and services are deployed separately.

## Root Cause
The regression was introduced in Ocelot 23.3 in the `DefaultConsulServiceBuilder.GetDownstreamHost` method, which incorrectly prioritizes `node.Name` over `entry.Service.Address`.

**Broken implementation:**
```csharp
protected virtual string GetDownstreamHost(ServiceEntry entry, Node node)
    => node != null ? node.Name : entry.Service.Address;
```

## Solution
Reverse the priority to use service instance address as primary, with node address as fallback only when service address is unavailable:

**Fixed implementation:**
```csharp
protected virtual string GetDownstreamHost(ServiceEntry entry, Node node)
    => !string.IsNullOrEmpty(entry?.Service?.Address)
        ? entry.Service.Address 
        : node?.Address ?? node?.Name;
```

## Changes Made
- Modified `DefaultConsulServiceBuilder.GetDownstreamHost` method
- Prioritize `entry.Service.Address` over node addresses
- Add proper null/empty checks for service address
- Maintain backward compatibility with fallback mechanism

## Testing Performed
- ✅ **Local Testing**: Built custom NuGet package and deployed in our Kubernetes environment
- ✅ **Functional Verification**: Confirmed gateway now correctly routes to service instance addresses

## Impact
- **Restores Ocelot 23.2.x behavior** for Consul service discovery
- **Fixes Kubernetes deployments** with separated Consul and service pods
- **Maintains compatibility** with all existing environments
- **No breaking changes** - only fixes incorrect routing logic

## Additional Context
This issue affects any environment where:
- Consul cluster is deployed separately from application services
- Services are dynamically scheduled across different nodes

Fixes #2325

### DIFF
--- a/src/Ocelot.Provider.Consul/DefaultConsulServiceBuilder.cs
+++ b/src/Ocelot.Provider.Consul/DefaultConsulServiceBuilder.cs
@@ -94,9 +94,9 @@ public class DefaultConsulServiceBuilder : IConsulServiceBuilder
             entry.Service.Port);
 
     protected virtual string GetDownstreamHost(ServiceEntry entry, Node node)
-        => !string.IsNullOrEmpty(entry?.Service?.Address)
-            ? entry.Service.Address
-            : node?.Address ?? node?.Name;
+        => !string.IsNullOrEmpty(entry?.Service?.Address) ? entry.Service.Address
+            : !string.IsNullOrEmpty(node?.Address) ? node.Address
+            : node?.Name;
 
     protected virtual string GetServiceId(ServiceEntry entry, Node node)
         => entry.Service.ID;
--- a/test/Ocelot.UnitTests/Consul/DefaultConsulServiceBuilderTests.cs
+++ b/test/Ocelot.UnitTests/Consul/DefaultConsulServiceBuilderTests.cs
@@ -118,19 +118,44 @@ public sealed class DefaultConsulServiceBuilderTests
     {
         Arrange();
 
-        // Arrange, Act, Assert: node branch
+        // Arrange, Act, Assert: entry.Service.Address is not empty, should return it regardless of node
         ServiceEntry entry = new()
         {
             Service = new() { Address = nameof(GetDownstreamHost_BothBranches_NameOrAddress) },
         };
-        var node = new Node { Name = "test1" };
+        var node = new Node { Name = "test1", Address = "test-address" };
         var actual = GetDownstreamHost.Invoke(sut, new object[] { entry, node }) as string;
+        actual.ShouldNotBeNull().ShouldBe(nameof(GetDownstreamHost_BothBranches_NameOrAddress));
+
+        // Arrange, Act, Assert: entry.Service.Address is empty, node.Address is not empty, should return node.Address
+        Arrange();
+        entry = new()
+        {
+            Service = new() { Address = string.Empty },
+        };
+        node = new Node { Name = "test1", Address = "test-address" };
+        actual = GetDownstreamHost.Invoke(sut, new object[] { entry, node }) as string;
+        actual.ShouldNotBeNull().ShouldBe("test-address");
+
+        // Arrange, Act, Assert: entry.Service.Address is empty, node.Address is empty, should return node.Name
+        Arrange();
+        entry = new()
+        {
+            Service = new() { Address = string.Empty },
+        };
+        node = new Node { Name = "test1", Address = string.Empty };
+        actual = GetDownstreamHost.Invoke(sut, new object[] { entry, node }) as string;
         actual.ShouldNotBeNull().ShouldBe("test1");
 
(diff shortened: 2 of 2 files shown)


src/Ocelot.Provider.Consul/DefaultConsulServiceBuilder.cs
```diff
@@ -94,7 +94,9 @@ protected virtual ServiceHostAndPort GetServiceHostAndPort(ServiceEntry entry, N
             entry.Service.Port);
 
     protected virtual string GetDownstreamHost(ServiceEntry entry, Node node)
-        => node != null ? node.Name : entry.Service.Address;
+        => !string.IsNullOrEmpty(entry?.Service?.Address) ? entry.Service.Address
+            : !string.IsNullOrEmpty(node?.Address) ? node.Address
+            : node?.Name;
```
raman-m, line 99: This approach is far too risky and could potentially disrupt thousands of environments across countless projects worldwide.

rongjieaaa: 我想了一下，确实，因为我们无法预期有用户已经有多少开始基于「node.Name优先」而使用了

Based on google translate:
I thought about it, indeed, because we can't predict how many users have started to use it based on "node.Name first"

raman-m: ## 🤣

PR #2337, opened 2025-11-16
#741 Add filtering options and logic to caching

### LINKED ISSUE #741: output caching  issue
Hi,
An endpoint with caching enabled caches output even if downstream request returns anything other than httpstatus code with 200. Is it by design, or a bug?

twinetek:
## Closes #741  
- #741

## Proposed Changes
  - Added an array of `HttpStatusCode` to `FileCacheOptions` and `CacheOptions` to act as a whitelist of status codes.
  - Before adding the response to the cache, check to see if the response status code is in the whitelist.
  - If the whitelist is null, assume all status codes are to be cached
  - Tests. _to be added_ (Would like some guidance on where to put these. I suspect [OutputCacheMiddlewareTests](https://github.com/ThreeMammals/Ocelot/blob/develop/test/Ocelot.UnitTests/Cache/OutputCacheMiddlewareTests.cs). I've also never used XUnit so there's going to be a bit of a learning curve.)
  - Looking for guidance. I believe this is where and how this cache filtering would be added, but I am new to the project and getting back into coding after several years. I've also never submitted a pull request to someone else's project before so I'm hoping I'm doing this correctly.


twinetek:
Raman,
I will work on implementing the include/exclude logic and the 2xx/3xx/etc. range option. Might take me a few days.
Cliff

twinetek:
Raman,
I've added some filtering logic. You can now specify the filtering and specifying whether to either use the list as a whitelist or a blacklist.

```c#
    public enum FilterType
    {
        Whitelist,
        Blacklist,
    }
```

The filter is added to `CacheOptions` using the new `HttpStatusCodeFilter` class:

```c#
CacheOptions = {
    TtlSeconds = 100,
    StatusCodeFilter = new HttpStatusCodeFilter(FilterType.Whitelist, new HttpStatusCode[] { HttpStatusCode.OK, HttpStatusCode.Forbidden }),
};
```

You can specify the codes in the filter with strings, including the 2xx, 3xx, etc. ranges

```c#
CacheOptions = {
    TtlSeconds = 100,
    StatusCodeFilter = new HttpStatusCodeFilter(FilterType.Whitelist, new string[] { "200", "4xx" }),
};
```

I have added some pretty exhaustive tests. 

Would you like it extended to allow the items in the list as `int`? Or a mixed array of `HttpStatusCode`, `string`, or `int`?

raman-m:
> Would you like it extended to allow the items in the list as `int`? Or a mixed array of `HttpStatusCode`, `string`, or `int`?


Oh no, my dear Copilot! 🤣  
Please revert the last two commits, such as https://github.com/ThreeMammals/Ocelot/pull/2337/commits/cdcf2730e8504ed52c78cd4972029d8fb8398e7a and https://github.com/ThreeMammals/Ocelot/pull/2337/commits/e8e4515a28d93148e15b3304223f352fb3ede13e.

src/Ocelot/Configuration/File/FileCacheOptions.cs
```diff
@@ -23,4 +26,6 @@ public FileCacheOptions(FileCacheOptions from)
     /// <remarks>If <see langword="null"/> then use global configuration with <see langword="false"/> by default.</remarks>
     /// <value><see langword="true"/> if content hashing is enabled; otherwise, <see langword="false"/>.</value>
     public bool? EnableContentHashing { get; set; }
+
+    public HttpStatusCodeFilter StatusCodeFilter { get; set; }
```
raman-m: ### $\color{red}{Critical\ issue\ with\ the\ property\ type!}$

```suggestion
    public int[] StatusCodes { get; set; }
```
Note that the type is not the old `HttpStatusCode[]` but rather `int[]`, an array of integers. If an Ocelot user is not a C# developer, they might not be familiar with the `HttpStatusCode` enumeration and its values!

src/Ocelot/Cache/OutputCacheMiddleware.cs
```diff
@@ -60,8 +60,16 @@ public async Task Invoke(HttpContext httpContext)
         cached = await CreateCachedResponse(downstreamResponse);
 
         var ttl = TimeSpan.FromSeconds(options.TtlSeconds);
-        _outputCache.Add(downStreamRequestCacheKey, cached, options.Region, ttl);
-        Logger.LogDebug(() => $"Finished response added to cache for the '{downstreamUrlKey}' key.");
+        if (options.StatusCodeFilter == null || options.StatusCodeFilter.PassesFilter(cached.StatusCode))
```
raman-m: - The `PassesFilter(cached.StatusCode)` method could serve as a small helper in the `CacheOptions` business object, though it depends on the final design.
- The expression `options.StatusCodeFilter.PassesFilter` gives the impression that an additional filter has been injected into the logic, but that doesn’t seem to be the case. Since we’re not varying filters, we’ll just have a single filter.

src/Ocelot/Cache/OutputCacheMiddleware.cs
```diff
@@ -60,8 +60,16 @@ public async Task Invoke(HttpContext httpContext)
...
+        {
+            _outputCache.Add(downStreamRequestCacheKey, cached, options.Region, ttl);
+            Logger.LogDebug(() => $"Finished response added to cache for the '{downstreamUrlKey}' key.");
```
raman-m, line 66: Please wrap the logging for the Debug scenario with a preprocessor directive, as another team member will likely request this 👇 
```suggestion
#if DEBUG
            Logger.LogDebug(() => $"Finished response added to cache for the '{downstreamUrlKey}' key.");
#endif
```

src/Ocelot/Configuration/Creator/CacheOptionsCreator.cs
```diff
@@ -6,7 +6,7 @@ namespace Ocelot.Configuration.Creator;
 public class CacheOptionsCreator : ICacheOptionsCreator
 {
     public CacheOptions Create(FileCacheOptions options)
-        => new(options?.TtlSeconds, options?.Region, options?.Header, options?.EnableContentHashing);
+        => new(options?.TtlSeconds, options?.Region, options?.Header, options?.EnableContentHashing, options?.StatusCodeFilter);
```
raman-m: It seems I missed this issue in the previous PR. Extending the argument list in a single constructor is not the right approach. It’s better to have a few overloaded constructors that cover most user scenarios.
Here’s what I propose:
```suggestion
        => new(options ?? new());
```
The null-coalescing operator (`??`) is used to handle argument null checks within the new constructor which will be designed by you

src/Ocelot/Configuration/Creator/CacheOptionsCreator.cs
```diff
@@ -58,6 +58,7 @@ protected virtual CacheOptions Merge(FileCacheOptions options, FileCacheOptions
         var header = options.Header.IfEmpty(globalOptions.Header).IfEmpty(CacheOptions.Oc_Cache_Control);
         var ttlSeconds = options.TtlSeconds ?? globalOptions.TtlSeconds;
         var enableHashing = options.EnableContentHashing ?? globalOptions.EnableContentHashing;
-        return new CacheOptions(ttlSeconds, region, header, enableHashing);
+        var statusCodes = options.StatusCodeFilter ?? globalOptions.StatusCodeFilter;
```
raman-m: There's a lack of defaulting! Refer to my Idea 2 in the [previous comment](https://github.com/ThreeMammals/Ocelot/pull/2337#issuecomment-3566745978). We could implement defaulting closer to the Dev Complete stage, and I'll help with that.
Finally, I expect the following improvement:
```suggestion
        var statusCodes = options.StatusCodes ?? globalOptions.StatusCodes ?? Ocelot.RecommendedStatusCodes;
```

src/Ocelot/Filter/IFilter.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Ocelot.Filter
+{
+    public interface IFilter<T>
+    {
+        bool PassesFilter(T value);
+    }
```
raman-m: Are you suggesting using multiple filters for different options?

src/Ocelot/Filter/HttpStatusCodeFilter.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+
+namespace Ocelot.Filter
+{
+    public class HttpStatusCodeFilter : Filter<HttpStatusCode>
```
raman-m: This filter will likely become useless because defining the values of the C# `HttpStatusCode` enumeration is challenging for Ocelot users who aren’t .NET developers. I’ve already raised this issue [here](https://github.com/ThreeMammals/Ocelot/pull/2337#discussion_r2553168624).

raman-m (changes requested):
Some great [tests have been added](https://github.com/ThreeMammals/Ocelot/pull/2337/files#diff-90ea0d8bf4ce64201097435dc591479ddc4533c3e5659ffa2a107af47f7cf474), but we are moving forward with redesigning the feature.

Consider this [comment](https://github.com/ThreeMammals/Ocelot/pull/2337#issuecomment-3566745978) as the updated specification, serving as a revision of the #741 requirements. This isn't the final version, as I'm open to new ideas and discussions.

src/Ocelot/Filter/IFilter.cs
twinetek: I thought perhaps having a generic Filter types could assist in the future. In particular, I saw https://github.com/ThreeMammals/Ocelot/issues/1587 and figured it could be used there, too. The type is unnecessary for the current feature though, so it can be removed.

src/Ocelot/Filter/HttpStatusCodeFilter.cs
twinetek: Yes, I can change all this to be an integer based filter and not an HttpStatusCode enumeration based filter.

src/Ocelot/Configuration/Creator/CacheOptionsCreator.cs
twinetek: Can do. I was just following the existing pattern. Not my project, so did not figure it was my place to make that change.

raman-m: Alright, never mind. I will handle it myself.

src/Ocelot/Filter/IFilter.cs
raman-m: > I thought perhaps having a generic Filter types could assist in the future.

Could assist in future?... Sorry? Are you a member of our team?

> In particular, I saw #1587 and figured it could be used there, too.

No, it is better to focus on the feature you're currently working on rather than trying to enhance or refactor other features and code areas. I've requested community help for #1587 since the author has no intention of contributing, so they have been unassigned.

I think this filter might be useful for `SecurityOptions`, but I am not entirely sure.

raman-m:
> I went with explicit whitelist/blacklist because I wanted to avoid handling a conflict.

Neither I nor the community are interested in what you wanted. The design should be straightforward, but I noticed a lot of code in this project that violates the KISS principle, serving no real purpose and adding unnecessary clutter.

> If you're including specific status codes, then you're already implicitly excluding all the status codes not in the specified list. There's no need to specify other status codes to exclude, since they aren't included. The opposite is true - if you're specifying a blacklist, then that implies that everything not in your blacklist is included. If you would still like me to pursue allowing the user to specify codes to include AND codes to exclude, I am happy to iterate with you on what that logic looks like.

Fair enough! Look at these use cases and you will see there is no conflict. And you are absolutely right; positive codes exclude any unwanted values. No opposite codes here:
```json
"CacheOptions": {
  "TtlSeconds": 888,
  "StatusCodes": [200, 301, 302, 303] // only positive codes to enable caching
}
```
Use case #741 covers the situation where users discussed the option of excluding a scenario. And no opposite codes here:
```json
"CacheOptions": {
  "TtlSeconds": 888,
  "StatusCodes": [-500, -404, -405, -410] // only negative codes to disable caching
}
```
Looks good? The goal isn't to propose another option, like a blacklisting array. Both scenarios can be covered by a single `StatusCodes` array. The rest of the design is handled by a validator, which ensures that mixing positive and negative codes is not allowed, making it impossible to define both including and excluding scenarios.

-----
> I will come back to this in a couple of days.

No need to rush, as this PR will be part of the [Winter'26](https://github.com/ThreeMammals/Ocelot/milestone/12) milestone. The upcoming [.NET 10](https://github.com/ThreeMammals/Ocelot/milestone/13) release will be free of new features, focusing instead on project upgrades, code reorganization, DevOps and possible bug fixes.

Please enjoy your weekend.

src/Ocelot/Configuration/File/FileCacheOptions.cs
raman-m: I've just pushed a new commit https://github.com/ThreeMammals/Ocelot/pull/2337/commits/9013937cef572597615f17dd9a6262ef29c7759e with the fix, but I assume the `HttpStatusCode` enumeration is parsed without issues by the JSON deserializer from the `System.Text.Json` namespace. I am not so sure about the `Newtonsoft.Json` deserializer, which is currently the default parser in Ocelot, though we plan to replace `Newtonsoft.Json` with `System.Text.Json` soon. If Newtonsoft handles C# enumerations well, hopefully, it might be worth rethinking the proposed property. Either way, the `int[] StatusCodes` property will remain the main one. I just commented out the line so we can go over this again.

src/Ocelot/Configuration/File/FileCacheOptions.cs
```diff
@@ -23,4 +23,7 @@ public FileCacheOptions(FileCacheOptions from)
     /// <remarks>If <see langword="null"/> then use global configuration with <see langword="false"/> by default.</remarks>
     /// <value><see langword="true"/> if content hashing is enabled; otherwise, <see langword="false"/>.</value>
     public bool? EnableContentHashing { get; set; }
+
+    //public HttpStatusCode[] StatusCodes { get; set; }
+    public int[] StatusCodes { get; set; }
```
RaynaldM:
```suggestion
    public IReadOnlyCollection<int> StatusCodes { get; set; }
```

RaynaldM (approved):

src/Ocelot/Configuration/File/FileCacheOptions.cs
raman-m: In the File model? 👀 Is this suggestion to create an immutable collection? What is the goal? OK, it is immutable, and what else?
Have you written a unit and acceptance tests for this project?
Could you focus on real problems in this project please?
When does your FR team plan to start contributing to the project?

raman-m:
> @ggnaegi commented [on December 3](https://github.com/ThreeMammals/Ocelot/pull/2337#issuecomment-3606927538)

Would you like to mentor this PR?
OK, good luck!

src/Ocelot/Configuration/CacheOptions.cs
```diff
@@ -53,4 +59,7 @@ public CacheOptions(int? ttlSeconds, string region, string header, bool? enableC
     public bool EnableContentHashing { get; }
 
     public bool UseCache => TtlSeconds > NoSeconds;
+
+    //public HttpStatusCode[] StatusCodes { get; }
+    public int[] StatusCodes { get; }
```
ggnaegi: In this case, for better code maintainability, I would use an HttpStatusCode[]. int[] is appropriate for FileCacheOptions when dealing with input data, but here we’re no longer working with input values — so using HttpStatusCode[] is the better fit.

We could convert the data in the constructor, and maybe check if the statuses exist...
```c#
private static HttpStatusCode[] ConvertToHttpStatusCodes(int[] statusCodes)
    {
        if (statusCodes is null || statusCodes.Length == 0)
        {
            return [];
        }

        return statusCodes.Select(code =>
        {
            if (!Enum.IsDefined(typeof(HttpStatusCode), code))
            {
                throw new ArgumentException($"Invalid HTTP status code: {code}", nameof(statusCodes));
            }
            
            return (HttpStatusCode) code;
        }).ToArray();
    }
```

src/Ocelot/Configuration/CacheOptions.cs
```diff
@@ -32,12 +37,13 @@ public CacheOptions(FileCacheOptions from, string defaultRegion)
...
     /// <param name="header">The header name to control cached value.</param>
     /// <param name="enableContentHashing">The switcher for content hashing. If not speciefied, false value is used by default.</param>
-    public CacheOptions(int? ttlSeconds, string region, string header, bool? enableContentHashing)
+    public CacheOptions(int? ttlSeconds, string region, string header, bool? enableContentHashing, int[] statusCodes)
     {
         TtlSeconds = ttlSeconds ?? NoSeconds;
         Region = region;
         Header = header.IfEmpty(Oc_Cache_Control);
         EnableContentHashing = enableContentHashing ?? false;
+        StatusCodes = statusCodes;
```
ggnaegi: Convert to HttpStatusCode[]

src/Ocelot/Configuration/File/FileCacheOptions.cs
```diff
@@ -23,4 +23,7 @@ public FileCacheOptions(FileCacheOptions from)
     /// <remarks>If <see langword="null"/> then use global configuration with <see langword="false"/> by default.</remarks>
     /// <value><see langword="true"/> if content hashing is enabled; otherwise, <see langword="false"/>.</value>
     public bool? EnableContentHashing { get; set; }
+
+    //public HttpStatusCode[] StatusCodes { get; set; }
```
ggnaegi, line 27: Remove the comment, and keep int[] for input data -> input data: string, bool, int etc.

ggnaegi (changes requested):
For me, the code is fine, but I have some concerns about its maintainability, because CachedResponse uses HttpStatusCode while here we’re using an array of int. I think it would be more appropriate to convert the array in CacheOptions and then use HttpStatusCode consistently throughout.

src/Ocelot/Configuration/CacheOptions.cs
twinetek: Done as suggested.

src/Ocelot/Configuration/CacheOptions.cs
twinetek: Done.

src/Ocelot/Configuration/File/FileCacheOptions.cs
twinetek: Done.

raman-m:
@twinetek commented [on Dec 10, 2025](https://github.com/ThreeMammals/Ocelot/pull/2337#issuecomment-3635379422)
> @ggnaegi - any other changes you'd like to see? Happy to make them.

Cliff,

I'd appreciate it if you'd ask for my input directly, rather than referencing Guillaume. He is no longer on the team and hasn't contributed [since Dec 2024](https://github.com/ThreeMammals/Ocelot/commits?author=ggnaegi). Please communicate with me going forward.

I'd also be very glad if you kept your PR branch in sync with develop, since there are quite a few conflicts. I'm happy to help resolve them.

**P.S.** My [earlier code review](https://github.com/ThreeMammals/Ocelot/pull/2337#pullrequestreview-3496781205) can be considered closed, as it focused on a "Filter feature" that we can handle in a separate PR later.

PR #2350, opened 2025-12-28
#2327 Handle null service name in RoundRobin constructor

### LINKED ISSUE #2327: Unit test failures and ArgumentNullException with .NET 10.0.100-preview.7.25351.101 SDK
**Just a heads up** 

### Summary
When running Ocelot unit and acceptance tests using dotnet-sdk-10.0.100-preview.7.25351.101, two tests fail, after investigation, it is caused by the breaking change https://github.com/dotnet/docs/issues/46890 .If you'd like to work around this on your end, you can either handle null values in your code or update your configuration files to use empty strings instead of null, e.g.:
`"ServiceName": ""`

### Details
- **SDK:** dotnet-sdk-10.0.100-preview.7.25351.101
- **OS:** Windows 11 24H2 x64
- **Test Project:** Ocelot.AcceptanceTests.dll 
- **Failing Tests:**
  - `Ocelot.AcceptanceTests.WebSocketTests.ShouldProxyWebsocketInputToDownstreamServiceAndUseLoadBalancer`
  - `Ocelot.AcceptanceTests.LoadBalancerTests.Should_load_balance_request_with_round_robin`

### Reproduction Steps
1. Install only dotnet-sdk-10.0.100-preview.7.25351.101 and set `DOTNET_ROLL_FORWARD_ON_NO_CANDIDATE_FX=2`.
2. Update `Ocelot.AcceptanceTests.runtimeconfig.json` frameworks section to reference SDK 10.0.0-preview.7.25351.101.
3. Build and run the acceptance tests with `dotnet vstest Ocelot.AcceptanceTests.dll`.
4. Observe the two test failures described below.

### Error Messages
it will will throw `"System.ArgumentNullException: 'Value cannot be null. (Parameter 'key')'"` exception from source code.

Tests pass on:
  - dotnet-sdk-8.0.411
  - dotnet-sdk-10.0.100-preview.6.25326.107
- Fail on:
  - dotnet-sdk-10.0.100-preview.7.25351.101

### Additional Findings
- In previous SDK versions, the parameter passed to `TryScanNext()` in `RoundRobin.cs` was an empty string. In preview 7 it is now null, causing the `ArgumentNullException`.

Please get details from https://github.com/dotnet/runtime/issues/117316 .

SharaawyMohamed:
## Fixes #2327 
- #2327 

## Proposed Changes
- Converts null service names to empty string
- Prevents `ArgumentNullException` when `IConfiguration` returns null


src/Ocelot/LoadBalancer/Balancers/RoundRobin.cs
```diff
@@ -18,7 +18,7 @@ public RoundRobin(Func<Task<List<Service>>> services, string serviceName)
     {
         ArgumentNullException.ThrowIfNull(services);
         _servicesDelegate = services;
-        _serviceName = serviceName;
+        _serviceName = serviceName?? string.Empty;
```
raman-m, line 21: Wow, is this really the bug fix? 🤣  
It doesn’t seem related to the .NET SDK version. Can it be reproduced in .NET 9?

SharaawyMohamed: Why are the CI tests failing when all tests pass successfully on my local machine?

raman-m: Don't warry, the build is green now. Use "Re-run job" button.

> Why are the CI tests failing when all tests pass successfully on my local machine?

Yes, tests pass reliably locally and remain stable in the CI environment. A few CPU-sensitive tests might occasionally fail even on local runs, but overall, the tests are optimized to perform well in both environments.
The fact is, we have many tests in Ocelot Core that run parallel logic and are sensitive to thread synchronization. Syncing threads can be unpredictable, but usually, a simple re-run does the trick.

raman-m:
Will you hold off on adding the `net10.0` TFM to the projects?
The problem now is that it’s impossible to verify if the fix works without the `net10.0` TFM.

raman-m:
@SharaawyMohamed
Please check the linked issue. I’ve closed it because I couldn’t reproduce the bug, and all acceptance tests are passing after upgrading to .NET 10. Do you think this pull request still makes sense?

PR #2351, opened 2025-12-29
#2346 Fix problem with downstream URLs where query parameter names get messed up if they contain a placeholder that has a non-empty value

### LINKED ISSUE #2346: Query parameter names containing id are corrupted when another route has {id} placeholder (Regression in 23.4.3)
After upgrading to Ocelot [23.4.3](https://github.com/ThreeMammals/Ocelot/releases/tag/23.4.3), query parameters whose names contain the substring id (e.g. customer_id) are incorrectly rewritten when another route with a {id} path parameter exists.

This causes query parameters to be corrupted and forwarded incorrectly to downstream services.

### Routes Configuration
```json
{
  "Routes": [
    {
      "UpstreamPathTemplate": "/finance/v1/payment-methods/{id}",
      "DownstreamPathTemplate": "/v1/payment-methods/{id}",
      "DownstreamScheme": "http",
      "DownstreamHostAndPorts": [
        {
          "Host": "localhost",
          "Port": 5003
        }
      ]
    },
    {
      "UpstreamPathTemplate": "/finance/v1/payment-methods",
      "DownstreamPathTemplate": "/v1/payment-methods",
      "DownstreamScheme": "http",
      "DownstreamHostAndPorts": [
        {
          "Host": "localhost",
          "Port": 5003
        }
      ]
    }
  ]
}
```

### Steps to Reproduce

Call the second route with a query parameter:
`GET /finance/v1/payment-methods?customer_id=5172980`

### Expected Behavior

Downstream request:
`/v1/payment-methods?customer_id=5172980`

Query parameters must remain unchanged.

### Actual Behavior

Ocelot rewrites the query parameter:
`/v1/payment-methods?customer5172980`

The `id` substring in `customer_id` is incorrectly treated as a route token and removed.

### Ocelot Version

- [23.4.3](https://github.com/ThreeMammals/Ocelot/releases/tag/23.4.3)

### Previously Working

- Yes (22.x and earlier 23.x)

bhargav-polara:
## Fixes #2346 
- #2346 

### Background

After upgrading to Ocelot [23.4.3](https://github.com/ThreeMammals/Ocelot/releases/tag/23.4.3), query parameters whose names contain the substring `id` (e.g., `customer_id`) are incorrectly rewritten or corrupted if another route exists with a `{id}` path parameter. See #2346 for details and repro steps.

### Problem

When Ocelot tries to remove consumed query string parameters from the downstream request, it was using a simple string `Contains` check. This caused substring matches (like the `id` in `customer_id`) to be incorrectly interpreted as route tokens—resulting in corruption (e.g., `customer_id=5172980` becoming `customer5172980`).

### Solution

**Changed logic in `RemoveQueryStringParametersThatHaveBeenUsedInTemplate` to use a Regex match instead of string `Contains`.**  
This ensures only exact parameter matches are identified and removed, and substrings like `id` in `customer_id` are preserved correctly.

```csharp
if (!Regex.IsMatch(
        downstreamRequest.Query,
        $@"(?<=^|\?|&){Regex.Escape(parameter)}",
        RegexOptions.IgnoreCase))
{
    continue;
}
```

---

### How to Reproduce

See steps in #2346:

1. Define two routes:
    - One with `/finance/v1/payment-methods/{id}`
    - One with `/finance/v1/payment-methods`
2. Call GET `/finance/v1/payment-methods?customer_id=5172980`
3. The downstream request should forward the query string unmodified.

### Testing

- Manually tested scenarios from [2346](https://github.com/ThreeMammals/Ocelot/issues/2346); parameters like `customer_id=5172980` are now correctly forwarded downstream without corruption.
- All existing unit and integration tests pass (`dotnet test`).

### Checklist
- [x] Branch based on `develop`
- [x] Code follows contribution guidelines
- [x] Tests passing


raman-m:
> ### Problem
> When Ocelot tries to remove consumed query string parameters from the downstream request, **it was using a simple string `Contains` check. This caused substring matches (like the `id` in `customer_id`) to be incorrectly interpreted as route tokens**—resulting in corruption (e.g., `customer_id=5172980` becoming `customer5172980`).
> 

Right! It was my fault when releasing the Ocelot [23.4.3](https://github.com/ThreeMammals/Ocelot/releases/tag/23.4.3) patch. Sometimes, when you solve a big problem, there’s a chance you might introduce a small, unexpected one 👶

My idea to use the `Contains` method works/worked well when we also check that both names are the same length. This means the `Contains` method requires both name lengths to be equal. However, we still need to parse the parameter in the `downstreamRequest.Query` string or verify its existence in some way. I think the easiest solution is to reuse the functionality of the `HttpContext.Request.Query` object 👇 
```csharp
IQueryCollection query = context.Request.Query;
bool exists = query.ContainsKey(name) && query.TryGetValue(name, out var pvalue) && pvalue == nAndV.Value;
if (!exists) // !downstreamRequest.Query.Contains(parameter)
{
    continue;
}
```
To do that, we just need an additional `context` argument, and that’s all. Parsing the parameter value doesn’t seem necessary, though it might depend. The algorithm following these lines could also be optimized.
https://github.com/ThreeMammals/Ocelot/blob/9fc4e78d314ad97d836b35c9b9bb188dba49d556/src/Ocelot/DownstreamUrlCreator/DownstreamUrlCreatorMiddleware.cs#L135

src/Ocelot/DownstreamUrlCreator/DownstreamUrlCreatorMiddleware.cs
```diff
@@ -132,7 +132,10 @@ protected static void RemoveQueryStringParametersThatHaveBeenUsedInTemplate(Down
         {
             var name = nAndV.Name.Trim(OpeningBrace, ClosingBrace);
             var parameter = $"{name}={nAndV.Value}";
-            if (!downstreamRequest.Query.Contains(parameter))
+            if (!Regex.IsMatch(
+                    downstreamRequest.Query,
+                    $@"(?<=^|\?|&){Regex.Escape(parameter)}",
+                    RegexOptions.IgnoreCase))
```
raman-m: The original idea in the [23.4.3](https://github.com/ThreeMammals/Ocelot/releases/tag/23.4.3) patch was to reduce the number of `Regex` objects. However, the non-cached version of the regular expression builds the `Regex` on the fly and then reuses it for matching. This is a resource-intensive algo that consumes both CPU and RAM. From what I understand, your solution uses a `Regex` to ensure there are no `=`, `?`, or `&` characters in the full query string before the parameter. In the end, I found this approach to be costly in terms of CPU & RAM usage.

And here’s the dilemma: should we optimize `Regex` usage by adding a cache, or explore [other potential Regex-free solutions](https://github.com/ThreeMammals/Ocelot/pull/2351#issuecomment-3703775650) instead?

test/Ocelot.UnitTests/DownstreamUrlCreator/DownstreamUrlCreatorMiddlewareTests.cs
```diff
@@ -662,6 +662,37 @@ public async Task ShouldNotFailToHandleUrlWithSpecialRegexChars(string urlPath)
         Assert.Equal((int)HttpStatusCode.OK, _httpContext.Response.StatusCode);
     }
 
+    [Fact]
+    [Trait("Bug", "2346")]
+    public async Task Should_not_corrupt_query_parameter_names_containing_id_when_route_has_id_placeholder()
```
raman-m: Having a single unit test is fine, but what about covering more edge cases from the original issue and perhaps adding even more scenarios?
Converting to a `Theory`?

raman-m (changes requested):
While this solution should fix the bug, it looks like we still need more thorough unit testing with different URLs. Additionally, the lack of acceptance tests is preventing me from approving this version of the bugfix.
Let’s work on finding the optimal solution, as this is a critical middleware for the routing feature. I recommend spending more time refining the approach.

bhargav-polara:
@raman-m I had tried out [this ](https://github.com/ThreeMammals/Ocelot/pull/2351#issuecomment-3703775650) your suggestion but after applying other testcases are getting failed. I have refactored code and added more test Cases.

src/Ocelot/DownstreamUrlCreator/DownstreamUrlCreatorMiddleware.cs
```diff
@@ -126,27 +127,64 @@ protected static string MergeQueryStringsWithoutDuplicateValues(string queryStri
...
+            foreach (var kvp in parsedQuery)
+            {
+                var queryKey = kvp.Key;
+                var queryValues = kvp.Value;
+                
+                // Check if this query parameter matches any placeholder (case-insensitive)
+                if (placeholderLookup.TryGetValue(queryKey, out var placeholderValue))
+                {
+                    // Remove only the values that match the placeholder value exactly
+                    var filteredValues = queryValues.Where(qv => !string.Equals(qv, placeholderValue, StringComparison.Ordinal)).ToList();
```
RaynaldM:
```suggestion
                    var filteredValues = queryValues.Where(qv => !string.Equals(qv, placeholderValue, StringComparison.Ordinal)).ToArray();
```

RaynaldM (approved):

src/Ocelot/DownstreamUrlCreator/DownstreamUrlCreatorMiddleware.cs
raman-m: @RaynaldM 
You approved the PR even though commit https://github.com/ThreeMammals/Ocelot/pull/2351/commits/21661a4eec7a53e32704de33987a95ce42ee6d1a [tests failed](https://github.com/ThreeMammals/Ocelot/actions/runs/20706286947/job/59455855467), seriously?..
The author's refactoring in https://github.com/ThreeMammals/Ocelot/pull/2351/commits/21661a4eec7a53e32704de33987a95ce42ee6d1a wasn't necessary, to be fair. Instead of taking [my suggestions](https://github.com/ThreeMammals/Ocelot/pull/2351#pullrequestreview-3621897853) with recommendation to resolve the issues, he went for [aggressive algo refactoring](https://github.com/ThreeMammals/Ocelot/pull/2351/commits/21661a4eec7a53e32704de33987a95ce42ee6d1a#diff-080d008993a0eca45a728e6499876a947129ddb8ae8e3310422f2c0dfc19a7e8R129-R130) without checking how the tests ran locally.
### If you are okay with approving such a contribution, then I am at a loss for words.

src/Ocelot/DownstreamUrlCreator/DownstreamUrlCreatorMiddleware.cs
```diff
@@ -126,27 +127,64 @@ protected static string MergeQueryStringsWithoutDuplicateValues(string queryStri
...
+            var rawQuery = downstreamRequest.Query;
+
+            int firstQuestionMark = rawQuery.IndexOf('?');
+            if (firstQuestionMark != -1 &&
+                rawQuery.IndexOf('?', firstQuestionMark + 1) != -1)
             {
-                continue;
+                rawQuery =
+                    rawQuery.Substring(0, firstQuestionMark + 1) +
+                    rawQuery.Substring(firstQuestionMark + 1).Replace("?", "&");
```
raman-m: `// Normalize legacy malformed query (?a=1?b=2)` Wow!
Malformed query strings can’t occur in this middleware because `HttpContext.Request` is already initialized and the URL is well-formed, including the `DownstreamRequest.Query` property which is preinitialized by the following code:
https://github.com/ThreeMammals/Ocelot/blob/9fc4e78d314ad97d836b35c9b9bb188dba49d556/src/Ocelot/Request/Middleware/DownstreamRequest.cs#L20
The `_request` field is an `HttpRequestMessage` that's being created here:
https://github.com/ThreeMammals/Ocelot/blob/9fc4e78d314ad97d836b35c9b9bb188dba49d556/src/Ocelot/Request/Mapper/RequestMapper.cs#L15-L22
And we should focus our attention on this method:
https://github.com/ThreeMammals/Ocelot/blob/9fc4e78d314ad97d836b35c9b9bb188dba49d556/src/Ocelot/Request/Mapper/RequestMapper.cs#L76
Thus, `DownstreamRequest.Query` is simply encoded `HttpRequest.QueryString` and/or `HttpRequest.Query`.

An acceptance test should be created to confirm this is possible, and if the test is developed and proves it, the code block can stay.

src/Ocelot/DownstreamUrlCreator/DownstreamUrlCreatorMiddleware.cs
```diff
@@ -126,27 +127,64 @@ protected static string MergeQueryStringsWithoutDuplicateValues(string queryStri
...
 
-            int questionMarkOrAmpersand = downstreamRequest.Query.IndexOf(name, StringComparison.Ordinal);
-            builder.Clear()
-                .Append(downstreamRequest.Query)
-                .Replace(parameter, string.Empty)
-                .Remove(--questionMarkOrAmpersand, 1);
-            downstreamRequest.Query = builder.Length > 0
-                ? builder.Remove(0, 1).Insert(0, QuestionMark).ToString()
+            // Parse the NORMALIZED query
+            var parsedQuery = QueryHelpers.ParseQuery(rawQuery);
```
raman-m: Helpful method, but I am not sure we need to parse the string at all. We just need to check if the query parameter exists. I'll keep this helper in mind if it becomes necessary.

src/Ocelot/DownstreamUrlCreator/DownstreamUrlCreatorMiddleware.cs
```diff
@@ -126,27 +127,64 @@ protected static string MergeQueryStringsWithoutDuplicateValues(string queryStri
...
+            
+            // Create a case-insensitive lookup for template placeholders
+            var placeholderLookup = templatePlaceholders
+                .ToDictionary(p => p.Name.Trim(OpeningBrace, ClosingBrace), p => p.Value, StringComparer.OrdinalIgnoreCase);
```
raman-m: Consider refactoring `PlaceholderNameAndValue` instead of trimming the values. Also, why use `ToDictionary`? It adds unnecessary pressure on RAM and the garbage collector by creating a new object on the heap!

src/Ocelot/DownstreamUrlCreator/DownstreamUrlCreatorMiddleware.cs
```diff
@@ -126,27 +127,64 @@ protected static string MergeQueryStringsWithoutDuplicateValues(string queryStri
...
+            
+            // Build new query string by filtering out matching placeholders
+            var remainingParameters = new List<string>();
+            
+            foreach (var kvp in parsedQuery)
```
raman-m: This is an iteration loop through parameters, not the collection of placeholders.
The question is: **why?** 😄

src/Ocelot/DownstreamUrlCreator/DownstreamUrlCreatorMiddleware.cs
```diff
@@ -126,27 +127,64 @@ protected static string MergeQueryStringsWithoutDuplicateValues(string queryStri
...
+                        }
+                    }
+                }
+                else
+                {
+                    // This query parameter doesn't match any placeholder, keep all values
+                    foreach (var value in queryValues)
+                    {
+                        remainingParameters.Add($"{queryKey}={HttpUtility.UrlEncode(value)}");
+                    }
```
raman-m: If you already have a parsed version of the query, like `var parsedQuery = QueryHelpers.ParseQuery(rawQuery)`, why create a new collection? 🤨
Do you think a dictionary doesn’t have a remove method, like `parsedQuery.Remove(name)`? 🤣 That is hilarious!

raman-m (changes requested):
Yahoo! The rodeo goes on! 🤠 
Did you notice the CI build [failed](https://github.com/ThreeMammals/Ocelot/actions/runs/20706286947/job/59455855467)? No rush, though I'm planning to revert the last commit https://github.com/ThreeMammals/Ocelot/pull/2351/commits/21661a4eec7a53e32704de33987a95ce42ee6d1a.

You're a talented vibe-coder, and the new algorithm seems solid and will handle query strings properly, but instead of the old 10 lines, we now have 30 lines of code proposed by the AI coding agent. This version will heavily consume RAM due to two new collections in the algorithm. Honestly, I prefer the previous `Regex`-based version since it's better to use more CPU cycles than to increase RAM usage. Have you seen how expensive RAM modules are these days? 😉

> I had tried out https://github.com/ThreeMammals/Ocelot/pull/2351#issuecomment-3703775650your suggestion but after applying other testcases are getting failed. **I have refactored code and added more test Cases.**

You've added new unit tests generated by the AI coding agent but wrote no acceptance tests. And you didn't run the tests locally, so now the old Routing acceptance tests are failing 👇 
image

Sorry, the last commit will be reverted because the old `Regex` version is more suitable.

bhargav-polara:
@raman-m I have refactored the code based on your suggestions and also made a few additional changes to address the failing test cases that appeared after applying them.

raman-m:
It is great if the unit tests and old acceptance tests are now passing.
Consider adding new acceptance tests to the [RoutingWithQueryStringTests](https://github.com/ThreeMammals/Ocelot/blob/develop/test/Ocelot.AcceptanceTests/Routing/RoutingWithQueryStringTests.cs) class. Be sure to include the test case for #2346 to confirm the bug fix works properly.
Please note, I will perform the final code review after writing the acceptance tests.

bhargav-polara:
@raman-m Added new acceptance tests

raman-m:
### Ready for code review ✔️

raman-m (approved):
## Ready for delivery ✅ 
- Code review ✔️ 
- Unit testing ✔️ ✔️ 
- Acceptance testing ✔️ ✔️

PR #2358, opened 2026-02-10
#2354 Follow-up: Redesign `PollKube` discovery using `PeriodicTimer`

raman-m:
## Follow up #2354 
- #2354
- and commit https://github.com/ThreeMammals/Ocelot/commit/070593c0da9284d24782608d7f04380216d6f467

## Proposed Changes
- Added the `Dispose` pattern for logging and `Kube` polling
- Fixed [xUnit1051](https://xunit.net/xunit.analyzers/rules/xUnit1051) warnings, which is a best practice when writing xUnit tests after migrating to the xUnit v3 framework
- ❗Redesigned the `PollKube` discovery provider to utilize `PeriodicTimer`, which was introduced in .NET 8
- Temporarily disabled `PollKube` acceptance tests since the work is in progress

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -553,9 +553,9 @@ Task("UnitTests")
 		if (double.Parse(lineCoverage) < MinCodeCoverage)
 		{
 			var whereToCheck = !IsRunningInCICD() ? CoverallsRepo : artifactsForUnitTestsDir;
-			var msg = $"# Code coverage fell below the threshold of {MinCodeCoverage}%. You can find the code coverage report at {whereToCheck}";
+			var msg = $"# Code coverage fell below the threshold of {MinCodeCoverage * 100}%. You can find the code coverage report at {whereToCheck}";
 			Warning(msg);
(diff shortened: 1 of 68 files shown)


raman-m:
Hello @ggnaegi!
Please review if you have time, or at least take a look at this piece of my art 😄 I have an idea to reuse passive polling in the `PollKube` provider, as it was implemented by you in the `PollConsul` provider. I'm calling it "passive polling."

In this PR, I am working on implementing another polling strategy, aka "active polling," which is based on `PeriodicTimer` callbacks, replacing old `Timer` callbacks as proposed by Tom earlier. After the release, we could unify both polling strategies and propose a service discovery option to switch between them.

**P.S.** Please note that the `PollKube` acceptance tests have been temporarily disabled, as the provider and tests are still under development . However, **I plan to release the Ocelot beta version in 1-2 days**, which should be well-received by the community as it will facilitate the much-awaited migration to the .NET 10 SDK.

Issue #2208, opened 2024-11-20
Consul Node.Name might not be a DNS Name


ggnaegi:
## Expected Behavior / New Feature

The consul service discovery provider should check if the node.Name could be a DNS/Host Name, otherwise return the service url.

## Actual Behavior / Motivation for New Feature

The consul service provider doesn't check if the node.Name is a host name. Depending on service discovery configuration, the node.Name might not be a host name.

https://developer.hashicorp.com/consul/docs/agent

Node name: This is a unique name for the agent. By default, this is the hostname of the machine, but you may customize it using the [-node](https://developer.hashicorp.com/consul/docs/agent/config/cli-flags#_node) flag.

## Steps to Reproduce the Problem

  1. Change the agent name, use something that is not related to the hostname.
  1. Try to use the Consul Service Discovery

related to: https://github.com/ThreeMammals/Ocelot/issues/2203, https://github.com/ThreeMammals/Ocelot/issues/954

raman-m:
Adding an extra check for the DNS name could be a solution, providing some logic to protect Ocelot from careless DevOps engineers who might write whatever they want in the configs (in nodes). 😃 Gui, please note that an additional DNS probe for the node hostname will add 100+ milliseconds or even seconds to the request processing time, but we can cache the IP. But we could propose Boolean switcher of cache and entire DNS probe.

ggnaegi:
@raman-m you right, I think we should foresee an option for that matter

raman-m:
@ggnaegi 🆗 man... You've opened the issue, and... What are you going to do with this?

raman-m:
@ggnaegi What's happening here?

PR #2370, opened 2026-03-22
fix: prioritise service instance address over Consul node address in GetDownstreamHost

### LINKED ISSUE #2325: Gateway routes traffic to Consul address instead of service instance address in Ocelot 23.3+

TomPallister:
## Fixes #2325 
- #2325 

Regression introduced in Ocelot 23.3 where `GetDownstreamHost` incorrectly used the Consul node's `Name` as the downstream host whenever a node was present, breaking environments (especially Kubernetes) where Consul nodes and service instances run on different hosts.

## Changes

- **`DefaultConsulServiceBuilder.GetDownstreamHost`**: Reversed the priority logic to use `entry.Service.Address` first (the actual service instance address), falling back to `node.Address` then `node.Name` only when the service address is unavailable.
- **Unit tests**: Replaced the existing `GetDownstreamHost_BothBranches_NameOrAddress` test (which validated the broken behaviour) with two new tests covering all branches of the corrected logic.

### Before (broken, 23.3+)
```csharp
protected virtual string GetDownstreamHost(ServiceEntry entry, Node node)
    => node != null ? node.Name : entry.Service.Address;
```

### After (fixed)
```csharp
protected virtual string GetDownstreamHost(ServiceEntry entry, Node node)
    => !string.IsNullOrEmpty(entry?.Service?.Address)
        ? entry.Service.Address
        : node?.Address ?? node?.Name;
```

## Test plan

- [ ] `GetDownstreamHost_ServiceAddressPresent_ReturnsServiceAddress` — verifies service address is used even when a node is present
- [ ] `GetDownstreamHost_ServiceAddressEmpty_FallsBackToNodeAddressThenName` — verifies fallback to `node.Address` then `node.Name` when service address is empty/null
- [ ] All existing `DefaultConsulServiceBuilderTests` pass
- [ ] Full unit test suite passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Ocelot.Provider.Consul/DefaultConsulServiceBuilder.cs
+++ b/src/Ocelot.Provider.Consul/DefaultConsulServiceBuilder.cs
@@ -94,7 +94,9 @@ public class DefaultConsulServiceBuilder : IConsulServiceBuilder
             entry.Service.Port);
 
     protected virtual string GetDownstreamHost(ServiceEntry entry, Node node)
-        => node != null ? node.Name : entry.Service.Address;
+        => !string.IsNullOrEmpty(entry?.Service?.Address)
+            ? entry.Service.Address
+            : node?.Address ?? node?.Name;
 
     protected virtual string GetServiceId(ServiceEntry entry, Node node)
         => entry.Service.ID;
--- a/test/Ocelot.UnitTests/Consul/DefaultConsulServiceBuilderTests.cs
+++ b/test/Ocelot.UnitTests/Consul/DefaultConsulServiceBuilderTests.cs
@@ -114,23 +114,51 @@ public sealed class DefaultConsulServiceBuilderTests
     private static MethodInfo GetDownstreamHost { get; } = Me.GetMethod("GetDownstreamHost", BindingFlags.NonPublic | BindingFlags.Instance);
 
     [Fact]
-    public void GetDownstreamHost_BothBranches_NameOrAddress()
+    public void GetDownstreamHost_ServiceAddressPresent_ReturnsServiceAddress()
     {
+        // Issue #2325: service instance address should take priority over Consul node address
         Arrange();
 
-        // Arrange, Act, Assert: node branch
+        // Arrange, Act, Assert: service address is returned even when node is present
         ServiceEntry entry = new()
         {
-            Service = new() { Address = nameof(GetDownstreamHost_BothBranches_NameOrAddress) },
+            Service = new() { Address = "10.0.0.5" },
         };
-        var node = new Node { Name = "test1" };
+        var node = new Node { Name = "consul-node-1", Address = "192.168.1.1" };
         var actual = GetDownstreamHost.Invoke(sut, new object[] { entry, node }) as string;
-        actual.ShouldNotBeNull().ShouldBe("test1");
+        actual.ShouldNotBeNull().ShouldBe("10.0.0.5");
 
-        // Arrange, Act, Assert: entry branch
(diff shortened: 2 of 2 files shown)


raman-m:
> To: @TomPallister

Tom, nice to see you! 🙂

The problem is deeper than it may seem at first glance. Please see the related issue #2208.
If the author shows a clear intention to contribute, [I am happy to help](https://github.com/ThreeMammals/Ocelot/issues/2325#issuecomment-4108980805). However, any potential PR will likely involve more complex changes than a simple method redesign.
I believe we should focus our attention on issue #2208 created by @ggnaegi, but I am not sure when he plans to return to  the issue created by him.

> Cc: @ggnaegi

PR #2379, opened 2026-04-08
#2375 #2376 Map non-ASCII header `HttpRequestException` to 400 Bad Request according to RFC 7230

Majdi-Zlitni:
## Fixes #2376 
- #2376 

## Discussed in #2375 
- #2375 

### Changes

- Created a new `BadRequestError` mapped to **HTTP 400 (Bad Request)**.
- Updated `HttpExceptionToErrorMapper` to specifically catch `HttpRequestException` instances complaining about **"only ASCII characters"**.
- Mapped this specific exception to `BadRequestError` instead of falling back to the generic `ConnectionToDownstreamServiceError` (**HTTP 502**).

---

### Why it was done

When a client sends a request containing **invalid header characters** (for example an emoji 💀), Ocelot forwards the request as-is.  
While building the outbound request, the underlying .NET `HttpClient` throws the following exception:

Previously, Ocelot's error handler mapped this exception to **HTTP 502 Bad Gateway**, which incorrectly implies a failure in the downstream service.

By intercepting this specific exception (**Late Catch**), the gateway correctly identifies it as a **malformed client request** and returns the industry-standard **HTTP 400 Bad Request** instead.

---

### Architectural note

The solution uses a **Late Catch strategy** in the exception mapper rather than introducing early validation middleware.

This avoids the CPU overhead of iterating through and validating every incoming request header, preserving Ocelot’s philosophy of acting as a **transparent, high-performance proxy**.

PR #2382, opened 2026-04-17
Follow up #2379 Review InvalidHeaderValueTests

raman-m:
## Follow up #2379 
- #2379


raman-m (commented):
## Ready for delivery ✅

PR #2300, opened 2025-07-13
#941 Added SseDelegatingHandler

vijay-karavadra:
## Closes #941
- #941 

## Proposed Changes
  - Added `SseDelegatingHandler` to handle SSE event stream requests

## Discussions
- [server-sent events](https://github.com/ThreeMammals/Ocelot/discussions?discussions_q=server+sent+events)
- #1943

Issue #1942, opened 2024-01-22
在使用异步流返回时，并不是实时刷新逐条返回结果，而是一次返回结果


alivy:
## Steps to Reproduce the Problem

  1. 当响应体中有text/event-stream时，返回结果为一次性加载返回，而并非实时刷新
  2. 即使我加载了自己的Middleware添加到OcelotPipelineExtensions时，也不能实时刷新，以异步流的形式收到结果
自定义拓展代码如下

  
`

    public class CustomResponseMiddleware : OcelotMiddleware
    {
        private readonly RequestDelegate _next;
        public CustomResponseMiddleware(RequestDelegate next, IOcelotLoggerFactory loggerFactory)
            : base(loggerFactory.CreateLogger<CustomResponseMiddleware>())
        {
            _next = next;
        }

        public async Task Invoke(HttpContext httpContext)
        {
            var response = httpContext.Items.DownstreamResponse();
            const string streamingContent = "text/event-stream";
            if (httpContext.Response.Headers["Content-Type"] == streamingContent
                || httpContext.Response.ContentType == streamingContent)
            {
                //await ForwardStream(httpContext.Response.Body, response);
                var stream = await response.Content.ReadAsStreamAsync();
                await WriteServerStreamEventsAsync(httpContext, stream);
            }
            await _next.Invoke(httpContext);
        }
        private static async Task WriteServerStreamEventsAsync(HttpContext httpContext, Stream contentStream)
        {
            var responseStream = httpContext.Response.Body;
            await contentStream.CopyToAsync(responseStream, httpContext.RequestAborted);
            await responseStream.FlushAsync(httpContext.RequestAborted);
        }
 }

`

raman-m:
Dear @alivy !
Official language of the project is English. Most of our team mates and even contributors don't understand Chinese!

raman-m:
Please update the description using English texts, otherwise the issue will be closed!

ggnaegi:
When there is text/event-stream in the response body, the returned result is loaded and returned at once, not refreshed in real time.
Even when I load my own Middleware and add it to OcelotPipelineExtensions, it doesn't refresh in real time, and the results are received as an asynchronous stream.
The custom extension code is as follows

ggnaegi:
@alivy @raman-m  我不懂中文，所以可能无法用中文表达清楚。I would like to learn your language though. I think, you should wait the latest changes that were pushed to the develop branch. We are not copying the request body and response body directly anymore.

[...]

PR #2383, opened 2026-04-22
#941 #1775 #1942 #1943 Add Server-Sent Events (SSE) streaming support

### LINKED ISSUE #941: Server-sent events protocol
## Documentation
- [Server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events) - Web APIs | MDN
- [Server-sent events](https://learn.microsoft.com/en-us/azure/application-gateway/for-containers/server-sent-events) and Application Gateway for Containers | Microsoft Learn
- [Using Server-sent events with Application Gateway (Preview)](https://learn.microsoft.com/en-us/azure/application-gateway/use-server-sent-events) | Microsoft Learn
- [Configure API for server-sent events in Azure API Management](https://learn.microsoft.com/en-us/azure/api-management/how-to-server-sent-events) | Microsoft Learn
- [System.Net.ServerSentEvents Namespace](https://learn.microsoft.com/en-us/dotnet/api/system.net.serversentevents) | Microsoft Learn

## Search
- Microsoft Bing search: [server sent events](https://www.bing.com/search?q=server%20sent%20events%20site%3Amicrosoft.com)

pixellos:
## Closes #941 
- #941 

### Proposed Changes

This PR adds native support for the Server-Sent Events (SSE) protocol in Ocelot.

* **HttpContextResponder**: Automatically detects 	ext/event-stream response content types and explicitly disables buffering on the underlying Kestrel response via IHttpResponseBodyFeature.DisableBuffering(). This ensures that downstream chunks are immediately forwarded to the client without experiencing buffering latency.
* **Acceptance Tests**: Added a comprehensive suite of SignalR and SSE acceptance tests using the BDDfy framework to prove that SSE chunks bypass the typical buffering mechanisms. Also mapped POST and OPTIONS in the SignalR specific test to support /hub/negotiate handshakes.

### Verification
Tests pass locally successfully using .NET 8 SDK. The implementation strictly adheres to the architectural suggestions in the linked issue.

### Predecessor
- #2300 

### Discussion
- #1775 
- #1942 
- #1943


raman-m:
@pixellos Mateusz
You have uploaded inconsistent solution which can't be built ❗
- Build error: https://github.com/ThreeMammals/Ocelot/actions/runs/24784935679/job/72867502451?pr=2383#step:9:98

If you add new packages or change the versions of existing ones then you must update packages lock files. That is why the build is failing. I wonder that such experienced .NET developer like you did such a mistake.

What is your LinkedIn profile?

test/Ocelot.AcceptanceTests/packages.lock.json
raman-m, line 1: Your updating the lock file is wrong! You can't remove `net10.0` TFM since our current TFMs in testing are 8-9-10 ❗ 
https://github.com/ThreeMammals/Ocelot/blob/c3720c550899db4aadd77ae77754bd9978101c34/test/Ocelot.AcceptanceTests/Ocelot.AcceptanceTests.csproj#L4

raman-m (changes requested):
I hope next time your solution will be compilable and we will have successful CI build.

raman-m:
I said
### What is your LinkedIn profile?

pixellos:
Hi, thanks for the feedback—I've just pushed the corrected lockfiles. 

I don't usually work in repositories with multi-targeted lockfiles, so the missing framework graphs slipped right past me during my local restore. Definitely my mistake for not reviewing the lockfile diff more thoroughly 😐. 

Probably I should mark it as draft from the start - as it is my first contribution to the Ocelot, to gather feedback on things like lockfiles first.

pixellos:
@raman-m I'd rather not post my Linkedin profile there - I can send it to You directly thru some other communication channel.

raman-m:
> @raman-m I'd rather not post my Linkedin profile there - I can send it to You directly thru some other communication channel.

@pixellos
Well... You are trying to contribute to a public repo, but you don't want to be public person.

FYI, my LinkedIn is published in my GitHub profile. As a maintainer of this repo, I sometimes ask users to share their social profiles. In the era of AI, the goal of doing this is to ensure the user is a real person (not an AI robot) and not a student. As the owner of this product, I need to be sure the contribution and PR will be professional.

If you can't post your LinkedIn here, you can send your social network links to the email presented in the [ThreeMammals](https://github.com/ThreeMammals) org, where you will find the email. Thanks!

raman-m:
> * **Acceptance Tests**: Added a comprehensive suite of SignalR and SSE acceptance tests using the BDDfy framework to prove that SSE chunks bypass the typical buffering mechanisms. Also mapped POST and OPTIONS in the SignalR specific test to support /hub/negotiate handshakes.

You've fixed the problem with `test/Ocelot.AcceptanceTests/packages.lock.json` file in commit https://github.com/ThreeMammals/Ocelot/pull/2383/commits/845e60c3f68654b7876e7ba4b339d9ff3b6b52f6, but the build is still red! &rarr; https://github.com/ThreeMammals/Ocelot/actions/runs/24887525494/job/72872424180?pr=2383

The tests you wrote are failing. Thus, I would say either your solution is not complete, or the tests are written in the wrong way.
### Thanks for submitting this PR and linking it to the issue. However, this PR will remain in draft until we have a green build...
Please continue development... and I will press the CI button to make builds.

pixellos:
Sure, I do already have fix locally, 

When I was running test on .net 10 and 8 there were small difference. 

Sorry for the inconvinience - I'll make sure my next push will have everything in order. 
Also, I've just sent the email. Cheers.

pixellos:
@raman-m - do we need to add documentation? I was thinking about new page.

Also, added tests as per contribution guide.

Also, I spotted that Polly were removed from lock, It is not part of my changeset, is that expected? I'm not that aware of this codebase to confirm or deny it.
```
Server-Sent Events (SSE)
========================

Ocelot supports proxying `Server-Sent Events (SSE) <https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events>`_ in real-time. 

When Ocelot detects a response with the ``text/event-stream`` content type, it automatically disables response buffering. This ensures that event chunks are forwarded to the client immediately as they are received from the downstream service, rather than being buffered by the gateway.

SignalR over SSE
----------------

This feature also enables support for `ASP.NET Core SignalR <https://learn.microsoft.com/en-us/aspnet/core/signalr/introduction>`_ when using the Server-Sent Events transport.

Configuration
-------------

No special configuration is required to enable SSE support. Ocelot handles the buffering control automatically based on the ``Content-Type`` header of the downstream response.

However, ensure that your routes are configured to allow the ``GET`` method, as SSE typically uses standard HTTP GET requests with a persistent connection.

.. code-block:: json

  {
    "UpstreamPathTemplate": "/events",
    "DownstreamPathTemplate": "/api/events",
    "DownstreamScheme": "http",
    "DownstreamHostAndPorts": [
      { "Host": "localhost", "Port": 5001 }
    ],
    "UpstreamHttpMethod": [ "Get" ]
  }

Notes
-----

* **Timeouts**: Since SSE connections are long-lived, ensure that any proxy or load balancer timeouts (including Ocelot's QoS or Request Timeout settings) are configured to accommodate the expected duration of your event streams.
* **Buffering**: Ocelot uses the native ASP.NET Core ``IHttpResponseBodyFeature.DisableBuffering()`` to control the response flow.

```

src/Ocelot/Responder/HttpContextResponder.cs
```diff
@@ -77,6 +77,13 @@ public async Task SetErrorResponseOnContext(HttpContext context, DownstreamRespo
 
     protected virtual async Task WriteToUpstreamAsync(HttpContext context, DownstreamResponse downstream)
     {
+        var isSse = downstream.Content?.Headers?.ContentType?.MediaType?.Equals("text/event-stream", StringComparison.OrdinalIgnoreCase) == true;
+        if (isSse)
+        {
+            var httpResponseBodyFeature = context.Features.Get<IHttpResponseBodyFeature>();
+            httpResponseBodyFeature?.DisableBuffering();
```
raman-m: Why do you null-check here?
I guess `IHttpResponseBodyFeature` is always added to the DI container due to Kestrel's web server functionality, right?

test/Ocelot.AcceptanceTests/Ocelot.AcceptanceTests.csproj
```diff
@@ -62,6 +62,7 @@
     <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="10.0.5" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="18.3.0" />
     <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="8.16.0" />
+    <PackageReference Include="Microsoft.AspNetCore.SignalR.Client" Version="8.0.5" />
```
raman-m: Please point my attention to the exact line in the tests that use this lib.

I'm worried that we might either create a separate extension package or reference the lib directly in the Ocelot project. What is your opinion?

test/Ocelot.AcceptanceTests/ServerSentEventsTests.cs
```diff
@@ -0,0 +1,272 @@
...
+
+namespace Ocelot.AcceptanceTests.ServerSentEvents;
+
+public class ServerSentEventsTests : Steps
+{
+    private readonly List<string> _receivedEvents = new();
+    private readonly Stopwatch _stopwatch = new();
+    
+    [Fact]
+    [Trait("Feat", "SSE")]
```
raman-m: 😄 We usually point to the issue number.

test/Ocelot.AcceptanceTests/ServerSentEventsTests.cs
```diff
@@ -0,0 +1,272 @@
...
+                if (!string.IsNullOrEmpty(line))
+                {
+                    _receivedEvents.Add(line);
+                    
+                    if (_receivedEvents.Count == 1)
+                    {
+                        _stopwatch.Stop();
+                    }
+                }
+            }
```
raman-m, line 98: How many chunks / events does this code send?
2 events, right?

raman-m (changes requested):
Thanks for writing excellent acceptance tests 😍

Since you test SignalR connection based on `sse` proto, I would recommend writing browser tests (a.k.a. JavaScript SignalR client) that connect to the Ocelot endpoint.
We don't have a dedicated project for browser tests (Selenium Webdriver, etc.), but the test can be added to the [Ocelot.ManualTest](https://github.com/ThreeMammals/Ocelot/tree/develop/test/Ocelot.ManualTest) project &rarr; [test/Ocelot.ManualTest/Tests](https://github.com/ThreeMammals/Ocelot/tree/develop/test/Ocelot.ManualTest/Tests)

Also, I believe that code coverage could be improved. The `HttpContextResponder` class has a risky covered [line 80](https://app.codecov.io/gh/ThreeMammals/Ocelot/pull/2383#240f43869792679b7c3d6894831da7b2-R80).

Also, see my other suggestions below 👇

raman-m:
@pixellos commented [on April 24](https://github.com/ThreeMammals/Ocelot/pull/2383#issuecomment-4313942170)
> do we need to add documentation? I was thinking about new page.

Yes, we do. But docs must be written after reaching "Dev Complete" status, and I write docs before/after approving a PR.
Let's return to the docs later...

pixellos:
Hi, will get to those questions later today, as im not near my pc, Thanks
for review!

sob., 25 kwi 2026, 12:46 użytkownik Raman Maksimchuk <
***@***.***> napisał:

raman-m:
Yeah, you could start on Monday or later.
Enjoy your weekend!

test/Ocelot.AcceptanceTests/ServerSentEventsTests.cs
pixellos: Hi, thanks for the review! 

Yes, 2 - the idea was, to check whether the first message is delivered before stream ends - thus check in line 104 for first and second event. 

If the SSE would be buffered, both events in code

```C#
await context.Response.WriteAsync("data: event1\n\n");
await context.Response.Body.FlushAsync();

// Wait deliberately to ensure chunks are sent asynchronously
await Task.Delay(1000);

await context.Response.WriteAsync("data: event2\n\n");
await context.Response.Body.FlushAsync();
```
Would arrive at the same time, after finish of whole handler. This test aim to check streaming behavior.

Maybe I overcomplicated it a bit - Will think about how to make it more straightforward.

src/Ocelot/Responder/HttpContextResponder.cs
pixellos: I kept null check as defensive measure to ensure it won't crash the app if - for some reason - it would be null. I can remove it, but maybe i can just put it along `if (isSse)` check

test/Ocelot.AcceptanceTests/ServerSentEventsTests.cs
pixellos: Thanks, will adjust.

test/Ocelot.AcceptanceTests/Ocelot.AcceptanceTests.csproj
pixellos: image
` var connection = new HubConnectionBuilder()` is used for creating client using this lib in tests, for testing SSE mode client/server behavior, 

> I'm worried that we might either create a separate extension package or reference the lib directly in the Ocelot project. What is your opinion?

Maybe I'm missing something - the SignalR client was just a example. 
We could build custom SSE server which Ocelot, after this change, will proxy just fine.
It is not necessary to include SignalR directly.

pixellos:
> Thanks for writing excellent acceptance tests 😍
> 
> Since you test SignalR connection based on `sse` proto, I would recommend writing browser tests (a.k.a. JavaScript SignalR client) that connect to the Ocelot endpoint. We don't have a dedicated project for browser tests (Selenium Webdriver, etc.), but the test can be added to the [Ocelot.ManualTest](https://github.com/ThreeMammals/Ocelot/tree/develop/test/Ocelot.ManualTest) project → [test/Ocelot.ManualTest/Tests](https://github.com/ThreeMammals/Ocelot/tree/develop/test/Ocelot.ManualTest/Tests)
> 
> Also, I believe that code coverage could be improved. The `HttpContextResponder` class has a risky covered [line 80](https://app.codecov.io/gh/ThreeMammals/Ocelot/pull/2383#240f43869792679b7c3d6894831da7b2-R80).
> 
> Also, see my other suggestions below 👇

Thanks for all the suggestions — I’ll go through them.
For the JS tests, I’d propose a Playwright‑based setup. It’s much more “batteries included”. I’ll prepare a draft in the next few days.

pixellos:
Hi, when I did negative tests - that without my change the events are indeed delayed I found something weird - tests weren't catching buffering issues because a basic Kestrel setup flushes immediately. 

In real-world scenarios (like Kubernetes with Nginx or using ResponseCompressionMiddleware), SSE events were being buffered and delayed. I reproduced this by adding compression to our test gateway, which immediately broke real-time streaming in the tests.

Also, when I was testing there was another problem next to compression - the Nginx were buffering the events as well :/. 
I added a header if the SSE is used. Please let me know whether that solution can stay, or need I to create separate PR for that.

Also, after some manual tests, sometimes the Content-Type is not exact, thus was added a test case and fix for it.

Also, added a preview of Playwright tests using browser and SignalR - I'm not used to the Cake, so please let me know if we shall do it like that, or You have different idea, I'll fix the linux setup if needed, as now I just wanted to showcase the solution.

testing/PortFinder.cs
```diff
@@ -22,7 +23,7 @@ public static int GetRandomPort()
         lock (SyncRoot)
         {
             ExceedingPortRangeException.ThrowIf(CurrentPort > EndPortRange);
-            while (!TryUsePort(++CurrentPort));
+            while (!TryUsePort(++CurrentPort)) ;
```
pixellos: Will remove space in next commit

build.cake
```diff
@@ -579,6 +583,20 @@ Task("AcceptanceTests")
...
+        var playwrightPath = $"./test/Ocelot.AcceptanceTests/bin/{compileConfig}/{latestTfm}/playwright.ps1";
+        
+        Information("Installing Playwright browsers...");
+        if (IsRunningOnWindows())
+        {
+            StartProcess("powershell", new ProcessSettings { Arguments = $"-File {playwrightPath} install chromium" });
+        }
+        else
+        {
+            Information("Skipping Playwright installation on non-Windows (TODO: support Linux/macOS in CI)");
```
pixellos: TBD: Implement after confirmation we'd go with playwright

test/Ocelot.AcceptanceTests/ServerSentEventsTests.cs
raman-m: Let's leave it as-is until I start the final code review after the current .NET 10 release, which is delayed, as usual 🤕

src/Ocelot/Responder/HttpContextResponder.cs
raman-m: Very odd if it can be null. Could you confirm that somehow in the tests, please?
If that's true, I would recommend adding an [OcelotBuilder](https://github.com/search?q=repo%3AThreeMammals%2FOcelot%20OcelotBuilder&type=code) dependency injection helper that will enable the feature 100%, thus ensuring this line will never generate an exception. Then we can be confident in removing this extra null-check.
Finally, I expect that the `if`-statement at line 81 might be extended.

test/Ocelot.AcceptanceTests/Ocelot.AcceptanceTests.csproj
raman-m: Thus, you are sure we don't need extra SignalR support on Ocelot's side, right?
Anyway, acceptance testing will show the whole picture for me. Sorry, I haven't worked closely with SignalR streaming in .NET apps.

src/Ocelot/Responder/HttpContextResponder.cs
pixellos: In Kestrel it could never be null, but there can be custom Web Framework that will have different set of features - I tested on IIS In-Process and on Kestrel - it is on both

> I would recommend adding an [OcelotBuilder](https://github.com/search?q=repo%3AThreeMammals%2FOcelot%20OcelotBuilder&type=code) dependency injection helper that will enable the feature 100%, thus ensuring this line will never generate an exception.

Maybe I haven't caught the idea, but if HostingModule is not supporting that feautre, how we can pollyfil it? By adding No-Op implementation? 

As in Kestrel and in IIS In/Out process this is present, so in normal case it won't ever be null.

test/Ocelot.AcceptanceTests/Ocelot.AcceptanceTests.csproj
pixellos: SignalR negotiates first one of three protocols, WebSockets (runs in ocelot just fine), SSE (this PR fixes it when the Response Compression is enabled), then LongPolling (there is small cavetat regarding this and QoS, but it is another topic)

It's why we don't need SignalR standalone, as it is using known transports, it's not adding anything custom

pixellos: Image
And there i'm limiting it it SSE to not fall-back to other methods

testing/PortFinder.cs
raman-m: Indeed, dude! Revert it! 😄

build.cake
raman-m: Can't get it! 🙄
Why are you updating the [build.cake](https://github.com/ThreeMammals/Ocelot/pull/2383/files/392434c77e58651784b566cee9c2481d25cdedf7#diff-c6eace5a70f0c2d920dfa97e988d4cad3d0131dd2b029bb2646c96222e866cf8) script? It is prepared for CI-CD only...

My understanding is that you're going to change the Manual Test project to perform browser tests... Or you just want to add a special target to `build.cake` to run browser tests from CI-CD?

If Playwright tests can't be run on Linux/MacOS, I don't see much sense in CI-CD testing because we must run them using Windows instance, right? What we need now is converting the Manual Testing app into a console app that can run in different modes, including testing mode. That will be the goal. After that, we could embed the redeveloped console app into the CI-CD environment.

pixellos: Okay, I missunderstood You. I can move them to the Manual tests.

They can be run on Linux/MacOS - but first i wanted to get feedback on the tests itself before adding more setup, as I'm developing on windows.

So as part of this PR what You'd suggest? I'm fine with both adding as ManualTest and having special browser Cake targets, I just need to know which direction @raman-m You would prefer.

raman-m: I don't require you to embed it to CI-CD. Let's focus on ManualTest project first which is console app.

> They can be run on Linux/MacOS - but first i wanted to get feedback on the tests itself before adding more setup, as I'm developing on windows.

Latest CI build &rarr; [build 616](https://github.com/ThreeMammals/Ocelot/actions/runs/24958610737?pr=2383)
Acceptance tests have failed due to [Playwright tests failing](https://github.com/ThreeMammals/Ocelot/actions/runs/24958610737/job/73376787602?pr=2383#step:9:322).  

A standard Ubuntu job runs an image with the Ubuntu OS with .NET SDK preinstalled, but without the Node.js framework. I'm afraid CI-CD requires running a special Docker image in a job with Playwright preinstalled. That will not be easy to set up in CI-CD right now. Let's return to this problem later when the feature has been developed.

The build-cake job runs on an Ubuntu Linux machine for pull requests. You could enable a Windows job temporarily for development if it's needed.
https://github.com/ThreeMammals/Ocelot/blob/c3720c550899db4aadd77ae77754bd9978101c34/.github/workflows/pr.yml#L198

raman-m: My idea is to run the ManualTest app in a local environment, which is a Windows 11 machine. If the browser tests pass, then I will make a release or roll out a patch.

However, running browser tests in CI-CD will be another task that is beyond the scope of the current issue.
Ideally, for browser testing we could have a separate job running Windows, but let's skip this for now, and concentrate on feat development.

pixellos: > Acceptance tests have failed due to [Playwright tests failing](https://github.com/ThreeMammals/Ocelot/actions/runs/24958610737/job/73376787602?pr=2383#step:9:322).

Yes, they failed, as I haven't added linux setup yet  - I'll try later on today to spin Ubuntu VM and develop the linux cake there just to showcase the idea, then will move it into the ManualTesting.

raman-m:
> @pixellos commented [on April 26](https://github.com/ThreeMammals/Ocelot/pull/2383#issuecomment-4322202684)

Wow! This is more complex than I thought.

> Hi, when I did negative tests - that without my change the events are indeed delayed I found something weird - tests weren't catching buffering issues because a basic Kestrel setup flushes immediately.
> 
> In real-world scenarios (like Kubernetes with Nginx or using ResponseCompressionMiddleware), SSE events were being buffered and delayed. I reproduced this by adding compression to our test gateway, which immediately broke real-time streaming in the tests.
> 
> Also, when I was testing there was another problem next to compression - the Nginx were buffering the events as well :/.
> I added a header if the SSE is used. **Please let me know whether that solution can stay, or need I to create separate PR for that.**
> 

The solution could be a dedicated [OcelotBuilder](https://github.com/search?q=repo%3AThreeMammals%2FOcelot+OcelotBuilder&type=code) helper that will add the required services to the DI container.
I'm not sure about adding it in this PR or later as a follow-up task.

But if the Ocelot SSE feature will be broken when the user enables the `ResponseCompressionMiddleware`, we must highlight this gotcha in the docs in the current PR. Later, we could develop a dedicated [OcelotBuilder](https://github.com/search?q=repo%3AThreeMammals%2FOcelot+OcelotBuilder&type=code) method to add the missing services.

> Also, after some manual tests, sometimes the `Content-Type` is not exact, thus was added a test case and fix for it.

The potential fix must be a part of this PR. If the header is required for successful SSE streaming, then Ocelot must handle this header correctly.

>Also, added a preview of Playwright tests using browser and SignalR - I'm not used to the Cake, so please let me know if we shall do it like that, or You have different idea, I'll fix the linux setup if needed, as now I just wanted to showcase the solution.

We have already discussed this problem, see above.
**Developed feature and SignalR support must work on Windows OS first.** It should be tested in ManualTest project having all browsers tests successfully passing on Windows. Later we will update the solution to test the feat on Linux/MacOS and run Playwright tests aka all browser tests in CI-CD, but that will be follow up task which you could start after merging this PR.

build.cake
raman-m: > Ubuntu VM

😂 Why? Do you want to play with Docker containers locally (WSL, etc.)? You'll be reinventing the wheel.
Just enable jobs for the [PR workflow](https://github.com/ThreeMammals/Ocelot/blob/develop/.github/workflows/pr.yml) and test whatever you want on any OS.

If you don't have Ubuntu OS installed on your workstation, maybe it's time to install Ubuntu?
I sometimes boot into Ubuntu with .NET SDKs preinstalled, pull the latest Ocelot code, and check something.
Also, you could run the ManualTest project as a terminal app, but you must boot into Ubuntu OS.

Finally, having an Ubuntu virtual machine locally is an acceptable solution. However, installing Ubuntu on a separate disk partition is a more reliable solution because booting is fast on modern SSDs.

test/Ocelot.AcceptanceTests/Ocelot.AcceptanceTests.csproj
```diff
@@ -9,7 +9,7 @@
     <AssemblyName>Ocelot.AcceptanceTests</AssemblyName>
     <OutputType>Exe</OutputType>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
-    <RuntimeIdentifiers>win-x64;osx-x64</RuntimeIdentifiers>
+    <RuntimeIdentifiers>win-x64;osx-x64;linux-x64</RuntimeIdentifiers>
```
pixellos, line 12: I needed to add it to have successful test run on WSL

pixellos:
I see that pipeline failed - probably I don't understand how Cake is wired there, as in logs I haven't seen `Installing Playwright browsers.` info from my changes

Anywa, tomorrow I'd do as You suggested - move this Playwright code to ManualTesting project, to not stale on this feature.

test/Ocelot.AcceptanceTests/Ocelot.AcceptanceTests.csproj
raman-m: 🆗

raman-m:
image

This Playwright app error in the frame confirms that we have to use a dedicated Docker image for browser tests with preinstalled Playwright, etc. Thus, automated testing is possible in GitHub Action as a separate job that deploys a special Docker container.
This confirms that locating the browser tests in the acceptance testing project might not be a good idea, at least for the first time.
If these tests are part of the Manual Testing project, we will run them locally, having all Playwright software preinstalled.


**P.S.** When I asked you to create a dedicated webpage for SSE proto testing, my idea was to have JavaScript scripts and a client performing a test from this webpage. The Manual Testing project is an Ocelot instance that starts as a console app. After that, we need to manually open a webpage in a browser and run the tests connecting to the local Ocelot instance. This is the general idea of this console app.
See example in [test/Ocelot.ManualTest/Tests/Bug0930.html](https://github.com/ThreeMammals/Ocelot/blob/develop/test/Ocelot.ManualTest/Tests/Bug0930.html) which performs WebSocket tests aka `ws` proto testing.

raman-m:
> I see that pipeline failed - probably I don't understand how Cake is wired there, as in logs I haven't seen `Installing Playwright browsers.` info from my changes
> 

image
I will help you to resolve this error later.

pixellos:
https://github.com/ThreeMammals/Ocelot/pull/2383#issuecomment-4352838986

Thanks for explaination, today I'm kinda busy, won't be able to finish manual testing, maybe today or on Sunday, as I'd have weekend with drill :)

> I will help you to resolve this error later.

I missed that somehow when I was looking at logs, so maybe it is good news, as we need only to restore tools before that command, and Playwright would succeeded with installing of browsers before Acceptance Test, and we'd have it automated already.

raman-m:
Please drill down more! After returning to this PR, please resolve the merge conflicts, since merging from develop is required.

pixellos:
Hi, I had a quick look on problem You highlighted in https://github.com/ThreeMammals/Ocelot/pull/2383#issuecomment-4352889607 - I see that for other tools we have at top of the file the declaration, 

I removed tools from my system, wsl chrome playwright binaries, and re-tested. The acceptance criteria should be fine this time - this should allow us to run it without any docker playwright container.

Conflicts resolved, if we can check on CI whether it will pass the tests, I'll try in meantime push separate branch with manual tests, when this restore fix won't work, I'll bring the manual test changes into this PR.

pixellos:
I did the manual testing showcase + branch
https://github.com/user-attachments/assets/92ec8fbd-b494-4cfd-b339-8ae87053ebaa

https://github.com/ThreeMammals/Ocelot/compare/develop...pixellos:Ocelot:feature/sse-manual-test

I can replace it when needed, but I'd like to first run GH Action on last commit - as the playwright should install fine on the agent.

Also I got an idea - I can run workflows under my Ocelot repo to test them:

image
And it seems latest changes are succeeding with installing PlayWright and running tests

src/Ocelot/Responder/HttpContextResponder.cs
raman-m: > **In Kestrel it could never be null**, but there can be custom Web Framework that will have different set of features - I tested on IIS In-Process and on Kestrel - **it is on both**

Why do you null-check? 
As far as I know, this feature is available for all web servers like Kestrel, HTTP.sys, and IIS. It simply cannot be null. But if it is null, the overall SSE streaming will fail, right?

Therefore, I recommend logging a warning or even an error if the feature object is null. We should include information such as the web server type, platform (aka OS), and .NET TFM (aka SDK version).

Finally, if the request is an SSE request with a `text/event-stream` content type, but the feature is null, then we must log an error because the final streaming will likely fail without the feature. So, better to warn.

> Maybe I haven't caught the idea, but if HostingModule is not supporting that feautre, how we can pollyfil it? By adding No-Op implementation?

We will not pollyfil anything because we are not web server authors 🤣 
Just log an error please!

test/Ocelot.AcceptanceTests/Ocelot.AcceptanceTests.csproj
raman-m: OK, got it! Since these will be automated tests, they will connect from a .NET client, which requires this extra SignalR package. True browser testing from a JavaScript client will be performed in the Manual Testing project. Resolving this thread...

raman-m:
image

Dear Pan Mateusz,

I don't understand why you don't commit with your GitHub-registered user @pixellos.
Is "Test User" an AI agent? What AI coding agents do you use? The problem is that the commits are not GPG-signed, and this is bad. FYI, all Copilot AI agent users are registered and have valid signatures. You can see an example of them in the latest commit in the develop branch.

Please make the username more friendly or start committing as your GH user.
Also, please commit with the @pixellos user because I request changes after code reviews personally from you, not from an AI coding agent. **Don't delegate all the work to AI, plz, since this is not productive and is risky.**
Thanks!

raman-m:
As you noted, I've started setting up automated browser tests in CI/CD (see my latest commits). Build [632](https://github.com/ThreeMammals/Ocelot/actions/runs/25259231939) is OK on Ubuntu but fails on macOS — likely because Playwright is poorly prepared for macOS.

Also, note that [Microsoft.Playwright.CLI](https://www.nuget.org/packages/Microsoft.Playwright.CLI) is deprecated (latest v1.2.3 from March 2022). We should replace it with the [Microsoft.Playwright](https://www.nuget.org/packages/Microsoft.Playwright) package ([v1.59.0](https://www.nuget.org/packages/Microsoft.Playwright/1.59.0) as of April 2026). [Official docs](https://playwright.dev/dotnet/) recommend installing via PowerShell script &rarr; [Installation Introduction](https://playwright.dev/dotnet/docs/intro#introduction)

That's why I added the ["Install Playwright"](https://github.com/pixellos/Ocelot/blob/feature/sse-support/.github/workflows/pr.yml#L197) step in the PR workflow. I'll continue working on the macOS issue next week and will do a thorough review after the first green build on all platforms.

> P.S. In the future, we need a separate job due to the difficulty of installing all browsers compared to a single Chromium browser. Ultimately, it is better to use a Docker container.

pixellos:
Hi, this weekend was a little short on time - I checked the containerized setup - it is only possible for tests on Linux as documentation says. Looking on doc I see we're missing one Csproj property   `<PlaywrightPlatform>all</PlaywrightPlatform>` - I'll give it a try.

About Test User - I hadn't had really well configured Git on my personal machine, will change it to github user/email.

About AI - yes, I'm using Google Gemini with Antigravity, but locally, not using agent.

Also, I succesfully run the changes on my repo actions, and it seems it failed on MacOS, but with same error as develop is, I don't see full CI test log, but as I can't see PW tests errors, I'd assume they're fine :)

src/Ocelot/Responder/HttpContextResponder.cs
pixellos: Hi, removed null check and branched out to the warning.

raman-m:
@pixellos Is the "Re-run job" button enabled for you?
Are you able to re-run the PR workflow?

pixellos:
I was running it on my https://github.com/pixellos/Ocelot/actions/runs/25304145349/job/74176505532 fork


On this repo I can't see any actionable buttons 
image

raman-m:
OK, anyway...
Did you see yesterday's successful build?
image
The CI setup has finished! Your commit https://github.com/ThreeMammals/Ocelot/pull/2383/commits/d8babdf3197b997384e7ff5aef897961f3ca44db was right.
And stop manipulating workflows! This is not your responsibility. Please focus on the feature and resolving issues of code reviews.
Thanks!

raman-m:
The CI build [PR 636](https://github.com/ThreeMammals/Ocelot/actions/runs/25304146498) for commit https://github.com/ThreeMammals/Ocelot/pull/2383/commits/920505e605f24eb06832729e9dea96e9c2a3b74d has turned green ✅ Congrats! Overall status failed due to coverage report. But [PR workflow](https://github.com/ThreeMammals/Ocelot/actions/workflows/pr.yml) run was successful.

src/Ocelot/Responder/HttpContextResponder.cs
raman-m: [I see](https://github.com/ThreeMammals/Ocelot/pull/2383/changes#diff-071902289f34bf33ffaaa53f16bf70119f20e53553b55185da173fb333159bee), thanks!

src/Ocelot/Responder/HttpContextResponder.cs
```diff
@@ -77,6 +82,29 @@ public async Task SetErrorResponseOnContext(HttpContext context, DownstreamRespo
 
     protected virtual async Task WriteToUpstreamAsync(HttpContext context, DownstreamResponse downstream)
     {
+        var mediaType = downstream.Content?.Headers?.ContentType?.MediaType;
+        var isSse = mediaType?.Equals("text/event-stream", StringComparison.OrdinalIgnoreCase) == true
+            || (mediaType?.StartsWith("text/event-stream", StringComparison.OrdinalIgnoreCase) == true && mediaType.Contains(';'));
```
raman-m: You abused the conditional null-check operator aka `?.`❗
**And the overall [coverage report fails](https://app.codecov.io/gh/ThreeMammals/Ocelot/pull/2383) due to uncovered logical branches.**
Look at lines 47 and 52 where there is the `downstream.Content.Headers` expression.
I don't think it can be null since this is a critical header of the HTTP protocol. In some scenarios, `MediaType` can be null. But we could substitute an empty value header.

```suggestion
        var mediaType = downstream.Content.Headers.ContentType?.MediaType ?? string.Empty;
        var isSse = mediaType.Equals("text/event-stream", StringComparison.OrdinalIgnoreCase)
            || (mediaType.StartsWith("text/event-stream", StringComparison.OrdinalIgnoreCase) && mediaType.Contains(';'));
```
If this suggestion is applied, then the code coverage report will turn green, I hope.

src/Ocelot/Responder/HttpContextResponder.cs
```diff
@@ -77,6 +82,29 @@ public async Task SetErrorResponseOnContext(HttpContext context, DownstreamRespo
...
+        {
+            var feature = context.Features.Get<IHttpResponseBodyFeature>();
+            if (feature != null)
+            {
+                feature.DisableBuffering();
+            }
+            else
+            {
+                var server = context.RequestServices?.GetService(typeof(IServer))?.GetType().Name ?? "Unknown";
+                _logger.LogWarning($"IHttpResponseBodyFeature is null for SSE request. Buffering cannot be disabled. Server: {server}, OS: {RuntimeInformation.OSDescription}, Framework: {RuntimeInformation.FrameworkDescription}");
```
raman-m: Similarly, you abused `?.` usage, and the code [coverage report](https://app.codecov.io/gh/ThreeMammals/Ocelot/pull/2383#240f43869792679b7c3d6894831da7b2-R98) highlighted this issue.

As far as I know, the `IServer` service is always available in the DI container due to the hosting web server. So, it cannot be null in acceptance tests, but it can be null in unit tests.
We can apply a workaround for the `Type` object. The `RequestServices` prop is always initialized in the context.
Also, I recommend falling back from the `Type` object to `Object` to use `ToString()`.

```suggestion
                object server = context.RequestServices.GetService(typeof(IServer)) ?? "Unknown";
                string serverInfo = server.ToString();
                _logger.LogWarning(() => $"{nameof(IHttpResponseBodyFeature)} is null for SSE request. Buffering cannot be disabled. Server: {serverInfo}, OS: {RuntimeInformation.OSDescription}, Framework: {RuntimeInformation.FrameworkDescription}");
```

src/Ocelot/Responder/HttpContextResponder.cs
```diff
@@ -77,6 +82,29 @@ public async Task SetErrorResponseOnContext(HttpContext context, DownstreamRespo
...
+            }
+
+            if (!context.Response.Headers.ContainsKey("X-Accel-Buffering"))
+            {
+                context.Response.Headers.Append("X-Accel-Buffering", "no");
+            }
```
raman-m: Looks like this is a workaround for some software.
The [dotnet repos do not have this header](https://github.com/search?q=org%3Adotnet%20X-Accel-Buffering&type=code) defined in the source code. Thus, this header is not part of the [.NET Runtime](https://github.com/search?q=repo%3Adotnet%2Fruntime%20X-Accel-Buffering&type=code) or [ASP.NET](https://github.com/search?q=repo%3Adotnet%2Faspnetcore%20X-Accel-Buffering&type=code).

I have googled this header, and it seems this is an [NGINX environment header](https://nginx.org/en/docs/http/ngx_http_proxy_module.html). It looks like your Ocelot instance is hosted in a special NGINX environment or has it in the proxying chain, and you are trying to fix your environment.
I am OK with fixing it for a special environment such as NGINX, but we have to rework this block. It seems it should be a `protected virtual` collection or method.

```suggestion
            ProcessSoftwareHeadersForServerSentEvents(context.Response.Headers);
```

New methods:
```csharp
    protected virtual void ProcessSoftwareHeadersForServerSentEvents(IHeaderDictionary headers)
    {
        SetupNginxSseHeaders(headers);
    }

    protected virtual void SetupNginxSseHeaders(IHeaderDictionary headers)
    {
        if (!headers.ContainsKey("X-Accel-Buffering"))
        {
            headers.Append("X-Accel-Buffering", "no");
        }
    }
```

.config/dotnet-tools.json
```diff
@@ -14,6 +14,13 @@
         "reportgenerator"
       ],
       "rollForward": false
+    },
+    "microsoft.playwright.cli": {
```
raman-m: Are you sure we still need this tool? It's a [deprecated package](https://www.nuget.org/packages/Microsoft.Playwright.CLI)❗
We install tooling via the script, right? The `Microsoft.Playwright` package is already referenced in the acceptance testing project.

testing/PortFinder.cs
```diff
@@ -7,7 +7,8 @@ namespace Ocelot.Testing;
 public static class PortFinder
 {
     private const int EndPortRange = 45000;
-    private static volatile int CurrentPort = 20000;
+    // Adding process-specific offset to avoid port clashing during parallel multi-target test runs
+    private static volatile int CurrentPort = 20000 + (Environment.ProcessId % 500) * 20;
```
raman-m: Interesting lifehack...
Do we allocate 10 000 ports per process? And how many processes do we have?

build.cake
```diff
@@ -1,5 +1,6 @@
-﻿#tool dotnet:?package=GitVersion.Tool&version=6.6.2
+#tool dotnet:?package=GitVersion.Tool&version=6.6.2
 #tool nuget:?package=ReportGenerator&version=5.5.4
+#tool dotnet:?package=Microsoft.Playwright.CLI&version=1.2.3
```
raman-m: The [Microsoft.Playwright.CLI](https://www.nuget.org/packages/Microsoft.Playwright.CLI) package is deprecated!

raman-m (changes requested):
This is the first round of code review. I expect a second round because we must have browser tests in the Manual Testing project.

Please follow the suggestions manually. **Do not use AI tools!**

raman-m:
Hi Mateusz,
We do have a predecessor. Another contributor opened this PR a year ago, but his contribution stopped.
It would be good if you or I reviewed the previous solution so that we could borrow the idea for your feature branch.

## Predecessor
- #2300

pixellos:
Hi, Thanks for the review, will look on it, but I'm short of time today, so probably I'll address the review tomorrow. I had quick look to mentioned PR - I don't see any reason to introduce new delegating handler to skip part of pipeline - the DisableBuffering is doing simillar thing, but it is not reading the line-by-line content as previous PR was. I'd close that old PR as it seems to be in very early stage of development.

pixellos:
Hi, sorry, this week is crazy for me, I'm trying to reproduce it using manual tests, but it seems one of my assumptions were wrong.

1. One problem I had originally was solved by thix X-Accel header - that can't be really tested using manual test
2. Problem with streaming **without** NGINX happens on IIS/IISExpress, not on kestrel,, 
 
Because of that i introduced both normal and IIS Express mode - setting IIS Express from command line is nightmare - i need to get back  to those changes one time more, thus moving to draft state.

I need to revise our Playwright test, but I was convinced that the 'ResponseCompression' does cause buffering, but cannot reproduce it reliably - so i need to spend more time on them.


https://github.com/user-attachments/assets/bf2197c3-b625-4368-9526-260fc5a6ccfd

test/Ocelot.ManualTest/IisExpressBootstrap.cs
```diff
@@ -0,0 +1,185 @@
+using System.Diagnostics;
```
pixellos, line 1: I need to think about it one more time.

testing/PortFinder.cs
pixellos: When running per-release tests in parallel, three processes active.

src/Ocelot/Responder/HttpContextResponder.cs
raman-m: Can't you just press the button to apply suggestion? 🙄 
Comparing to `true` is redundant!

raman-m:
> - commit https://github.com/ThreeMammals/Ocelot/pull/2383/commits/d8f0eb73c0e1b491de3fd98543a89de3dbc8e6db
> - build [642](https://github.com/ThreeMammals/Ocelot/actions/runs/25464827088) failed

Why do you commit tests which fail? 🫤

src/Ocelot/Responder/HttpContextResponder.cs
raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/2383/commits/acb2b0c91e46deef0306c85fc7533c6d296523a6 ✔️

src/Ocelot/Responder/HttpContextResponder.cs
raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/2383/commits/acb2b0c91e46deef0306c85fc7533c6d296523a6 ✔️

pixellos:
> Why do you commit tests which fail? 🫤
Wanted to show this IIS manual test to get opinion of it as priority, havent checked the test after small refactor, sorry.

https://github.com/ThreeMammals/Ocelot/pull/2383#issuecomment-4392678106 - do You have any thought on this? As we can't do automated IIS tests on other machine than windows.

Merged latest develop

test/Ocelot.ManualTest/IisExpressBootstrap.cs
pixellos: I did few tries, but this solution seems to be most stable from tested so far.

raman-m:
> @pixellos commented [on May 7](https://github.com/ThreeMammals/Ocelot/pull/2383#issuecomment-4392678106)

> 1. One problem I had originally was solved by thix X-Accel header - that can't be really tested using manual test

OK, but how will you test this Nginx env? Will you deploy manually, or will you deploy an official NuGet package and then perform testing?

> 2. Problem with streaming **without** NGINX happens on IIS/IISExpress, not on kestrel,,

But it works on Kestrel, right? And why do you waste your time playing with IIS?

> Because of that i introduced both normal and IIS Express mode - setting IIS Express from command line is nightmare - i need to get back  to those changes one time more, thus moving to draft state.

Did you know that the [Ocelot team doesn't recommend deploying to IIS](https://ocelot.readthedocs.io/en/latest/introduction/gotchas.html)? Do you really host Ocelot in IIS for microservices? 😄 

IIS is good for website hosting, but if you're deploying microservices, it's better to use Kestrel. Kestrel is a cross-OS web server that works perfectly on any OS for deploying containers (Docker, K8s pods).

> I need to revise our Playwright test, but I was convinced that the 'ResponseCompression' does cause buffering, but cannot reproduce it reliably - so i need to spend more time on them.

ResponseCompression is an IIS feature, right? It's up to you how much time you want to waste on this web server.

raman-m:
> @pixellos commented [on May 11](https://github.com/ThreeMammals/Ocelot/pull/2383#issuecomment-4418732099)

> https://github.com/ThreeMammals/Ocelot/pull/2383#issuecomment-4392678106 - do You have any thought on this?

Already answered, see above.

> As we can't do automated IIS tests on other machine than windows.

But you/we can skip tests if the platform is non-Windows, right? )))

> Merged latest develop 

Thanks! I have marked the [PR as ready for review](https://github.com/ThreeMammals/Ocelot/pull/2383#event-25378766265). I will do the code review shortly.

test/Ocelot.ManualTest/IisExpressBootstrap.cs
raman-m: Well... I can't imagine why the IIS environment is critical for you. Do you or your team host Ocelot on IIS?
Having IIS bootstrap support would be nice, but I don't require it. Performing manual tests for running Ocelot on Kestrel is absolutely fine.

raman-m:
Hi man! @pixellos 
Are you waiting for my code review? Will do today... I was busy prev week as you.
Do you still want to stabilize IIS Express bootstrapping in tests?

pixellos:
Hi @raman-m - please wait with code review, hadn't had time to get back to this topic, I'll try to find time this week, but I can't promise it. 

Will try to address Your comments and concerns, for now I'd mark as draft and let's back off this topic later on.

raman-m:
In that case, I'll contact you after the [.NET 10](https://github.com/ThreeMammals/Ocelot/milestone/13) release...

raman-m:
> @pixellos marked this pull request as draft [on May 18, 2026](https://github.com/ThreeMammals/Ocelot/pull/2383#event-25650542445)

Could we mark this PR as "Ready for review" and continue development? I made the latest [v25](https://github.com/ThreeMammals/Ocelot/releases/tag/25.0.0) release a month ago, and I'm fully free to help with PRs.
You need to turn this PR into "Ready for review" confirming your readiness to develop, and I'll make a 2nd code review.
How was your summer?

pixellos:
HI @raman-m  - i'll do my best to get back to this topic this weekend.

Also, can we agree on scope? 
I don't need anymore the IIS handling, nginx handling (moved to different stack in which i don't have any of problems that were reason for this PR in first place), and the Kestrel works as/is without any changes in ocelot.

I have proposal, to keep only Playwright handling as smoke tests, to be sure that SSE thru ocelot works on kestrel.

What do You think?

raman-m:
Mateusz,
I'm not sure about the scopes, but what I am sure of is adding the PR to the current release so it will be in version 25.1 by the end of this year. I'll do the next code review shortly, please be patient.

First, could you look at https://github.com/ThreeMammals/Ocelot/issues/941#issuecomment-2241441031 and explain to me what the author meant by "enabled request body buffering"?
**What does this code block actually do?** 👇 
```C#
app.Use(async (context, next) => {
    context.Request.EnableBuffering();
    await next();
});
```
I'd like to ensure this option should be managed in case of SSE streaming.

raman-m:
I've got an idea on scopes! 😄 
Since we intro new protocol support we must have the following:
- automated browser test in acceptance testing. It could be a separate project in future.
- a dedicated testing webpage in ManualTest project. Thus manual tester opens it in any browser and runs tests by pressing JavaScript buttons.
- optional. Integrate automated test into manual test project which can be ran in special mode of terminal app. This is my area of responsibility.
- a dedicated sample in Ocelot.Sampes.slnx solution.
- a dedicated docs as rst of a webpage for SSE proto. This is my area of responsibility.
- optional. Support configuration in JSON for a new `sse` scheme (to be discussed). It depends on either the SSE feat is built-in Core  or the pipeline supports SSE request on the fly.

To be reviewed...

pixellos:
Hi @raman-m 
https://github.com/ThreeMammals/Ocelot/pull/2423
I created new PR, to superseed this one.
After doing TDD tests of this one more time, i can see that it now runs nativelly when not on IIS.

As i did wrote, i'm not interested in IIS support anymore, so https://github.com/ThreeMammals/Ocelot/pull/2423 should showcase that SSE thru ocelot works.

So no changes are needed, and especially not those in current PR.

Created this new PR to just add tests to showcase those. 

Please let me know if what i just mentioned is enough for You.

Cheers

raman-m:
> I created new PR #2423, to supersede this one.

Why? We did a lot for feature delivery right in this PR. In my opinion, the PR readiness is 70%+.
The goal of PR #2423 is to show that Ocelot supports SSE in the current version, right? It would be a bit surprising, but Ocelot has supported streaming requests since v23.x+. Thus, this PR proves that fact.
The goal of the current PR is to test everything and inform the community so they can use the feature and test it with various libraries and software generating of SSE traffic.

> After doing TDD tests of this one more time, i can see that it now runs natively when not on IIS.

Congrats! This is awesome achievement. FYI [IIS webserver is not supported](https://ocelot.readthedocs.io/en/develop/introduction/gotchas.html#iis) in our project.

> As i did wrote, i'm not interested in IIS support anymore, so https://github.com/ThreeMammals/Ocelot/pull/2423 should showcase that SSE thru ocelot works.

OK

> So no changes are needed, and especially not those in current PR.

We still need to develop this PR more. Since you're no longer motivated, I will continue development on my own.

> Created this new PR to just add tests to showcase those. 

I've compared both versions of those tests. `ServerSentEventsTests` class is now cleaned up from SignalR logic, strictly focusing on native SSE emulation. Thus, the first version of these tests could be renamed to `SignalRServerSentEventsTests`. Also I noted that the `GivenOcelotIsRunningWithCompression` method is used indicating usage of `app.UseResponseCompression()` which is weird. Why do you enable this compression feat? Does it support SSE somehow?

And `SSEBrowserTests` have been optimized after refactoring. They could be merged into the first version.

Finally, I'm impressed by the fact that you created a second PR. Therefore,
Given: 2 branches on the local machine in the repo.
Given: the AI agent created a second branch containing redeveloped code.
When: you want to deliver the second version to the first version branch.
Then: I recommend using `git cherry-pick` command.
VS IDE also has a helpful command to do this in the Git GUI.

Please DO NOT create additional PRs while the current one is open.
**I will merge 2nd branch to the current one today.**

> Please let me know if what i just mentioned is enough for You.

It is enough! Thank you!

src/Responder/HttpContextResponder.cs
```diff
@@ -75,12 +81,52 @@ public async Task SetErrorResponseOnContext(HttpContext context, DownstreamRespo
...
     protected virtual async Task WriteToUpstreamAsync(HttpContext context, DownstreamResponse downstream)
     {
+        if (IsServerSentEventsStreaming(context, downstream))
+        {
+            var feature = context.Features.Get<IHttpResponseBodyFeature>();
+            if (feature == null)
+            {
+                var server = context.RequestServices.GetService(typeof(IServer)) ?? "Unknown";
+                _logger.LogWarning(() => $"{nameof(IHttpResponseBodyFeature)} is null for SSE request. Buffering cannot be disabled. Server: {server}, OS: {RuntimeInformation.OSDescription}, Framework: {RuntimeInformation.FrameworkDescription}");
+                return;
```
ggnaegi, line 106: I think there we have a problem, what do you think @raman-m? it returns too early and never reach `ReadAsStreamAsync` and `CopyToAsync` and the caller will never get any body back at all.

src/Responder/HttpContextResponder.cs
```diff
@@ -12,10 +16,12 @@ namespace Ocelot.Responder;
 public class HttpContextResponder : IHttpResponder
 {
     private readonly IRemoveOutputHeaders _removeOutputHeaders;
+    private readonly IOcelotLogger _logger;
 
-    public HttpContextResponder(IRemoveOutputHeaders removeOutputHeaders)
+    public HttpContextResponder(IRemoveOutputHeaders removeOutputHeaders, IOcelotLoggerFactory loggerFactory)
```
ggnaegi, line 21: We are modifying the constructor of a public API, we might have a breaking change, no?

src/Responder/HttpContextResponder.cs
```diff
@@ -75,12 +81,52 @@ public async Task SetErrorResponseOnContext(HttpContext context, DownstreamRespo
...
+    public const string TextEventStreamMediaType = MediaTypeNames.Text.EventStream;
+#else
+    public const string TextEventStreamMediaType = "text/event-stream";
+#endif
+
+    protected virtual bool IsServerSentEventsStreaming(HttpContext context, DownstreamResponse downstream)
+    {
+        var mediaType = downstream.Content.Headers.ContentType?.MediaType ?? string.Empty;
+        return mediaType.Equals(TextEventStreamMediaType, StringComparison.OrdinalIgnoreCase)
+            || mediaType.Contains(TextEventStreamMediaType, StringComparison.OrdinalIgnoreCase);
```
ggnaegi, line 94: Why contains? We should be more defensive here, as stated there, as `This event stream format's [MIME type](https://mimesniff.spec.whatwg.org/#mime-type) is [text/event-stream](https://html.spec.whatwg.org/multipage/iana.html#text/event-stream).`

acceptance/ServerSentEvents/SSEBrowserTests.cs
```diff
@@ -0,0 +1,249 @@
...
+            else if (e.Text == "Received: Message 3")
+            {
+                msg3Tcs.TrySetResult(sw.ElapsedMilliseconds);
+            }
+        };
+
+        var html = $@"
+            <html>
+            <head>
+                <script src='https://cdnjs.cloudflare.com/ajax/libs/microsoft-signalr/8.0.0/signalr.min.js'></script>
```
ggnaegi, line 198: this is an external dependency that could break over time. Great that you offer E2E tests, i would add the signalr reference as an asset.

ggnaegi (changes requested):

PR #2385, opened 2026-04-22
#2384 Built-in Circuit Breaker for Quality of Service (express edition, no Polly required)

ocelotgateway:
## Closes #2384 
- #2384 

## Predecessors
- #2380 
- #561 

## Proposed Changes
Ocelot's QoS feature previously had no built-in circuit breaker implementation — it depended entirely on the external `Ocelot.QualityOfService.Polly` package. This adds a lightweight, built-in circuit breaker so QoS works out of the box.

## Core implementation

- **`src/Ocelot/Infrastructure/DesignPatterns/CircuitBreaker.cs`** — Thread-safe state machine (`Closed → Open → HalfOpen → Closed`) inspired by HystrixCircuitBreaker. `MinimumThroughput ≤ 0` disables circuit-opening while still allowing timeout enforcement.

- **`src/Ocelot/QualityOfService/CircuitBreakerDelegatingHandler.cs`** — `DelegatingHandler` that integrates with `MessageInvokerPool`'s handler chain. Counts 5xx responses and non-cancellation exceptions as failures. Enforces per-request timeout via `QoSOptions.Timeout` using a linked `CancellationTokenSource`. Key subtlety: `OperationCanceledException` is re-thrown without recording a failure in the inner `SendAndTrackAsync`, preventing double-counting when the QoS timeout fires.

```csharp
// OcelotBuilder now registers a default — overridable by external packages via AddSingleton after AddOcelot()
Services.TryAddSingleton<QosDelegatingHandlerDelegate>(provider =>
    (route, contextAccessor, loggerFactory) => new CircuitBreakerDelegatingHandler(route, loggerFactory));
```

## DI registration

- **`OcelotBuilder.cs`** — Registers `CircuitBreakerDelegatingHandler` as the default `QosDelegatingHandlerDelegate` via `TryAddSingleton`. Polly (or any other QoS package) can still override by registering after `AddOcelot()`.

## Tests

- **`CircuitBreakerTests.cs`** (new, 12 tests) — Unit tests covering all state transitions and edge cases including zero/negative `MinimumThroughput`.

- **`DynamicRoutingTests.cs`** — Removed `Skip` from `ShouldApplyGlobalQosOptions_ForAllDynamicRoutes` and `ShouldApplyGlobalQosOptions_ForAllDynamicRoutes_WithGroupedOpts`. Removed the `FakeQualityOfServiceDelegatingHandler` override from `WithDiscoveryAndQualityOfService` — both tests now run against the real built-in handler and pass.

---
[...]

Issue #2378, opened 2026-04-06
Split the monorepo with the goal of having independent releases of projects


raman-m:
A long time ago, in a distant kingdom, King Tom migrated micro-projects to this repository to adopt feature #734. However, the [Ocelot extension projects'](https://github.com/ThreeMammals/Ocelot/tree/24.1.0/src) releases depend on unpredictable release dates of the appropriate integrated library projects. This delays the Ocelot release significantly. As a result, these packages must be deprecated, and their respective sources should be moved to separate micro-repos with individual release strategies.

## Predecessors
- #734 
- #2334 
- #2363 
- #2371 

## Tasks
- [x] Deprecate Ocelot.Provider.Polly project &rarr; **Done** in #2380 ✔️ 
- [x] Deprecate Ocelot.Provider.Consul project &rarr; **Done** in #2388 ✔️ 
- [ ] Deprecate Ocelot.Provider.Kubernetes project

raman-m:
Needs to shorten file paths as the last task

PR #2387, opened 2026-04-30
#2386 Overridable WebSocket buffer size via virtual property and custom middleware injection

### LINKED ISSUE #2386: Add configurable WebSocket buffer size per route

Each route should support an optional `WebSocketBufferSize` configuration property
that overrides the default receive/send buffer size used in `WebSocketsProxyMiddleware`.

The buffer size in `WebSocketsProxyMiddleware` is hardcoded to 4096 bytes.
This causes choppy/degraded video streams over WebSocket connections when hosting
with HTTP.sys. HTTP.sys is more sensitive to small receive buffers than Kestrel
due to differences in kernel-level I/O handling — 4096 bytes is insufficient
for high-bandwidth routes such as video streaming.

  1. Host Ocelot with HTTP.sys (not Kestrel)
  2. Configure a WebSocket route proxying a video stream
  3. Observe choppy/degraded stream quality caused by the 4096-byte buffer limit

## Specifications

  - Version: 25.0
  - Platform: .NET 10, Windows (HTTP.sys)
  - Subsystem: WebSockets

erannevo:
## Closes #2386
- #2386

## Changes

- `DefaultWebSocketBufferSize` changed from `private const int` to `protected virtual int` property in `WebSocketsProxyMiddleware`, allowing subclasses to override the buffer size
- `WebSocketsProxyMiddleware` property added to `OcelotPipelineConfiguration` — follows the same middleware injection pattern as `AuthenticationMiddleware`, `AuthorizationMiddleware`, etc.
- `OcelotPipelineExtensions` updated to use `UseIfNotNull<WebSocketsProxyMiddleware>` so custom implementations are injected into the WebSocket pipeline
- Unit tests added for default buffer size value and subclass override

## Usage

```csharp
// 1. Subclass the middleware
public class LargeBufferWebSocketsProxyMiddleware : WebSocketsProxyMiddleware
{
    public LargeBufferWebSocketsProxyMiddleware(IOcelotLoggerFactory logging, RequestDelegate next, IWebSocketsFactory factory)
        : base(logging, next, factory) { }

    protected override int DefaultWebSocketBufferSize => 65536;
}

// 2. Inject via OcelotPipelineConfiguration
app.UseOcelot(new OcelotPipelineConfiguration
{
    WebSocketsProxyMiddleware = (ctx, next) =>
    {
        var mw = new LargeBufferWebSocketsProxyMiddleware(
            ctx.RequestServices.GetRequiredService<IOcelotLoggerFactory>(),
            _ => next(),
            ctx.RequestServices.GetRequiredService<IWebSocketsFactory>());
        return mw.Invoke(ctx);
    }
});
```

## Verification

- Unit tests pass on net8.0, net9.0 and net10.0
- Manually verified: overriding buffer size to 65536 on a video streaming route hosted with HTTP.sys resolved choppy/degraded stream quality
- Acceptance test failures are pre-existing and unrelated to this change:
  - `DefaultVersionPolicyTests` — HTTP version negotiation, infra-dependent
  - `ClientWebSocketTests.Http20CLient_DirectConnection` — HTTP/2 WebSocket, not supported in local env
  - `ConsulServiceDiscoveryTests` — requires running Consul instance
  - `LoadTests.ShouldLoadRegexCachingHeavily_NoMemoryLeaks` — memory/timing sensitive
  - `MultipleAuthSchemesFeatureTests` — infra-dependent


raman-m:
Thanks for opening linked PR for the issue #2386, but...

>  ### Changes
>  - Add `WebSocketBufferSize` (nullable `int?`) property to `FileRoute` and `DownstreamRoute`
> 

No, man! We haven't discussed that, and I don't accept such changes — at least for now, maybe in the future after code reviews.

You are proposing changes that are too complex for a problem that can be solved with [Metadata](https://ocelot.readthedocs.io/en/latest/features/metadata.html). You define an extra property as a Metadata option, then you read this option in custom middleware inheriting `WebSocketsProxyMiddleware`. That's it — problem solved.
The only problem is injecting a custom middleware for web sockets into Ocelot's pipeline...

There are 2 ways to evolve:
1. Make each route configurable. But you know not all routes are WebSocket proto ones.

2. Tune `WebSocketsProxyMiddleware` and inject its descendant class into the pipeline. This is a more global configuration approach. Per-route tuning would require a dedicated option on the route — your initial proposal.

I recommend starting with the global config or a custom middleware setup, because that will solve the design problem of leaking the middleware injection. The original Ocelot's middleware should be more configurable via its virtual properties and methods.

erannevo:
Thanks for the feedback, Raman!

If I understand correctly, the preferred approach is to read `WebSocketBufferSize` from `route.MetadataOptions.Metadata` directly inside `WebSocketsProxyMiddleware`, rather than adding a new first-class property to `FileRoute` and `DownstreamRoute`. This would keep the route model clean and reuse the existing Metadata infrastructure.

Before I rework the PR, I want to confirm:
1. Should `WebSocketsProxyMiddleware` read from Metadata internally, or is the intent that users provide their own subclass middleware?
2. If the latter — is there an existing mechanism to inject a custom WebSocket middleware into Ocelot's pipeline, or is that something that needs to be added first?

Looking forward to your thoughts so we can get this right!

raman-m:
@erannevo What is your LinkedIn? Your GH profile is empty.

raman-m:
> Should `WebSocketsProxyMiddleware` read from Metadata internally, or is the intent that users provide their own subclass middleware?

No, of course not! We shouldn't read custom props defined in [Metadata](https://ocelot.readthedocs.io/en/latest/features/metadata.html) options in any other Ocelot core middlewares.

The second approach: **in own subclass middleware!**
[Metadata](https://ocelot.readthedocs.io/en/latest/features/metadata.html) was specifically designed for custom overrides, not for changing the route options schema. Read [the docs](https://ocelot.readthedocs.io/en/latest/features/metadata.html)!
However, `WebSocketsProxyMiddleware` could be made more virtual.

> If the latter — is there an existing mechanism to inject a custom WebSocket middleware into Ocelot's pipeline, or is that something that needs to be added first?

**There is no such mechanism!** The [WebSockets pipeline](https://github.com/ThreeMammals/Ocelot/blob/fe672ec02c2b7c4dc5c5b8167bf5f39d24363d47/src/Ocelot/Middleware/OcelotPipelineExtensions.cs#L37) is poor in its functionality. The pipeline must be extended to support dependency injection. I will help you redesign the WS pipeline during multiple code reviews, because the pipeline is a very important part of the Ocelot core.
Here is WS pipeline:
https://github.com/ThreeMammals/Ocelot/blob/fe672ec02c2b7c4dc5c5b8167bf5f39d24363d47/src/Ocelot/Middleware/OcelotPipelineExtensions.cs#L36-L46

erannevo:
> > `feature/websocket-buffer-size-25.0`
> 
> Interesting feature branch name 🥲 I am not sure the feature will be delivered in version 25.0 because it is related to [.NET 10](https://github.com/ThreeMammals/Ocelot/milestone/13). I expect that feature could be part of the next release after the current one. I hope the release process will be faster soon after I [split the monorepo](https://github.com/ThreeMammals/Ocelot/issues/2378) in the current version.

We’re currently on version 24.1 (.NET 8), so I built the fix locally for now.
I assumed that if I’m already submitting a PR, it would make sense to target the latest release branch.

It’s been a bit of a journey 😅
We’re running Ocelot on Kestrel and discovered that Kestrel doesn’t support Kerberos double-hop, so we had to refactor to http.sys. That’s when I ran into the WebSocket performance issue.

After some digging (and a bit of help from Claude 😄), I forked the 24.1 branch, applied the fix locally, and the issue was resolved—without needing to bypass the WebSocket middleware.

erannevo:
While we wait for the right version to target, I've been thinking about the right approach based on your feedback.

To confirm — is this what you had in mind?

1. Make `DefaultWebSocketBufferSize` a `protected virtual` property in `WebSocketsProxyMiddleware`
2. Add `WebSocketMiddleware` override to `OcelotPipelineConfiguration` (same pattern as `AuthenticationMiddleware`)
3. Update `OcelotPipelineExtensions` to use the configured type if set, fall back to `WebSocketsProxyMiddleware`

No route model changes, no signature changes — users subclass and inject. Happy to start reworking the PR once we're aligned.

raman-m:
> @erannevo commented [on April 30](https://github.com/ThreeMammals/Ocelot/pull/2387#issuecomment-4354249280)

> While we wait for the right version to target, I've been thinking about the right approach based on your feedback.

If the PR is (will be) ready (with "Ready for delivery" status) and gets merged into the develop branch, I can create a 25.1 minor version with your feature included — but not earlier than the major version 25 release. Sorry.

> To confirm — is this what you had in mind?
> 1. Make `DefaultWebSocketBufferSize` a `protected virtual` property in `WebSocketsProxyMiddleware`
> 2. Add `WebSocketMiddleware` override to `OcelotPipelineConfiguration` (same pattern as `AuthenticationMiddleware`)
> 3. Update `OcelotPipelineExtensions` to use the configured type if set, fall back to `WebSocketsProxyMiddleware`
> 

Yes, it is. This is the correct plan. **Please go ahead.**
Here is an example of injection as part of the [Middleware Injection](https://github.com/ThreeMammals/Ocelot/blob/develop/docs/features/middlewareinjection.rst) feature:
https://github.com/ThreeMammals/Ocelot/blob/fe672ec02c2b7c4dc5c5b8167bf5f39d24363d47/src/Ocelot/Middleware/OcelotPipelineExtensions.cs#L92

**The goal**: extend the WebSocket pipeline with [middleware injection](https://github.com/ThreeMammals/Ocelot/blob/develop/docs/features/middlewareinjection.rst) and prepare the desired middleware for user overrides.

> No route model changes, no signature changes — users subclass and inject. Happy to start reworking the PR once we're aligned.

We will return to this functionality during dev phase 2. Adding JSON configuration still makes sense, but I will brainstorm how to extend the [route schema](https://github.com/ThreeMammals/Ocelot/blob/develop/docs/features/configuration.rst#route-schema) in the right way. This is the most painful part of every feature that requires JSON configuration, because we must have consensus and excellent usability. It seems we have to introduce dedicated WebSocket options — aka a [File](https://github.com/ThreeMammals/Ocelot/tree/develop/src/Ocelot/Configuration/File) model.

raman-m:
>@erannevo commented [on April 30](https://github.com/ThreeMammals/Ocelot/pull/2387#issuecomment-4353585554)

I'll answer you later... after the weekend...

erannevo:
Hi @raman-m, I've reworked the PR based on your feedback.

The new approach:
- `DefaultWebSocketBufferSize` is now a `protected virtual` property in `WebSocketsProxyMiddleware`
- `WebSocketsProxyMiddleware` override added to `OcelotPipelineConfiguration`, following the same middleware injection pattern as `AuthenticationMiddleware`
- `OcelotPipelineExtensions` updated to use `UseIfNotNull<WebSocketsProxyMiddleware>`

PR description has been updated with the full change summary and a usage example. All tests pass on net8.0, net9.0 and net10.0.

Looking forward to your review!

raman-m:
> PR description has been updated with the full change summary and a usage example. All tests pass on net8.0, net9.0 and net10.0.

Good! I see unit tests only. Please write at least one acceptance test utilizing custom middleware.
Also, it makes sense to develop a small sample in the `Ocelot.Samples.slnx` solution.
After this, we will start the second development phase.

And why did you remove the old commits? Did you perform a complete rebase onto develop?
Did you save the old feature branch somehow?

raman-m:
@erannevo commented [on April 30](https://github.com/ThreeMammals/Ocelot/pull/2387#issuecomment-4353585554)

Well...

>I assumed that if I’m already submitting a PR, it would make sense to target the latest release branch.
>It’s been a bit of a journey 😅

Yep! Submitting a PR doesn't mean the PR is ready for delivery. Some PRs stay open for years in our backlog for multiple reasons: inactive author, irrelevant changes, absurd proposed changes, or unprofessional and low-quality code. Starting from 2025, there is another problem: PRs opened by AI bots or users who don't write code manually.

There are two main reasons why a PR stays open and is not merged: 1) not following the [Development Process](https://ocelot.readthedocs.io/en/latest/building/devprocess.html) (code reviews, tests, docs), and 2) not reaching the "Ready for Delivery" status ([example](https://github.com/ThreeMammals/Ocelot/pull/2385#pullrequestreview-4191711511)). I give this status when the [Development Process](https://ocelot.readthedocs.io/en/latest/building/devprocess.html) has been followed by a contributor and the PR has good quality and is a useful pro contribution. All of this is about professional development.

Finally, I would say: if you want your PR to be merged in the current release, then at least the PR must have the "[Ready for Delivery](https://github.com/ThreeMammals/Ocelot/pull/2385#pullrequestreview-4191711511)" status. Usually, I don't merge external users' PRs for releases that upgrade the .NET SDK because it slows down migration velocity and delays the release date. However, I could make an exception — which is why we have two patches and one new feature in the current [.NET 10](https://github.com/ThreeMammals/Ocelot/milestone/13) release.

>We’re running Ocelot on Kestrel and discovered that **Kestrel doesn’t support Kerberos double-hop, so we had to refactor to HTTP.sys**. That’s when I ran into the *WebSocket performance issue*.

Yep! HTTP.sys is a web server running on Windows OS only, with deep integration with Windows Authentication providers and the Windows OS kernel. On the other hand, Kestrel is a cross-OS, cross-platform web server, abstracting away from the Windows API.

It would be good if Ocelot's "WebSocket performance issue" were a single problem. But if there is something else we can review, I'd like to hear your feedback.

erannevo:
> > PR description has been updated with the full change summary and a usage example. All tests pass on net8.0, net9.0 and net10.0.
> 
> Good! I see unit tests only. Please write at least one acceptance test utilizing custom middleware. Also, it makes sense to develop a small sample in the `Ocelot.Samples.slnx` solution. After this, we will start the second development phase.
> 
> And why did you remove the old commits? Did you perform a complete rebase onto develop? Did you save the old feature branch somehow?

Thanks, I’ll add at least one acceptance test using custom middleware and also prepare a small sample under `Ocelot.Samples.slnx`.

Regarding the commits — yes, I rebased because the original approach was not accepted, and my main motivation is to get the bug fixed as soon as possible. We are currently using an Ocelot package from our local feed with this fix, and I’d prefer not to maintain a local patched version long-term.

I still have the original proposed feature branch/fix saved, and I’m happy to help finish that too without rushing it, once we temporarily solve the current performance issue.

I’m committed to my full-time job, but I’ll do my best to close this PR as proposed.

erannevo:
Thanks for the context Raman.

> Yep! Submitting a PR doesn't mean the PR is ready for delivery. Some PRs stay open for years in our backlog for multiple reasons: inactive author, irrelevant changes, absurd proposed changes, or unprofessional and low-quality code. Starting from 2025, there is another problem: PRs opened by AI bots or users who don't write code manually.
> 

To address the AI comment — I do use AI also to move faster and navigate the codebase; without it, finding the root cause and matching solution would have taken much longer. That said, AI hallucinates a lot — Claude initially pointed me toward HTTPS buffer tuning, which I ruled out immediately — the stream runs over WebSocket, not HTTP response bodies, so that would have done nothing. I write and own every line, and I understand what it does and why. The actual fix came from reading `WebSocketsProxyMiddleware` directly and tracing the hardcoded 4096-byte buffer through the HTTP.sys receive path.

> It would be good if Ocelot's "WebSocket performance issue" were a single problem. But if there is something else we can review, I'd like to hear your feedback.

Regarding other things to review in Ocelot — once we ship our release, I'll sync with the relevant stakeholders and report back if anything else surfaces. We use Ocelot fairly extensively: API gateway over multiple backend services, WebSocket proxying, Consul-based service discovery with custom load balancers (least-connections, sticky sessions, capacity-aware), Windows/Kerberos authentication with impersonation, circuit breaking via Polly, and a custom middleware pipeline. If we hit friction in any of those areas, you'll hear from me.

I'll get the acceptance test and sample added this week.

raman-m:
> Regarding the commits — yes, I rebased because the original approach was not accepted, and **my main motivation is to get the bug fixed as soon as possible.** We are currently using an Ocelot package from our local feed with this fix, and I’d prefer not to maintain a local patched version long-term.

Got it! We can agree on including the patch in the next beta version, i.e., 25.0 Beta 3. I do have to make a beta release due to internal testing tasks. Thus, your feature (the patch) may be included in this Beta 3 version already.

> I rebased because the original approach was not accepted 

[My old answer](https://github.com/ThreeMammals/Ocelot/pull/2387#issuecomment-4353136863):
> No, man! We haven't discussed that, and I don't accept such changes — at least for now, **maybe in the future after code reviews.**

You understood incorrectly. Your original solution is right, but it is the most complicated design. I proposed extending the WebSocket middleware first, and then we can return to developing the JSON feature. This is the second development round.

Please make a backup of these old commits, since I didn't do that locally.

> P.S. Please focus on pushing commits, not writing comments 🥲

erannevo:
> @raman-m commented [on May 6](https://github.com/ThreeMammals/Ocelot/pull/2387#issuecomment-4388048729)

Hi Raman,
I've pushed the requested changes (acceptance test + sample).
The original per-route solution is preserved on my fork for phase 2:
https://github.com/erannevo/Ocelot/compare/develop...feature/websocket-configurable-buffer-size

Eran

raman-m:
Hi Eran!

> I've pushed the requested changes (acceptance test + sample).

Very well. Thanks! Today I will do code review and possibly push commits.

We need to clarify the priority of delivery of this feat since you said your team needs an official package with this patch. Thus, I am now planning the Beta 3 release (see PR #2389). The feat can be included in this beta 3 release, but we need to intensify our efforts on development. If not included in Beta 3, then it might be included in the final v25 release.

> The original per-route solution is preserved on my fork for phase 2:
https://github.com/erannevo/Ocelot/compare/develop...feature/websocket-configurable-buffer-size

Please merge! I'd like to review complete solution.

erannevo:
Done - https://github.com/ThreeMammals/Ocelot/pull/2390

src/Ocelot/Middleware/OcelotPipelineExtensions.cs
```diff
@@ -42,7 +42,7 @@ public static RequestDelegate BuildOcelotPipeline(this IApplicationBuilder app,
                 ws.UseMiddleware<DownstreamRequestInitialiserMiddleware>();
                 ws.UseMiddleware<LoadBalancingMiddleware>();
                 ws.UseMiddleware<DownstreamUrlCreatorMiddleware>();
-                ws.UseMiddleware<WebSocketsProxyMiddleware>();
+                ws.UseIfNotNull<WebSocketsProxyMiddleware>(configuration.WebSocketsProxyMiddleware);
```
raman-m: This works just fine, but I need to rethink this design once again in the 2nd round.
Seems we might have other pipeline options, or we could have another version of the `Use` method that injects the type of custom middleware. The problem is how we inject the middleware. I'll explain it to you in the next issue...

samples/WebSocket/Program.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        await next();
+    }
+});
+
+app.UseWebSockets();
+var wsPipeline = new OcelotPipelineConfiguration
+{
+    WebSocketsProxyMiddleware = MyWebSocketsProxyMiddleware,
+};
```
raman-m: What happens if we comment out the `app.UseWebSockets()` method at line 39? 😆

Seriously, it is a question of valid app setup. Since we recommend a user and require them to enable WS functionality in the [WebSockets docs](https://ocelot.readthedocs.io/en/latest/features/websockets.html#configuration), maybe it is time to ensure that this ASP.NET system middleware has been enabled in an automated way (aka `Microsoft.AspNetCore.WebSockets.WebSocketMiddleware` type). This is critical middleware to be added by Ocelot, but it is not, and we still ask the user to add middleware manually, which is not good.

Finally, I'd say we must always add the system middleware in Ocelot Builder.
This is an old pipeline design issue, and it is out of the scope of this PR. Just noting it once again. 🤔 

2. The next issue is anonymous methods, which work fine, but it is better to use named delegates, so I assign the named delegate in the options.

samples/WebSocket/Program.cs
```diff
@@ -0,0 +1,59 @@
...
+    return await File.ReadAllTextAsync(htmlFilePath, cancellation);
+}
+static Task MyWebSocketsProxyMiddleware(HttpContext context, Func<Task> next)
+{
+    Task Next(HttpContext ctx) => next();
+    var loggerFactory = context.RequestServices.GetRequiredService<IOcelotLoggerFactory>();
+    var factory = context.RequestServices.GetRequiredService<IWebSocketsFactory>();
+    var middleware = new CustomWebSocketsProxyMiddleware(loggerFactory, Next, factory);
+    return middleware.Invoke(context);
+}
```
raman-m: And here is the real issue of flawed design.
This method acts as a factory method. We call the service provider, propagate services, create an instance of middleware, and call the middleware correctly.
Adapting the `next` argument by converting it to a `RequestDelegate` is an indicator of flawed design.
We should not convert, adapt, or introduce other anonymous injection methods.
We need something useful to attach the user's middleware knowing only the type:

```c#
app.UseMiddleware<CustomWebSocketsProxyMiddleware>();
```

and nothing more. So, we must know the type of the user's middleware and inject it in the simplest and clearest way, avoiding any developer mistakes.

I propose to consider the following designs:
- Extend `OcelotPipelineConfiguration` options to have a middleware type only
- Extend Ocelot Builder (aka `IOcelotBuilder`) to introduce new `Use` or `Add` helpers

Tomorrow I will try to redesign the injection approach, since the current one is unsafe and requires updating docs where we must explain "how to" in detail. I believe the injection should be simpler.

raman-m (changes requested):
Today, I've tested the sample and proposed a dedicated webpage with Copilot's help, for sure 🥲 and I'm satisfied with the testing results.

The current issue is how we inject the user's middleware. Read my suggestions below.

P.S. You can relax, since I will handle the rest of the work to reach "Ready for delivery" status. I don't require you to fix this, as it would take a long time during pair programming.

samples/WebSocket/Program.cs
raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/2387/commits/3ce3aa43e21fca5de0e366d0b810f88e40285b50 ✔️

samples/WebSocket/Program.cs
raman-m: Partially fixed in commit https://github.com/ThreeMammals/Ocelot/pull/2387/commits/977ecd6ea0f3957e7f2cea4c96c742c400f8179b ☑️ 
**TODO &rarr; Need propagation of `WebSocketOptions` ❗**

src/Ocelot/Middleware/OcelotPipelineExtensions.cs
raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/2387/commits/977ecd6ea0f3957e7f2cea4c96c742c400f8179b ✔️

testing/Steps/WebSocketsSteps.cs
```diff
@@ -242,6 +242,18 @@ void ConfigureWebHost(IWebHostBuilder b) => b
...
     }
 
+    protected Task StartOcelotWithWebSockets(int port, Action<IServiceCollection> configureServices, OcelotPipelineConfiguration pipelineConfig)
+    {
+        var url = DownstreamUrl(port);
+        void ConfigureWebHost(IWebHostBuilder b) => b
+            .UseUrls(url)
+            .ConfigureLogging(WithConsole);
+        void WithWebSocketsAndConfig(IApplicationBuilder app)
+            => app.UseWebSockets().UseOcelot(pipelineConfig).Wait();
```
raman-m: duplicate force call of `UseWebSockets()` due to version after commit https://github.com/ThreeMammals/Ocelot/pull/2387/commits/977ecd6ea0f3957e7f2cea4c96c742c400f8179b

test/Ocelot.AcceptanceTests/WebSockets/WebSocketsFactoryTests.cs
```diff
@@ -45,6 +50,34 @@ public void ShouldProxyWebsocketInputToDownstreamServiceAndUseLoadBalancer()
...
+    public async Task ShouldProxyWebsocketInputToDownstreamServiceUsingCustomMiddleware()
+    {
+        var port = PortFinder.GetRandomPort();
+        var route = GivenRoute("/ws", port);
+        GivenThereIsAConfiguration(GivenConfiguration(route));
+        int ocelotPort = PortFinder.GetRandomPort();
+        bool customMiddlewareInvoked = false;
+        var pipelineConfig = new OcelotPipelineConfiguration
+        {
+            WebSocketsMiddleware = (context, next) =>
```
raman-m: This is injecting via a custom middleware factory controlling all aspects of instance creation.
Need a second test for injection via a middleware type defined in `pipelineConfig` (could be a unit test). This acceptance test can be converted to a `[Theory]`.

raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/2387/commits/8e38475b0c2f4180c5057b4d18c58c33d834f83e ✔️

testing/Steps/WebSocketsSteps.cs
raman-m: Fixed in commit https://github.com/ThreeMammals/Ocelot/pull/2387/commits/8e38475b0c2f4180c5057b4d18c58c33d834f83e ✔️

raman-m (approved):
## Ready for delivery ✅ 
- Code review ✔️ ✔️
- Unit testing ✔️✔️✔️
- Acceptance testing ✔️
- Added sample project ✔️🔥 
- Updated docs ✔️✔️

### P.S.
- TODO &rarr; Need propagation of `WebSocketOptions` ❗

raman-m:
@erannevo 
Hi Eran!

A new version [25.0.0-beta.3](https://github.com/ThreeMammals/Ocelot/releases/tag/25.0.0-beta.3) has been released, which includes this PR feature❗
Please update your projects to use the [25.0.0-beta.3](https://www.nuget.org/packages/Ocelot/25.0.0-beta.3) package and let me know the testing results.

**Note: [Phase 2 feature](https://github.com/ThreeMammals/Ocelot/pull/2390) will be included in the final official release of v25. Please be patient!**

Thanks for your contribution!

PR #2396, opened 2026-06-05
#679 Fix URL claim type authorization

### LINKED ISSUE #679: Ocelot doesn't handle correctly RouteClaimsRequirement with a key as an Url
While creating JWT for a user in my authentication service I use `System.Security.Claims.ClaimTypes` static class with defined string constants for various claims. So, `ClaimTypes.Role == "http://schemas.microsoft.com/ws/2008/06/identity/claims/role"`:
```csharp
var claims = new List<Claim>
{
    new Claim("ID", user.Id.ToString()),
    new Claim(ClaimTypes.Name, user.Username)
};
claims.AddRange(user.Roles.Select(role => new Claim(ClaimTypes.Role, role)));
```
Then, when for some Route in **RouteClaimsRequirement** I write: `"http://schemas.microsoft.com/ws/2008/06/identity/claims/role" : "Admin"`
```json
"RouteClaimsRequirement": {
        "http://schemas.microsoft.com/ws/2008/06/identity/claims/role": "Admin"
}
```
Such Route just disappears somewhere in the guts of middleware (I didn't manage to track down where this happens) and a request results in 404 because a route is not found:
> Error Code: UnableToFindDownstreamRouteError Message: Unable to find downstream route for path: /api/entities/, verb: POST errors found in ResponderMiddleware. Setting error response for request path:/api/entities/, request method: POST

When I use my own claim type like "Role", this works fine. So I assume there are some issues with (de)serialization of a string containing colons or slashes, basically as any URL.

## Specifications
  - Version: 12.0.1 - **19.0.2**

marko1olo:
## Fixes #679 
- #679 

## Summary
- Preserve URL-shaped `RouteClaimsRequirement` keys such as `ClaimTypes.Role` when binding Ocelot JSON configuration.
- Add a post-configure pass that reconstructs nested configuration sections into their original claim-type keys before route authorization uses them.
- Cover direct authorization with URL claim types and configuration binding for mixed plain/URL claim keys.

## Checks

- `git diff HEAD~1 HEAD --check`
- Static search confirmed the post-configure hook is registered and both regression tests are present.

I did not run `dotnet test` locally because the workstation had active `dotnet` processes and sustained 100% CPU during this pass.

### DIFF
--- a/src/Ocelot/Authorization/RouteClaimsRequirementPostConfigureOptions.cs
+++ b/src/Ocelot/Authorization/RouteClaimsRequirementPostConfigureOptions.cs
@@ -1,0 +1,61 @@
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Options;
+using Ocelot.Configuration.File;
+
+namespace Ocelot.Authorization;
+
+public sealed class RouteClaimsRequirementPostConfigureOptions : IPostConfigureOptions<FileConfiguration>
+{
+    private readonly IConfiguration _configuration;
+
+    public RouteClaimsRequirementPostConfigureOptions(IConfiguration configuration)
+    {
+        _configuration = configuration;
+    }
+
+    public void PostConfigure(string name, FileConfiguration options)
+    {
+        if (options?.Routes == null)
+        {
+            return;
+        }
+
+        var routes = _configuration.GetSection(nameof(FileConfiguration.Routes));
+        for (var i = 0; i < options.Routes.Count; i++)
+        {
+            var route = options.Routes[i];
+            var routeClaims = routes.GetSection(i.ToString())
+                .GetSection(nameof(FileRoute.RouteClaimsRequirement));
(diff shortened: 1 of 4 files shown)


raman-m:
**Latest [build 719](https://github.com/ThreeMammals/Ocelot/actions/runs/27423336635) failed.**
We need to review and update the [Should_fix_issue_240](https://github.com/search?q=repo%3AThreeMammals%2FOcelot%20Should_fix_issue_240&type=code) acceptance test. **Until it is fixed, the PR will remain in draft.**

Once you have fixed the test and are ready to collaborate, we can discuss further. Please return to the repo with a clear intention to contribute successfully.
Before that, read the issue #679 thread thoroughly. It contains ideas from other GitHubers, which should at least be considered.

raman-m:
> I did not run `dotnet test` locally because the workstation had active `dotnet` processes and sustained 100% CPU during this pass.

Prove it first!

PR #2406, opened 2026-06-22
#2403 Add `SecurityOptions` support to the WebSocket pipeline

### LINKED ISSUE #2403: IP allow and block list (SecurityOptions) bypassed for WebSocket upgrade requests
### Summary
Ocelot enforces its per-route IP allow and block lists (`SecurityOptions.IPAllowedList` / `IPBlockedList`) through `SecurityMiddleware`, which is only registered in the main HTTP request pipeline. When an incoming request is a WebSocket upgrade request, Ocelot forks the request into a separate, much shorter pipeline that does not include `SecurityMiddleware`. As a result, a client whose IP address is explicitly blocked (or not present in the allow list) can still reach the protected downstream service by sending a WebSocket upgrade request, completely bypassing the operator's IP filtering control.

### Details
The Ocelot pipeline is assembled in `src/Ocelot/Middleware/OcelotPipelineExtensions.cs`. For normal HTTP requests the pipeline includes the security module:

```csharp
// This security module, IP whitelist blacklist, extended security mechanism
app.UseMiddleware<SecurityMiddleware>();
```

`SecurityMiddleware` runs `IPSecurityPolicy` (`src/Ocelot/Security/IPSecurityPolicy.cs`), which compares `context.Connection.RemoteIpAddress` against the route's expanded `IPBlockedList` / `IPAllowedList` and sets a pipeline error when the client IP is blocked or not allowed.

WebSocket upgrade requests are forked away before that middleware ever runs:

```csharp
// If the request is for WebSockets upgrade we fork into a different pipeline
app.UseWebSockets();
app.MapWhen(context => context.WebSockets.IsWebSocketRequest, app => app.ConfigureWebSockets(configuration));
```

The forked pipeline is defined in the same file and contains no security, authentication, or authorization middleware:

```csharp
public static void ConfigureWebSockets(this IApplicationBuilder app, OcelotPipelineConfiguration configuration)
{
    app.UseMiddleware<DownstreamRouteFinderMiddleware>();
    app.UseMiddleware<MultiplexingMiddleware>();
    app.UseMiddleware<DownstreamRequestInitialiserMiddleware>();
    app.UseMiddleware<LoadBalancingMiddleware>();
    app.UseMiddleware<DownstreamUrlCreatorMiddleware>();
    app.UseIfNotNull<WebSocketsProxyMiddleware>(configuration.WebSocketsMiddlewareType);
    app.UseIfNotNull<WebSocketsProxyMiddleware>(configuration.WebSocketsMiddleware, configuration.WebSocketsMiddlewareType is null);
}
```

`SecurityMiddleware` is registered exactly once in the codebase, only in the main pipeline. Because `MapWhen` short-circuits the request into the fork, the IP allow/block evaluation never executes for a WebSocket upgrade. The same defect is present in the 24.1.0 release, where the fork is written inline:

```csharp
app.MapWhen(httpContext => httpContext.WebSockets.IsWebSocketRequest,
    ws =>
    {
        ws.UseMiddleware<DownstreamRouteFinderMiddleware>();
        ws.UseMiddleware<MultiplexingMiddleware>();
        ws.UseMiddleware<DownstreamRequestInitialiserMiddleware>();
        ws.UseMiddleware<WebSocketsProxyMiddleware>();
    });
```

The WebSockets documentation lists authentication and authorization as unsupported over WebSockets, but it does not state that the IP allow/block list security feature is silently bypassed. An operator who restricts a route by IP reasonably expects that restriction to apply to all requests to that route, including WebSocket upgrades.

### PoC
Prerequisites: .NET 10 SDK, the Ocelot source tree. A gateway is configured with one route that proxies to a downstream WebSocket service and that has `SecurityOptions.IPBlockedList` containing the address the client connects from (`127.0.0.1` in this test). A downstream service exposes both an HTTP endpoint and a WebSocket echo endpoint on port 6001.

1. Gateway program (references the local Ocelot project), listening on `http://127.0.0.1:6000`:

```csharp
using System.Net.WebSockets;
using System.Text;
using Microsoft.AspNetCore.Builder;
using Ocelot.Configuration.File;
using Ocelot.DependencyInjection;
using Ocelot.Middleware;

// Downstream service on 127.0.0.1:6001 (HTTP + WebSocket echo)
var downBuilder = WebApplication.CreateBuilder();
downBuilder.WebHost.UseUrls("http://127.0.0.1:6001");
var down = downBuilder.Build();
down.UseWebSockets();
down.Use(async (ctx, next) =>
{
    if (ctx.WebSockets.IsWebSocketRequest)
    {
        using var ws = await ctx.WebSockets.AcceptWebSocketAsync();
        var buf = new byte[1024];
        var res = await ws.ReceiveAsync(new ArraySegment<byte>(buf), CancellationToken.None);
        var msg = Encoding.UTF8.GetString(buf, 0, res.Count);
        var reply = Encoding.UTF8.GetBytes("DOWNSTREAM-WS-ECHO:" + msg);
        await ws.SendAsync(new ArraySegment<byte>(reply), WebSocketMessageType.Text, true, CancellationToken.None);
        await ws.CloseAsync(WebSocketCloseStatus.NormalClosure, "bye", CancellationToken.None);
        return;
    }
    await next();
});
down.MapGet("/{**rest}", () => "DOWNSTREAM-HTTP-OK");
_ = down.RunAsync();

// Ocelot gateway on 127.0.0.1:6000, route blocks 127.0.0.1
var cfg = new FileConfiguration();
cfg.Routes.Add(new FileRoute
{
    UpstreamPathTemplate = "/proxy/{everything}",
    UpstreamHttpMethod = new HashSet<string> { "GET" },
    DownstreamPathTemplate = "/{everything}",
    DownstreamScheme = "http",
    DownstreamHostAndPorts = new List<FileHostAndPort> { new("127.0.0.1", 6001) },
    SecurityOptions = new FileSecurityOptions { IPBlockedList = new List<string> { "127.0.0.1" } },
});

var gwBuilder = WebApplication.CreateBuilder();
gwBuilder.WebHost.UseUrls("http://127.0.0.1:6000");
gwBuilder.Configuration.AddOcelot(cfg);
gwBuilder.Services.AddOcelot(gwBuilder.Configuration);
var gw = gwBuilder.Build();
await gw.UseOcelot();
_ = gw.RunAsync();
await Task.Delay(3000);

// 1) Normal HTTP request from the blocked IP
using var http = new HttpClient();
var r = await http.GetAsync("http://127.0.0.1:6000/proxy/anything");
Console.WriteLine($"[HTTP] status={(int)r.StatusCode}");

// 2) WebSocket upgrade request from the same blocked IP
using var wsc = new ClientWebSocket();
await wsc.ConnectAsync(new Uri("ws://127.0.0.1:6000/proxy/anything"), CancellationToken.None);
var send = Encoding.UTF8.GetBytes("hello-from-blocked-ip");
await wsc.SendAsync(new ArraySegment<byte>(send), WebSocketMessageType.Text, true, CancellationToken.None);
var rbuf = new byte[1024];
var rres = await wsc.ReceiveAsync(new ArraySegment<byte>(rbuf), CancellationToken.None);
Console.WriteLine($"[WS]   reply=\"{Encoding.UTF8.GetString(rbuf, 0, rres.Count)}\"");
```

2. Run it. Observed output:

```
[HTTP] status=401
warn: Ocelot.Responder.Middleware.ResponderMiddleware[0]
      ResponderMiddleware found 1 error ->
      UnauthenticatedError: This request rejects access to 127.0.0.1 IP
      Setting error response for request: GET /proxy/anything

info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://127.0.0.1:6001/anything - - -
[WS]   reply="DOWNSTREAM-WS-ECHO:hello-from-blocked-ip"
```

The normal HTTP request from the blocked IP is rejected with status 401 by `SecurityMiddleware`. The WebSocket upgrade request from the very same blocked IP is proxied through to the downstream (note the downstream log line `GET http://127.0.0.1:6001/anything`) and the client receives the downstream's echo, proving the IP block was bypassed.

### Impact
An attacker whose source IP is explicitly blocked by a route's `SecurityOptions` (or simply absent from a configured allow list) can still reach the protected downstream service by issuing a WebSocket upgrade request to that route. Operators who rely on Ocelot's IP allow/block list to restrict network access to a downstream WebSocket service (or any route reachable via upgrade) lose that protection. The confidentiality and integrity impact is whatever the downstream WebSocket channel exposes (reading and writing messages to the otherwise IP-restricted service).

CurtisRobertOliver:
## Fixes #2403 
- #2403 

## Proposed Changes

- **Enforce IP security on WebSocket upgrades**: Move `SecurityMiddleware` into `ConfigureWebSockets` to apply IP allowed/blocked lists to WebSocket upgrade requests, preventing security bypass.
- **Return 403 Forbidden for IP violations** (not 401): Introduce new `SecurityError` class replacing `UnauthenticatedError` in `IPSecurityPolicy`. IP checks are authorization concerns, not authentication.
- **Refactor security services**: 
  - New `SecurityError` maps to `OcelotErrorCode.SecurityError` and status 403.
  - `SecurityMiddleware` refactored with `HandleWebSocketErrors()` method for RFC 6455 compliance (reject upgrade with 403 on error).
  - `IPSecurityPolicy` simplified: removed nested conditionals, replaced with early-exit logic and `SecurityError`.
- **Update tests**: Correct status codes (401 → 403) in `SecurityOptionsTests`. Add `WebSocketsSecurityTests` for blocked/allowed IP scenarios. Add `SecurityErrorTests` and `SecurityMiddlewareTests` coverage.
- **Documentation**: Update `errorcodes.rst`, `routing.rst`, and `websockets.rst` to document 403 returns and WebSocket security support (v25.0+).

## Tests

- **Updated**: `OcelotPipelineExtensionsTests` (middleware count: 6→7; indices adjusted).
- **Updated**: `SecurityOptionsTests` (status: 401→403).
- **Added**: `WebSocketsSecurityTests`, `SecurityErrorTests`, `SecurityMiddlewareTests`.
- **Updated**: `ErrorsToHttpStatusCodeMapperTests` (added `SecurityError` to forbidden mappings).

## Docs

- `errorcodes.rst`: 403 returned by authorization and security (IP) middlewares.
- `routing.rst`: WebSocket security note and v25.0 reference.
- `websockets.rst`: Security Options listed in Supported features.


CurtisRobertOliver:
The -.0001% decrease in coverage is explained by the removal of 2 lines (which were included in test coverage) from the ConfigureWebSockets method:

`app.UseMiddleware<DownstreamRouteFinderMiddleware>();`
`app.UseMiddleware<MultiplexingMiddleware>();`


The proposed fix makes these lines redundant, as they are executed in the main pipeline (line 43 & line 46) prior to the new location of the WebSocket fork at line 53.

raman-m:
> @CurtisRobertOliver commented [on June 22](https://github.com/ThreeMammals/Ocelot/pull/2406#issuecomment-4773505296)

Curtis, thanks for your interest in Ocelot. 🐈‍⬛ 

Don't worry about unit test coverage now because first we must agree on the final solution, which will have stable code to be covered by tests.

Also, I merged an important PR yesterday that changed the folder structure, so this PR requires branch rebasing to resolve merge conflicts.
I'll review the PR in 6–8 hours (pretty busy during the day).

My quick opinion is that it would be hard to move middlewares to a new position due to the historical order of all middlewares, but this is true for the regular HTTP pipeline. For the WebSocket pipeline, we just need to add security middleware into the existing WS-configurator method.

FYI, the HTTP pipeline and WS pipeline are different due to the natural difference in the ws protocol, so some HTTP middlewares can't be used for the WS protocol. In theory, everything is possible, and we could merge both pipelines into one by having a simple checker in every middleware that checks if the request is a WebSocket one. And this is my idea.

If we will do this redesign in this PR, it will be a great PR solving and closing a batch of issues.
I believe it is possible to have one single pipeline without branching pipelines.

> P.S. I won't verify your account since you have a LinkedIn link in your GitHub profile.
> What you're missing on LinkedIn is a link back to your GitHub account. But it's up to you, as I can see you're a software engineer 😉

CurtisRobertOliver:
Thanks for your insight, Raman! The branch has now been rebased. 

It sounds like my approach of moving the WebSocket fork is not viable, due to resulting alterations of the historic pipeline order. If I understand your proposed solution correctly, it should just take a single line to add security middleware to the WebSocket pipeline. This seems a lot cleaner than my approach. 

What do you think about focusing this PR to unblock issue 2403 (via your suggested solution of adding security middleware to the WS pipeline)? I would be happy to take ownership of the goal to combine WS & HTTP into one single pipeline via a separate PR/issue. That said, I'd like to further discuss the scope of which middlewares require updates, and what exactly each WebSocket check needs to achieve within those middlewares. I don't want to block issue 2403 in the meantime, so I will update this PR shortly to revert my changes & implement your proposed minimal solution. 

 @raman-m let me know if you have any thoughts or concerns here. Thanks again!

CurtisRobertOliver:
This PR has been updated to contain the minimal solution for issue 2403.

After some research, I think I have a better understanding of your point regarding the middleware changes needed to support a single pipeline for both WS & HTTP. Certain middleware which attempt to shape the response will not function with WebSocket upgrades, as they may attempt to interact with the response after the connection has already been upgraded. Merging WS & HTTP into a single pipeline requires an IsWebSocket check inside each middleware (as you suggested), and if true opt out of any of the following operations: 

1. Read/consume the request body
2. Buffer, wrap, or replace the response stream 
3. Write to or transform the response 
4. Reliance on synchronous operations related to the original request/response

Let me know if you'd like me to open up a separate issue to track and discuss this effort. I am interested in discussing & working on this further.

raman-m:
> I would be happy to take ownership of the goal to combine WS & HTTP into one single pipeline via a separate PR/issue. That said, I'd like to further discuss the scope of which middlewares require updates, and what exactly each WebSocket check needs to achieve within those middlewares.

OK. But we will merge both pipelines in the next release. It's a bit risky to do this in the current release, since it's about to go out.
Yes, we will only fix bug #2403 and nothing more.

acceptance/WebSockets/WebSocketsSecurityTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        route.DownstreamScheme = Uri.UriSchemeWs;
+        route.SecurityOptions = new() { IPBlockedList = [ip] };
+        var gateway = new UriBuilder(Uri.UriSchemeWs, "localhost", ports[1]).Uri;
+        using var ws = new ClientWebSocket();
+        ws.Options.SetRequestHeader(ForwardedHeadersDefaults.XForwardedForHeaderName, ip);
+        this.Given(_ => GivenWebSocketsServiceIsRunningAsync(ports[0], "/", EchoAsync))
+            .And(_ => GivenThereIsAConfiguration(GivenConfiguration(route)))
+            .And(_ => StartOcelotBehindForwardedHeaders(ports[1]))
+            .When(_ => WhenIConnect(ws, gateway))
+            .Then(_ => ThenTheUpgradeIsRejected())
```
raman-m: We have a problem here!

acceptance/WebSockets/WebSocketsSecurityTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            .Then(_ => ThenTheUpgradeSucceeds())
+            .And(_ => ThenTheDownstreamEchoesMessage(ws))
+        .BDDfy();
+    }
+
+    private async Task WhenIConnect(ClientWebSocket ws, Uri gateway)
+        => _connect = await Record.ExceptionAsync(() => ws.ConnectAsync(gateway, CancelMe));
+
+    // A blocked upgrade returns a non-101 status, so ConnectAsync throws WebSocketException (it never bypasses to OK).
+    private void ThenTheUpgradeIsRejected() => _connect.ShouldBeOfType<WebSocketException>();
```
raman-m: **No status check ❗**

I suspect the WS client should not fail here with an exception at all, but that needs to be double-checked. When you debug this scenario, you'll notice that the client fails during response validation with:
- "The server returned status code '200' when status code '101' was expected."

In this scenario, Ocelot returns 200 OK, which is incorrect because the gateway should reject the request and respond with `403 Forbidden` instead.

acceptance/WebSockets/WebSocketsSecurityTests.cs
```diff
@@ -0,0 +1,90 @@
...
+    private async Task ThenTheDownstreamEchoesMessage(ClientWebSocket ws)
+    {
+        const string message = "allowed-ip-probe";
+        using var cts = CancellationTokenSource.CreateLinkedTokenSource(CancelMe);
+        cts.CancelAfter(TimeSpan.FromSeconds(3)); // guard against an indefinite receive hang
+        await ws.SendAsync(Encoding.UTF8.GetBytes(message), WebSocketMessageType.Text, true, cts.Token);
+        var buffer = new byte[256];
+        var result = await ws.ReceiveAsync(buffer, cts.Token);
+        Encoding.UTF8.GetString(buffer, 0, result.Count).ShouldBe(message);
+        await ws.CloseAsync(WebSocketCloseStatus.NormalClosure, "done", cts.Token);
```
raman-m, line 173: The client reports the correct `NormalClosure` status, but **the status code is 0, which is unexpected.**

We need to examine the HTTP conveyor and verify that the `SecurityMiddleware` returns the status, and that it is reflected in the `HttpStatusCode` property after successful closure in this WebSocket scenario. Currently, the `HttpStatusCode` property is 0.
I expect that we have possibly another bug with returning expected correct 101 status from Ocelot 😕

raman-m (changes requested):
We need to work more on this bug fix solution.
I've expressed my mood below.

CurtisRobertOliver:
Sorry @raman-m , that was a misclick. I did not mean to request another review, I am still reviewing your comments.

acceptance/WebSockets/WebSocketsSecurityTests.cs
CurtisRobertOliver: Ah , my test is missing a key assertion that HttpsStatuscode == 101.  I'll investigate further why we are receiving a 200 when we should be receiving a 403.

acceptance/WebSockets/WebSocketsSecurityTests.cs
CurtisRobertOliver: It looks like SecurityMiddleware is indeed not setting the status code. I missed this due to the status code gap in my tests. I think the solution here is to add ResponderMiddleware to the WS pipeline as well to ensure status codes are properly set when completing the WS upgrade handshake.

CurtisRobertOliver: To confirm my understanding, what should we expect to happen on a blocked WS upgrade? 

A. connection abort (no http status)

OR

B. 403 forbidden response (standard response for declined WS upgrades)

I think option B is the correct approach based on https://datatracker.ietf.org/doc/html/rfc6455

CurtisRobertOliver:
Ok, should be fixed! Summary of changes: 

1) SecurityMiddleware now manually sets 403 Forbidden on denied WebSocket upgrades
2) The HttpStatusCode being 0 was due to a test client config issue. I forgot to set CollectHttpResponseDetails=true, so .NET was not recording the status 
3) I had to swap out a direct assertion (`HttpStatusCode == 403`) for the `StackTrace.ShouldNotContain(ValidateResponse)`, because 403 status fails that test

raman-m:
@CurtisRobertOliver
I have a few concerns regarding this PR. The changes feel quite automated, and I would prefer if we could review the logic manually to ensure everything is solid before merging.

**Could you please take a moment to verify your identity by linking your LinkedIn to your GitHub profile (under Contact Info)?** This is just a standard verification step on our end. **In the meantime, please pause further activity on this PR until we get that sorted out.**

Thanks for your understanding!

CurtisRobertOliver:
Hello, I added a link to my GitHub in the contact info section of LinkedIn. 

I am using Claude code to assist with this PR. I am not automating the work, but I am using Claude as a tool. 

I apologize if that is against your policy, I did not see any mention of restricting AI assistants in the contributor documentation.

acceptance/WebSockets/WebSocketsSecurityTests.cs
raman-m: Fixed in https://github.com/ThreeMammals/Ocelot/pull/2406/commits/de290c919fc34912d2ce2f2f8f9c515d7855980c

acceptance/WebSockets/WebSocketsSecurityTests.cs
raman-m: Fixed in commits https://github.com/ThreeMammals/Ocelot/pull/2406/commits/de290c919fc34912d2ce2f2f8f9c515d7855980c and https://github.com/ThreeMammals/Ocelot/pull/2406/commits/cf167a367e6d88994ac01839f8c262b99e94723b  ✔️

acceptance/WebSockets/WebSocketsSecurityTests.cs
raman-m: Option A doesn't apply well here because the user won't see the reason, while Option B is more informative since it returns a 403 status code.
Great that you referenced the RFC 6455 specification 👍

OK It's fixed in commits https://github.com/ThreeMammals/Ocelot/pull/2406/commits/de290c919fc34912d2ce2f2f8f9c515d7855980c and https://github.com/ThreeMammals/Ocelot/pull/2406/commits/cf167a367e6d88994ac01839f8c262b99e94723b  ✔️

raman-m:
> Hello, I added a link to my GitHub in the contact info section of LinkedIn.

Congrats!
I have no objections to the use of AI. I do, however, warn contributors when they violate the development process.
After the .NET 10 release, I will publish an official announcement regarding our identity verification procedure for new contributors.

CurtisRobertOliver:
Thank you! I appreciate your feedback and candor. My main goal here is to improve my skills as a contributor, so I am always open to constructive criticism :)

raman-m:
Technically the bug fix is ready, but I want to draw attention to the following issues:

1. An HTTP request with a failed security check returns 401 Unauthorized (see [IPSecurityPolicy](https://github.com/ThreeMammals/Ocelot/blob/develop/src/Security/IPSecurityPolicy.cs) class). I propose we fix this in the current pull request, because if the WebSocket security response is 403 but the HTTP security response is 401, then we have inconsistent status codes.
https://github.com/ThreeMammals/Ocelot/blob/6062d5a5ab3bc1baa929d0965e5a28e3cbf2d1fd/src/Security/IPSecurityPolicy.cs#L23
Both cases should return the same 403 status.
**I think you could handle this case.** I propose to add a dedicated `IpSecurityError` based on 403 status inside to replace current invalid `UnauthenticatedError` which is based on 401 status.

2. As I noted earlier in https://github.com/ThreeMammals/Ocelot/issues/2403#issuecomment-4728794195, there's a log warning about auto-replaced schemes. This is the most critical bug uncovered by #2403 sample: an attacker could exploit protocol auto-switching to reach a WebSocket downstream by setting http as the downstream scheme. We added this feat after requesting the feat by community user. But auto upgrade is not good from security perspective.

   I know this is out of scope for now, but I wanted to flag it again. I'll open a dedicated issue later.
  Since Ocelot doesn't support fallback between protocols, we should reject requests that try to use a mismatched protocol (e.g., `ws` → `http`).
  I don't have time to investigate this deeply now, but it should be a priority for the next release.

3. We need to include the original #2403 user case and create an acceptance test that wraps George Chen's code and adapts it for an auto-test. **I will take responsibility for this piece of work.** Thus, we will close the bug officially.

CurtisRobertOliver:
> I think you could handle this case. I propose to add a dedicated IpSecurityError based on 403 status inside to replace current invalid UnauthenticatedError which is based on 401 status.

Sounds good! I'll take this on to make sure the HTTP pipeline lines up with these changes to WS.

CurtisRobertOliver:
Hi Raman, anything else I can help complete on this PR? I think it's pretty much ready to ship!

raman-m (approved):
## Ready for delivery ✅ 
- Code review ✔️ ✔️ 
- Unit tests ✔️ ✔️ 
- Acceptance tests ✔️ ✔️ 
- Updated docs ✔️ ✔️ 

### TODO Browser testing
Ideally, we could add a manual test to the `Ocelot.ManualTest` project, but since it is not ready for automated testing and there is a tight timeline for the current hot release, we are skipping both manual and automated browser tests for now.

raman-m:
@CurtisRobertOliver 
> Hi Raman, anything else I can help complete on this PR? I think it's pretty much ready to ship!

Nothing more. I completed the remaining work as per the [development process](https://ocelot.readthedocs.io/en/develop/building/devprocess.html), and the PR has been merged after receiving the "Dev Complete" status.
I hope you enjoyed your first contribution to Ocelot and that our [dev process](https://ocelot.readthedocs.io/en/develop/building/devprocess.html) is now clear to you.
Looking forward to more contributions from you! 😍 
Thank you!

CurtisRobertOliver:
Thank you for your help getting this over the finish line! I'm looking forward to contributing more and following your lead from this PR to improve my future contributions. 

I am considering taking on the larger piece of work to merge the WS & HTTP pipelines which we mentioned earlier in this PR. 

If you have any other issues you think would be better suited for a newer C# contributor like myself, I am open to suggestions :)

Thank you again!

raman-m:
Yes, yes, yes, but after the [.NET 10](https://github.com/ThreeMammals/Ocelot/milestone/13) release.

Before future contributions, though, I'm curious: how do you use Ocelot at your company or in your other projects? I don't imagine corporate usage is considered confidential info.

CurtisRobertOliver:
> Yes, yes, yes, but after the .NET 10 release.

Absolutely, I understand this effort will be deferred until after the next release. 

I do not use Ocelot at work or for personal projects, I came in through the "help wanted" tag. I was looking for open issues in projects with active & knowledgeable maintainers, and this one fit the bill!

raman-m:
> I am considering taking on the larger piece of work to merge the WS & HTTP pipelines which we mentioned earlier in this PR.

Yes, this is a nice-to-have follow-up task. No issue created yet, but the idea is clear. Let me know if you need official issue.
Feel free to start working on a draft PR on your own. Just don't expect my help until the next release cycle.

> I came in through the "help wanted" tag.

Right. There are other issues with the "[help wanted](https://github.com/ThreeMammals/Ocelot/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22help%20wanted%22)" label that you could consider as a next possible contribution. However, most of them are complex tasks requiring medium to large effort.

Finally, I recommend that you contribute to the task about merging the HTTP and WS pipelines.

CurtisRobertOliver:
> Feel free to start working on a draft PR on your own. Just don't expect my help until the next release cycle.

Completely understood, I know that you are very busy preparing for the .NET 10 release. Thank you again for your guidance & advice!

PR #1372, opened 2020-11-13
Add 'multipart/form-data' support for file uploads

myall200:
## Fixes #714 
- #714 

## Motivation
Introduce a support of file streaming (file uploads) via the gateway, without body content caching.
Current loading of entire body to a memory in Ocelot core is a source of errors, unstable responses because downstream services produce errors during long lasting operation of the request.

## Proposed Changes
- add `multipart/form-data` support for file uploads

## Related to
- #792 
- #714 
- #617 
- #1685

PR #2421, opened 2026-08-31
test: add multipart/form-data rerouting regression coverage for #714

### LINKED ISSUE #714: Multipart/form-data is not rerouted (error 404)
## Expected Behavior
A multipart/form-data http POST request is redirected to the desired endpoint.

## Actual Behavior
An error 404 - not found http response is given.

## Description
I won't include steps to reproduce the problem etc, because I'm pretty sure the feature is just not implemented (yet - or intentional).

## Question 
Is rerouting a file upload using multipart/form-data desired functionality, or is in an intentionally left out thing?
I could imagine you'd not want to reroute such a file upload because of the overhead - why not just upload the file directly to the API? 
However, if the overhead is as little as the normal http requests are (about 1 ms), then I'd think it would be preferable to support multipart/form-data http post requests in reroutes.

Edit: 
The reason I'm certain that I haven't just incorrectly configured the reroute, is as follows:
When I send the request with no file in the parameter, the reroute works and gets into my other API.
When I send the request with a file in the parameter, error 404 occurs, thrown by Ocelot presumably.

rodri-oliveira-dev:
## Closes #714
- #714

### Summary
Adds regression coverage and a real browser sample for #714 against the current `develop` branch.

The original issue reported that `multipart/form-data` requests containing a file returned 404 instead of being rerouted. After validating the current request pipeline, the original issue is no longer reproducible for regular, non-aggregated routes: Ocelot streams the raw request body and preserves the original `Content-Type`, including the multipart boundary.

No production-code change under `src/` is required for the regular rerouting scenario. This PR documents and protects that behavior.

### Coverage added
* Acceptance coverage that reroutes `POST /upload` to `/api/files` using `multipart/form-data`
* Verifies that the downstream route is reached successfully
* Verifies the downstream path
* Verifies the multipart `Content-Type` and boundary
* Verifies a regular form field
* Verifies the file field name, original filename, file content type, and exact file contents
* Unit coverage for `RequestMapper` confirming that the multipart body and `Content-Type` boundary are preserved unchanged
* Unit coverage for the multiplexer context clone confirming that a cloned multipart request body can still be read as form data without copying `Request.Form`

### Real sample
Adds `samples/MultipartFormData` with two small apps:
* `ApiGateway`: serves a valid HTML page containing a real `enctype="multipart/form-data"` form with a text field and file input, then forwards `POST /upload` through Ocelot
* `DownstreamService`: handles `POST /api/files`, reads the request using `ReadFormAsync`, and returns the received form field, file metadata, file contents, and request content type

The sample README includes the run instructions and uses only the existing ASP.NET Core/Ocelot project setup.

### Review follow-up
* The one-test constants in the acceptance test were moved to local variables and passed into helpers where needed.
* The requested real sample/browser form scenario was added with an actual HTML upload page.
* I reviewed the `Form = from.Form` idea from #1372. It was not technically necessary for this PR: the regular route works through the existing raw-body streaming path, and the multiplexer clone can read multipart form data from the cloned body plus copied content headers. Therefore this PR intentionally does not add `Form = from.Form`.
* POST aggregation/multiplexing support remains out of scope and should be treated as a future feature/design discussion.

### Validation
Validated locally with .NET SDK `10.0.400`:
* `dotnet restore Ocelot.slnx` passed
* `dotnet restore Ocelot.Samples.slnx` passed, with existing ServiceFabric/Steeltoe dependency warnings outside this PR
* `dotnet build Ocelot.slnx --no-restore` passed with 0 warnings/errors
* `dotnet build Ocelot.Samples.slnx --no-restore` passed, with the same existing ServiceFabric/Steeltoe warnings
* `dotnet format Ocelot.slnx --no-restore --verify-no-changes --include acceptance/Request/MultipartFormDataTests.cs unit/Multiplexing/MultiplexingMiddlewareTests.cs unit/Request/Mapper/RequestMapperTests.cs` passed
* `dotnet format Ocelot.Samples.slnx --no-restore --verify-no-changes --include samples/MultipartFormData/ApiGateway/Program.cs samples/MultipartFormData/DownstreamService/Program.cs` passed
* Issue #714 unit tests passed on `net8.0` and `net10.0`
* Issue #714 acceptance test passed on `net8.0` and `net10.0`
* Full unit suite passed on `net8.0` and `net10.0` with `--culture invariant` (`1901` passed per TFM)
* Full acceptance suite on `net10.0` was run; it had one unrelated `ClientWebSocketTests.Http20ClientWhenDirectConnectionThenShouldConnect` HTTP/2 WebSocket timeout. Rerunning the suite excluding only that test passed (`358` passed, `3` skipped).
* The new sample was run manually: `GET http://localhost:5567/` returned `200`, and a real multipart `POST /upload` through the gateway returned `200` with the expected text field, file metadata, exact file contents, and multipart boundary received downstream.

Local note: `net9.0` builds successfully, but the .NET 9 runtime is not installed in this environment, so local test execution was limited to `net8.0` and `net10.0`.

### Production impact
No production files under `src/` are changed.

The current implementation already supports the regular `multipart/form-data` rerouting scenario, so this PR adds regression coverage and a real executable sample to prevent that behavior from being unintentionally broken in the future.

Refs #714


acceptance/Request/MultipartFormDataTests.cs
```diff
@@ -0,0 +1,76 @@
...
+
+public sealed class MultipartFormDataTests : Steps
+{
+    private const string UpstreamPath = "/upload";
+    private const string DownstreamPath = "/api/files";
+    private const string FieldName = "description";
+    private const string FieldValue = "issue-714";
+    private const string FileFieldName = "file";
+    private const string FileName = "test.txt";
+    private const string FileContent = "multipart-file-content";
```
raman-m: There's only one test, but these constants are used just for it. I'd suggest defining them as local variables and passing some as arguments instead.

raman-m (changes requested):
> ### Production impact
> 
> **No production files under `src/` are changed.**
> 
> The current implementation already supports the regular `multipart/form-data` rerouting scenario, so this PR only adds regression coverage to prevent that behavior from being unintentionally broken in the future.
> 
> Refs #714
> 

I'm going to change the requirements. I can't approve this without real improvements.
We need to test it using a sample or via a browser test with an actual HTML page.

raman-m:
> ### Scope
> 
> This PR intentionally covers the regular, non-aggregated routing scenario discussed in #714.
> 
> Aggregated routes using POST/multiplexing are intentionally out of scope.
> 
> As discussed in the issue, aggregation currently has additional constraints around POST requests and raises a broader design question about combining potentially heterogeneous downstream responses such as JSON, XML, HTML, headers/status-only responses, or other payload types.
> 
> That scenario should therefore be treated separately from the regular multipart rerouting behavior validated by this PR.
> 

Agreed. This could become a future feature proposal. I'll develop the concept further and probably create an official feature issue during the current release. **In cooperation with you or without.**

acceptance/Request/MultipartFormDataTests.cs
rodri-oliveira-dev: Adjusted, thanks.

rodri-oliveira-dev:
Thanks, @raman-m. I added the requested real sample in `samples/MultipartFormData`.

It now includes an actual HTML page served by the gateway with `enctype="multipart/form-data"`, a text field, and a file input. The form posts to `POST /upload`, Ocelot forwards it to `/api/files`, and the downstream service reads the request with `ReadFormAsync` and returns the received field, file metadata, file contents, and multipart content type.

I kept the automated regression coverage as well: the acceptance test validates the forwarded form field, file field name, filename, file content type, exact file contents, downstream path, and multipart boundary; the unit tests cover `RequestMapper` raw body/content type preservation and the multiplexer clone behavior.

I also reviewed the `Form = from.Form` idea from #1372. I did not add it because it was not technically necessary for this PR: regular routing already works through the raw body streaming path, and the cloned multiplexer request can parse multipart data from the copied body plus content headers. POST aggregation/multiplexing remains out of scope for this PR.

Validation passed locally for restore/build, sample build, formatting checks, issue #714 unit/acceptance tests on net8/net10, full unit suite on net8/net10 with invariant culture, and full net10 acceptance suite when excluding only the unrelated HTTP/2 WebSocket direct-connection timeout.

rodri-oliveira-dev:
Added the maintainer-requested Multiplexer edge case in `acceptance/Request/MultipartFormDataTests.cs`.

Coverage now includes a single POST route referenced by `Key` from an `Aggregate` with `UpstreamHttpMethod = [Post]`. The test posts `multipart/form-data` to the aggregate upstream path and validates the downstream path, multipart `Content-Type` and boundary, text form field, file field name, filename, MIME type, exact file contents, and success status.

I also made the route upstream path distinct from the aggregate upstream path, so `/upload` is exercised through the aggregate route rather than the regular route.

Local validation:

```text
dotnet test acceptance\Ocelot.Acceptance.csproj -f net10.0 --filter FullyQualifiedName~MultipartFormDataTests
Passed: 2/2

dotnet test unit\Ocelot.UnitTests.csproj -f net10.0 --filter FullyQualifiedName~RequestMapperTests|FullyQualifiedName~MultiplexingMiddlewareTests
Passed: 55/55
```

Note: running without `-f net10.0` is blocked in my local environment because the net9.0 targeting pack/runtime is not installed. No production code changes were needed for this edge case.

